PR #240, opened 2026-04-21
fix(routecraft): attach DSL sugar and auth header types to public package

ex0b1t:
Two `declare module` blocks targeted relative source paths:
`declare module "./builder.ts"` in `dsl.ts`, and
`declare module "../exchange.ts"` in `auth/types.ts`. tsup bundles the
package declarations into a single `dist/index.d.ts`, and those
relative specifiers survive verbatim into the published file where
they no longer resolve in a consumer's module graph. TypeScript
silently drops the augmentations and the methods / header keys vanish
from the public `@routecraft/routecraft` types.

Concretely this made `.log`, `.debug`, `.map`, and `.schema` disappear
from the published `RouteBuilder` type, and the `routecraft.auth.*`
keys disappear from `RoutecraftHeaders`. The augmentations worked in
our own test suite only because those tests import from `../src/`,
which lets the relative specifier resolve during source compilation.

Retarget both augmentations to `"@routecraft/routecraft"`, matching
the pattern used by every other augmentation in the codebase
(`cron/source.ts`, `mail/shared.ts`, `direct/shared.ts`, `split.ts`,
etc). The same interface-merging rules apply because `RouteBuilder` is
exported as a class (which produces a merge-compatible interface
symbol) and `RoutecraftHeaders` is already an interface.

Regression test
- New `examples/test/dsl-sugar.test.ts` imports from
  `@routecraft/routecraft` (not `../src/`) and exercises `.log`,
  `.debug`, `.map`, `.schema` via `expectTypeOf`. The examples
  workspace links against the built package through pnpm, so if any of
  these augmentations detaches from the public entry again, both
  `pnpm test` and `pnpm build` fail on type-check.

Verified against the built `dist/index.d.ts`: all nine
`declare module` blocks now target `"@routecraft/routecraft"`.

Co-authored-by: Claude <claude@anthropic.com>

---
## Summary by cubic
Fixes lost type augmentations in `@routecraft/routecraft` so consumers see DSL sugar (`.log`, `.debug`, `.map`, `.schema`) and auth header types again, and tightens CI to avoid stale `dist/` in PRs.

- **Bug Fixes**
  - Retargeted augmentation `declare module` blocks in `dsl.ts` and `auth/types.ts` to `@routecraft/routecraft`; documented the rule in `.standards/type-safety-and-schemas.md`; added `.log()` in the `hello-world` template to assert the public types.
  - Mail: `resolveImapOptions` now forwards `verify` for named accounts.
  - CI/tests: scope `**/dist` cache keys to the PR head SHA; rebuild `@routecraft/routecraft` at the start of integration tests.

<sup>Written for commit 1b8d95ddf3bd2caea3a76993b8e4aa8d2b91b62b. Summary will update on new commits.</sup>

### DIFF
--- a/packages/create-routecraft/templates/examples/hello-world/capabilities/hello-world.ts
+++ b/packages/create-routecraft/templates/examples/hello-world/capabilities/hello-world.ts
@@ -20,6 +20,8 @@ const greetRoute = craft()
     }),
   )
   .transform((result) => `Hello, ${result.body.name}!`)
+  // `.log()` is sugar for `.tap(log())`: a type-preserving log tap.
+  .log()
   .to(log());
 
 // "hello-world" caller: dispatches a user id to the greet service.
--- a/packages/create-routecraft/templates/examples/hello-world/capabilities/hello-world.ts
+++ b/packages/create-routecraft/templates/examples/hello-world/capabilities/hello-world.ts
@@ -20,7 +20,6 @@ const greetRoute = craft()
     }),
   )
   .transform((result) => `Hello, ${result.body.name}!`)
-  // `.log()` is sugar for `.tap(log())`: a type-preserving log tap.
   .log()
   .to(log());
 
--- a/packages/create-routecraft/test/integration.test.ts
+++ b/packages/create-routecraft/test/integration.test.ts
@@ -3,7 +3,7 @@ import { mkdir, rm, readFile, readdir, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { existsSync } from "node:fs";
-import { exec, spawn, type ExecOptions } from "node:child_process";
+import { exec, execSync, spawn, type ExecOptions } from "node:child_process";
 import { promisify } from "node:util";
 import { generateProjectStructure, type InitOptions } from "../src/lib.js";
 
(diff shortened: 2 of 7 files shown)


packages/create-routecraft/templates/examples/hello-world/capabilities/hello-world.ts
```diff
@@ -20,6 +20,8 @@ const greetRoute = craft()
     }),
   )
   .transform((result) => `Hello, ${result.body.name}!`)
+  // `.log()` is sugar for `.tap(log())`: a type-preserving log tap.
```
ex0b1t: remove this comment

packages/create-routecraft/test/integration.test.ts
```diff
@@ -214,6 +214,18 @@ const packagesBuilt = existsSync(
 );
 const integrationTest = packagesBuilt ? test : test.skip;
 
+// Rebuild the routecraft package so its dist/ matches the checked-out
+// source. On local runs this catches stale dist after editing source without
+// running `pnpm build`. On CI it is a belt-and-suspenders against cache
+// issues on the `pull_request_target` workflow (see cache key scoping in
+// ci.yml). Cheap — one tsup run on a single package.
```
ex0b1t, line 221: Is this not just a hack to make the it work but then real projects will fail?

do we need this?

packages/routecraft/src/auth/types.ts
```diff
@@ -144,7 +144,10 @@ export interface OAuthValidatorAuthOptions {
   validator: OAuthTokenVerifier;
 }
 
-declare module "../exchange.ts" {
+// Must target the published package specifier; a relative path would survive
```
ex0b1t: this is not the right place to put this, this belongs in one of the standard docs or something or not at all not this bunch of comments

packages/routecraft/src/dsl.ts
```diff
@@ -139,7 +139,14 @@ registerDsl("schema", {
 // Module augmentation: TypeScript types for built-in sugar
 // ---------------------------------------------------------------------------
 
-declare module "./builder.ts" {
+// Must target the published package specifier, not `"./builder.ts"`. After
```
ex0b1t: see same duplicated comments instead of a standard

PR #242, opened 2026-04-22
Isolate MCP local tools from direct routes with dedicated registry

ex0b1t:
## Summary

This PR decouples MCP local tool routes from the direct adapter's registry, giving `mcp()` its own dedicated `MCP_LOCAL_TOOL_REGISTRY`. This allows `mcp()` and `direct()` routes to coexist with identical endpoint names without collision, and ensures MCP tool discovery only exposes tools defined via `mcp()`, not all direct routes.

## Key Changes

- **New MCP Local Tool Registry**: Introduced `MCP_LOCAL_TOOL_REGISTRY` store key and `McpLocalToolEntry` interface to hold MCP-specific tool metadata (title, description, schema, outputSchema, annotations, icons) and invocation handlers
- **Isolated McpSourceAdapter**: Refactored `McpSourceAdapter.subscribe()` to populate its own registry instead of delegating to the direct adapter, with full validation and handler wrapping inline
- **Header Validation Merge**: Implemented header schema validation that merges validated values over originals, preserving MCP-injected headers (tool name, session, auth principal) even when schemas strip unknowns
- **Direct Registry Cleanup**: Removed `keywords` and `annotations` fields from `DirectRouteMetadata` (adapter-specific metadata now lives on each adapter's own types); kept `title` and `outputSchema` for structural compatibility
- **Test Refactoring**: Updated MCP tests to use `invokeTool()` helper that directly invokes registry entries instead of relying on producer/consumer patterns; simplified test setup by removing producer routes
- **TestContext Enhancement**: Extracted `awaitRoutesReady()` helper to avoid duplication between `startAndWaitReady()` and `test()` methods; clarified that `startAndWaitReady()` does not await `ctx.start()` completion (allowing it to work with blocking sources like direct and mcp)

## Implementation Details

- MCP endpoints are URL-encoded when stored in the registry to keep keys like `"a/b"` and `"a-b"` distinct
- Duplicate endpoint detection prevents multiple `mcp()` routes from registering the same endpoint in a context
- Entry handlers perform body and header validation before invoking the route consumer, with proper error emission for validation failures
- Cleanup listeners ensure registry entries are removed when routes are aborted
- `McpServer.getAvailableTools()` now reads from `MCP_LOCAL_TOOL_REGISTRY` instead of filtering the direct registry, and converts entries to MCP wire format with full tool shape support (title, outputSchema, icons, annotations)

https://claude.ai/code/session_013c6nKtL7J3WfGr3CeHZp1Q

---
## Summary by cubic
Isolated MCP local tools behind a dedicated `MCP_LOCAL_TOOL_REGISTRY` so `mcp()` and `direct()` can share endpoint names without collisions, and MCP discovery no longer includes direct routes. Also moved schemas under `input`/`output`, preserved adapter-injected headers by merging validated headers over originals, and enforced strict MCP tool-name validation.

- **New Features**
  - Added `MCP_LOCAL_TOOL_REGISTRY` and `McpLocalToolEntry`; `mcp()` registers tools here with duplicate-endpoint protection and tool-name validation (`^[A-Za-z0-9_-]{1,64}$`).
  - `McpServer` serves `tools/list` from the MCP registry, flattening `input`/`output` to MCP wire fields and forwarding `title`, `outputSchema`, `annotations`, and `icons`.
  - `mcp()` and `direct()` merge validated headers over the original headers to keep MCP/session/auth metadata intact.

- **Migration**
  - Schemas now live under `input`/`output`:
    - `{ schema: X }` → `{ input: { body: X } }`; `{ headerSchema: H }` → `{ input: { headers: H } }`; `{ outputSchema: O }` → `{ output: { body: O } }`.
  - `direct()` no longer accepts `keywords`, `annotations`, or `icons`. Move `annotations`/`icons` to `mcp()`; `keywords` is removed.
  - `McpServerOptions` is standalone and uses `input`/`output`; `description` remains required.
  - `McpPluginOptions.tools` now receives an `McpLocalToolEntry` (not `DirectRouteMetadata`).
  - MCP discovery reads `MCP_LOCAL_TOOL_REGISTRY`; `ADAPTER_DIRECT_REGISTRY` is for in-process direct routes only.
  - Tests: `TestContext.startAndWaitReady()` does not await `ctx.start()`; call `t.stop()` for teardown.

<sup>Written for commit 44e122379b97c8a9146152f74bfa8c734088f6a4. Summary will update on new commits.</sup>


apps/routecraft.dev/src/app/docs/reference/adapters/page.md
```diff
@@ -302,10 +302,15 @@ craft()
 
 **Options:**
 - `channelType` - Custom direct channel implementation (default: in-memory)
-- `schema` - Body validation schema (StandardSchema compatible: Zod, Valibot, ArkType)
-- `headerSchema` - Header validation schemas (can be optional/required)
-- `description` - Human-readable description for route discovery
-- `keywords` - Keywords for route categorization
+- `schema` - Body validation schema, Standard Schema compatible (Zod, Valibot, ArkType). Equivalent to the MCP `Tool.inputSchema` when exposed as a tool
```
ex0b1t: lets make a breaking change and just call it inputSchema and outputSchema

and headerSchema must say its input, we also need a output header schema 

or lets think about this, should we not combine it so that schema is exchange and not just body?

ex0b1t: for direct endpoints this might be a really good feature so we can assert things like is authenicated etc on some direct routes even before entry?

ex0b1t: this is probably also where mcp(tools) diverge as they only need body, but that is fine as channels will call direct endpoints, direct endpoints are re-usable parts

apps/routecraft.dev/src/app/docs/reference/adapters/page.md
```diff
@@ -302,10 +302,15 @@ craft()
...
+- `headerSchema` - Header validation schema; when provided, validated values are merged on top of the request headers so adapter-injected metadata is preserved
+- `title` - Human-readable display title for discovery consumers (agents, MCP clients)
+- `description` - Human-readable description surfaced to agents and (when exposed via `mcp()`) MCP clients
+- `outputSchema` - Standard Schema describing the output body; forwarded to discovery consumers
+- `annotations` - Opaque annotations pass-through; `mcp()` narrows this to `McpToolAnnotations`
+- `icons` - Opaque icons list forwarded to discovery consumers (typed per the MCP spec by `mcp()`)
```
ex0b1t: under direct we document icons, annotations this is not what i asked for these are two distinct adapters and direct must not refer to mcp, mcp is in another module

apps/routecraft.dev/src/app/docs/reference/adapters/page.md
```diff
@@ -302,10 +302,15 @@ craft()
...
+
+> `direct()` mirrors the MCP `Tool` shape so in-process agents can discover direct routes directly. `mcp()` maintains a separate registry (`MCP_LOCAL_TOOL_REGISTRY`) for externally exposed tools; direct and mcp routes can coexist on identical endpoint strings because they live in different registries.
```
ex0b1t: this should be removed

apps/routecraft.dev/src/app/docs/reference/adapters/page.md
```diff
@@ -456,29 +461,31 @@ craft()
 
 #### Route Registry
 
-All direct routes are registered and can be queried. Routes with descriptions and keywords are more discoverable:
+Each direct route registers in `ADAPTER_DIRECT_REGISTRY` with the full tool-shape metadata, so in-process agents can discover and document them the same way they would MCP tools:
```
ex0b1t: i expect seperate sections for direct and mcp registries they are not connected

examples/src/hello-world.ts
```diff
@@ -9,9 +9,7 @@ type GreetInput = z.infer<typeof GreetInput>;
 // Standard Schema passed to direct(), so no cast is required.
 const greetRoute = craft()
   .id("greet")
-  .from(
-    direct("greet", { description: "Greet a user by id", schema: GreetInput }),
-  )
+  .from(direct("greet", { schema: GreetInput }))
```
ex0b1t: direct also allows descriptions and we should in our docs alway add them but if a user will never use mcp or ai then they can skip them as they are optional but examples should have it.

apps/routecraft.dev/src/app/docs/reference/adapters/page.md
ex0b1t: Landed in `b9b4d29`. Direct + MCP both moved to nested-by-direction:

- `direct(...)`: `{ input: { body, headers }, output: { body, headers }, title?, description? }`
- `mcp(...)`: same shape plus the MCP-only fields (`description` required, `annotations`, `icons`).

So `inputSchema` / `outputSchema` / `inputHeaderSchema` / `outputHeaderSchema` are not separate keys — they're addressed as `input.body`, `input.headers`, `output.body`, `output.headers`. Same information, less prefix noise, and any future per-direction concerns (`query`, `params`, `attachments`) slot in without bloating the top level.

Skipped the unified-exchange-schema variant: the DX cost (every common-case body-only validation becomes nested under `body:`, builder type inference has to dig through the wrapper) didn't justify the cross-field validation use case, which `.refine()` already covers.

Comments #2/#3/#4/#5 also addressed in the same commit.

PR #510, opened 2026-07-28
Blog figures: fix Safari rendering, add PNG export and a full-screen view

ex0b1t:
## What

Blog figures and covers rendered wrong in Safari at every viewport width, were unreadable on a phone in every browser, and showed up as a raw Markdoc tag in the markdown versions of posts. This fixes all three, and gives every figure a stable PNG URL so posts can be syndicated with their artwork.

## The Safari bug

`ScaledFrame` scaled fixed-canvas artwork with an SVG `viewBox` + `<foreignObject>`. **WebKit paints foreignObject content at 1:1, ignoring the viewBox, as soon as anything in the subtree is positioned.** Every one of these drawings positions something, so Safari showed the top-left corner of the canvas at full size: covers with no title, figures with one label and acres of empty paper. This affected the post hero, the blog index cards, the home teasers and every in-post figure, on desktop as well as mobile.

Isolated factorially: the trigger is `position: relative` on the foreignObject's direct child. Explicit `width`/`height` attributes, an `xmlns`, `overflow: hidden`, a static wrapper and `translateZ(0)` all fail to help. Only dropping foreignObject does, so the frame now measures its own width and applies a CSS transform. WebKit and Chromium now render identically.

## Mobile

A 1600x900 drawing in a 358px column is complete but unreadable, and the old 34rem floor left a third of every figure behind a horizontal scroller. The figure now fits the column and opens full-screen on tap, showing the exported PNG at 1229x692 against 358x201 before, theme-matched, with both edges reachable.

## Figure pages and export

`/figures/` and `/figures/<id>/` show each figure with the URLs and markdown snippet for its PNGs. `bun run figures:export` drives Chromium over those pages and writes `<id>.png` and `<id>-dark.png` at 2x into `public/`. Light is the unsuffixed default. The PNGs are committed, so the Pages workflow never installs a browser, and they now back the lightbox as well as syndication.

## Raw markdown

`/raw/blog/<slug>.md` passed the Markdoc tag straight through, so anything reading a post off the site saw a literal `{% diagram id="four-gates" /%}` and no artwork. That is the version an LLM crawls and the version you paste into dev.to. The tag now becomes a real markdown image of the light PNG with the figure's alt text and its caption beneath, at an absolute URL.

That needed a figure's words where JSX cannot be loaded, so `alt` and `caption` move to `figures/manifest.mjs` (plain JavaScript, keyed by id) alongside the PNG path helpers. Drawings keep their own `.tsx`; `index.ts` joins the halves and throws if a drawing has no entry.

## Also

The home page scrolled 159px sideways at 390px wide in both engines: the guardrail cards are grid items with default `min-width: auto`, so each column grew to its widest unwrappable code line. Found while checking mobile, fixed here.

## Verification

Screenshot-verified in **WebKit and Chromium**, light and dark, at 390px and 1440px:

- Covers and figures render identically in both engines
- Every surface measures 0px horizontal overflow
- Overlay serves the dark PNG in dark mode and both scroll edges are reachable
- All 12 exported PNGs checked by mean luminance (light 228-238, dark 33-44), since the dark files landing within a few kB of the light ones is exactly what a silent no-op would look like
- All 5 diagram tags across the raw blog files convert, none left behind

`bun run all` green; 1932 tests pass.

## Notes for review

- **`ScaledFrame`'s `fill` mode is removed.** No caller passed it, and reimplementing crop-anchoring in transform maths for zero callers is not worth carrying.
- **Artwork is hidden until first measurement.** The frame reserves space through `aspect-ratio` so the reveal shifts no layout, and a `@media (scripting: none)` fallback pans at 1:1 rather than leaving an empty box.
- **A figure is now authored across two files**: the drawing in `.tsx`, its words in `manifest.mjs`. The reason is in `.standards/content-and-docs.md`; the build throws if the halves drift.
- **Re-run `figures:export` when a figure changes.** The site serves those PNGs now, so a skipped export ships a post whose figure enlarges to the previous drawing.
- `playwright` is a new devDependency of the site app; the browser binary stays an explicit opt-in download.


ex0b1t:
Worked through all ten findings. Eight applied in 3104d898, one rejected, one partially applied.

**Applied**

| Finding | Note |
|---|---|
| Guardrails scroll region not keyboard-reachable (both reviewers) | Correct, and inconsistent with this same PR: the figure scroller got `tabIndex`/`role`/`aria-label` and this one did not. Now matches, with a focus ring. |
| `div` inside `button` in the lightbox trigger | Correct. Trigger is now an empty labelled button stretched over the artwork, so the click target and focus behaviour are unchanged and the tree is valid. Verified: 0 buttons contain block elements. |
| `caption=""` still showed alt text in the overlay | Correct, `text \|\| alt` defeated the documented no-caption behaviour. |
| Export hangs when Chromium is missing | Correct, the listening server kept the event loop alive after the launch error. |
| Manifest guard accepted empty `alt`/`caption` | Correct, an empty alt is exactly the unnamed image the guard exists to prevent. |
| Export script header described the removed foreignObject implementation | Correct, stale after the Safari fix. |
| Standards note path prefixes | Correct, the rest of the file uses `apps/routecraft.dev/`. |

**Partially applied: base-path-aware PNG URLs** (`figures/[id]/page.tsx`)

The inconsistency was real and is fixed: the page displayed an absolute URL while linking a root-relative one, which disagree under `NEXT_PUBLIC_BASE_PATH`. It now links what it displays.

I did not extend this to the lightbox ``. Every public asset on the site is referenced root-relative (markdown images, `/images/blog/*`), and `NEXT_PUBLIC_BASE_PATH` is supported by `next.config.mjs` but never set in CI. Making only the newest call sites base-path-aware would create an inconsistency without fixing the general case. Worth doing as its own change if a mounted deployment is ever actually used.

**Rejected: "Avoid `ScaledFrame` in the OG image path"** (`ScaledFrame.tsx`)

`ScaledFrame` does not reach Satori. `blog-og-image.tsx` imports and renders `<BlogCover>` directly, and `BlogCover`'s own body contains no frame. `ScaledFrame` is reachable only through `BlogCoverFrame` ← `BlogCoverInline`, which is the browser path and is not called by `createBlogOgImage`. The build confirms it: OG images are `force-static` and generate successfully, which a `useLayoutEffect` in the Satori tree would not survive.

The separation the comment asks for is the separation that already exists: `BlogCover` is the Satori-compatible component, `BlogCoverInline` is the browser wrapper.

All checks green, 1932 tests pass. Lightbox re-verified in WebKit and Chromium, light and dark, at 390px and 1440px after the trigger restructure.

PR #513, opened 2026-08-03
feat(routecraft): delegation-aware Principal and authorize()

### LINKED ISSUE #512: feat(routecraft): delegation-aware Principal and authorize() (subject/actor primitives)
# Delegation-aware Principal and authorize()

> **Revised during implementation (PR #513).** Two deliberate changes from the original text, both agreed in design review: (1) the actor's own scopes were **removed from the delegation intersection**; effective scopes are `intersect(subject.scopes, ceiling)`. Rationale: wherever a capability runs on a shared system account there is no "the actor's own access" for a scope to attach to, and intersecting the actor's scopes made "the agent may do X only when a user grants it, never standalone" inexpressible. Actor reachability is enforced by `authorize({ actor })` per route instead. (2) The bare-subject-matcher eslint rule is deferred to a follow-up issue; the JSDoc no longer promises it. The tables below are updated to match.

## User story

As a Routecraft user building agentic capabilities, I want the framework to distinguish three cases at the authorization boundary:

1. a **user acting directly** (subject = user, no actor),
2. an **agent acting on behalf of a user** (subject = user, actor = agent),
3. an **agent acting under its own standing authority** (subject = agent, no actor),

so that a route can declare, for example, "humans only", "only agent X on a user's behalf", or "autonomous agents only", and the framework enforces it regardless of which IdP or consent mechanism produced the identity.

## Scope

Core primitives only: data shapes and pure functions in `@routecraft/routecraft`. Everything here is agnostic of the consent/grant infrastructure, exactly like `authenticate()`/`authorize()` today: the caller resolves authority, the framework mints and verifies.

**Out of scope** (follow-up issues): grant stores, consent ceremonies/UIs, channel identification, the agent-boundary auto-delegation in `@routecraft/ai`, token exchange on outbound MCP hops, the `confidence` field, the bare-subject-matcher lint rule.

The design-time proposal document was removed from the tree once the feature shipped (PR #513): this issue is the spec of record, the rules live in `.standards/security.md` s12, and the API in the `delegate` / `authorize` operation references.

## Standards grounding

Every wire-facing field maps to a published claim. TypeScript names stay spelled out (as `subject`/`expiresAt` already are), JSDoc cites the wire claim.

| Field | Wire claim | Spec | Status |
|---|---|---|---|
| `actor` | `act` | RFC 8693 s4.1 | RFC |
| `mayAct` | `may_act` | RFC 8693 s4.4 | RFC |
| `subjectProfile` | `sub_profile` | draft-mcguinness-oauth-actor-profile-00 | draft (cite as "aligned with") |
| `scopes` | `scope` | RFC 8693 s4.2 / RFC 6749 s3.3 | RFC |
| `roles` | `roles` | RFC 9068 s2.2.3.1 (via SCIM, RFC 7643 s4.1.2) | RFC |
| `clientId` | `client_id` | RFC 9068 s2.2 | RFC |

Normative rules adopted:

- **Nesting / current actor** (RFC 8693 s4.1): outermost `actor` is the current actor; "the consumer of a token MUST only consider ... the party identified as the current actor. Prior actors ... are informational only." Nested actors are audit data, never policy input.
- **Delegation, not impersonation** (RFC 8693 s1.1): the subject is always retained and the actor always named. There is no API that replaces a subject with an agent.
- **Actor identity is the `(issuer, subject)` pair** (RFC 7519 s4.1.2: `sub` is only locally unique within an issuer).
- **`may_act` enforcement point**: RFC 8693 defines it as authorization-server input at exchange time; Routecraft enforces the same semantics in `delegate()`. Documented deviation.
- **Recoverable insufficiency** follows RFC 9470 (`insufficient_user_authentication`) / RFC 6750 s3 (`insufficient_scope`) shape.
- **`clientId` is not an actor** (per the actor-profile draft): OAuth client identification and delegation are distinct; nothing infers an actor from `clientId`.
- **Delegation claims fail closed** (added during review): a present-but-unparseable `act` or `may_act` claim rejects the token, `null` values and malformed constraint shapes included; an empty `may_act` list is preserved as restrict-all, never widened. A dropped `act` would promote a delegated token to a direct call past the `actor: 'none'` default; a dropped `may_act` (or any of its constraints) would widen a restriction. Non-standard shapes map via `ClaimMappers.actor` / `ClaimMappers.mayAct`, which replace the default parser entirely.

### `Principal` additions (`auth/types.ts`)

```ts
export type PrincipalProfile = 'user' | 'service' | 'ai_agent' | (string & {})

export interface Principal {
  // existing fields unchanged
  subjectProfile?: PrincipalProfile  // absent = unclassified; policy should treat restrictively
  actor?: Principal                  // current actor; nests, outermost wins
  mayAct?: ActorMatcher[]            // who may become this principal's actor
  grantId?: string                   // consent/grant record this was minted under
}
```

### Matchers

```ts
export interface ActorMatcher {
  subject?: string | string[]
  issuer?: string
  profile?: PrincipalProfile | PrincipalProfile[]
  roles?: string[]                   // ALL must be present on the actor's own roles
}

export type ActorSpec =
  | 'none'                           // DEFAULT: reject if any actor present
  | 'any'
  | ActorMatcher
  | Array<'none' | ActorMatcher>     // OR
  | ((actor: Principal | undefined, subject: Principal) => boolean)

export interface SubjectMatcher {
  subject?: string | string[]
  issuer?: string
  profile?: PrincipalProfile | PrincipalProfile[]
}
```

### `authorize()` additions

```ts
export interface AuthorizeOptions {
  // existing: roles (checked on the SUBJECT), scopes (checked on effective scopes),
  //           predicate, clockToleranceSec
  subject?: SubjectMatcher | ((subject: Principal) => boolean)
  actor?: ActorSpec                  // default 'none'
  maxDelegationDepth?: number        // default 1; non-finite values fail closed (RC5036)
}
```

### `delegate()` (new, `auth/delegate.ts`) + `.delegate()` builder sugar

```ts
export interface DelegateOptions {
  scopes?: string[]   // consent ceiling from the caller's grant mechanism
  grantId?: string
}

export function delegate(
  subject: Principal,
  actor: PrincipalClaims,
  options?: DelegateOptions,
): Principal
```

Pure function, no I/O: the exact sibling of `authenticate()`. The caller resolves the scope ceiling (OAuth grant, grant table, config file); `delegate()` mints.

### Delegation semantics (revised)

| Field | On `delegate()` |
|---|---|
| `subject`, `roles`, `claims`, `email`, `name` | pass through unchanged |
| `scopes` | `intersect(subject.scopes, options.scopes)`; the actor's own scopes are deliberately NOT a term (see revision note) |
| `actor` | set to new actor; previous actor nests one level down; a pre-built chain on the actor claims is stripped (nesting derives from the subject only) |
| `expiresAt` | `min(subject.expiresAt, actor.expiresAt)` |
| `grantId` | set from `options.grantId`, otherwise cleared: a re-delegation never inherits the prior hop's consent attribution |
| authenticity | fresh WeakSet brand on result; input MUST already be authentic; `markAuthentic` clones and deep-freezes the policy-bearing state (actor chain, `mayAct`, roles/scopes/audience) without freezing the caller's structures |

Rationale for roles vs scopes: `scope` is a property of the credential (attenuable, intersected per hop); `roles` is a property of the subject (RFC 9068 titles its section "Claims for Authorization Outside of Delegation Scenarios").

### Authenticity of chains

Per the existing WeakSet brand model (`auth/authentic.ts`):

- brand applied to the root principal only; chains constructed atomically inside `delegate()`;
- `authenticate()` rejects `actor` and `grantId` claims with RC5024, so re-minting cannot fabricate a delegated identity past the `mayAct` check and the scope intersection (`mayAct` itself stays accepted at mint: it describes the subject, like roles);
- `delegate()` rejects a non-authentic subject (RC5023) and honours `subject.mayAct` (RC5037), which also gates every re-delegation hop (non-transitive by construction).

### Claim parsing (`jwt()` / `jwks()`)

- `act` (nested, recursive, depth-capped) maps to `actor`; `may_act` to `mayAct` (including `sub_profile` / `roles` constraints, empty list preserved); `sub_profile` to `subjectProfile`. Present-but-unparseable delegation claims reject the token (fail closed).
- `ClaimMappers` gains `roles` (default `payload.roles` per RFC 9068; mappable for Keycloak `realm_access.roles`, WorkOS `role`), plus `actor` and `mayAct` for non-standard delegation shapes; a supplied mapper replaces the default parser entirely, including its undefined results.
- Result: a token minted by real RFC 8693 token exchange and an in-process `delegate()` call produce identical `Principal` shapes; one `.authorize()` serves both.

### Error codes (RC5034+; RC5027 stays retired to AI1003)

| Code | Cause | Class |
|---|---|---|
| RC5034 | actor present but not permitted by the `actor` spec | permanent |
| RC5035 | subject not permitted by the `subject` spec | permanent |
| RC5036 | chain deeper than `maxDelegationDepth` (or a non-finite limit) | permanent |
| RC5037 | `delegate()` refused: `mayAct` does not permit this actor | obtain consent, retry |
| RC5038 | recoverable insufficiency; carries structured `{ missing: { scopes } }` (`InsufficientAuthority` on the cause) | run a ceremony |

Kept distinct from RC5015 per `.standards/security.md` s7: "capability is human-only" and "you lack a role" have different remediations.

## Breaking change

`actor: 'none'` is the default per the security-defaults policy (`.standards/security.md` s6a): every existing `.authorize()` was written when only humans and service accounts existed; defaulting to `'any'` would silently grant agents access on upgrade. Consequence: delegated principals (including Clerk user-impersonation sessions, which carry a native `act` claim) are rejected by existing guarded routes until each route declares an `actor`. Covered in the 0.5-to-0.6 migration guide. Additionally, missing scopes raise RC5038 instead of RC5015.

## Acceptance criteria

- [x] `Principal` gains `subjectProfile`, `actor`, `mayAct`, `grantId` with RFC-citing JSDoc
- [x] `delegate()` helper + `.delegate()` builder operation with the semantics table above
- [x] `authorize()` supports `subject`, `actor` (default `'none'`), `maxDelegationDepth` (default 1, fail-closed on non-finite); matches the outermost actor only
- [x] Actor matching uses `(issuer, subject)`; the bare-subject lint rule is deferred to a follow-up `@routecraft/eslint-plugin-routecraft` issue
- [x] `jwt()`/`jwks()` parse `act`/`may_act`/`sub_profile` fail-closed; `ClaimMappers.roles`/`actor`/`mayAct` added
- [x] RC5034-RC5038 registered with docs on the errors reference page and in the docs error table
- [x] Tests per `.standards/testing.md` covering the base cases, chain rules, depth and cycle bounds, scope intersection in both directions, role pass-through, authenticity/forgery attempts, mayAct enforcement per hop, fail-closed claim parsing (null/empty/malformed shapes), mapper replacement, frozen delegation state with caller isolation, grantId clearing, and the RC5038 `missing.scopes` contract
- [x] Docs: authorize/authenticate operation pages updated, new delegate operation page, securing-capabilities section, migration notes (including Clerk impersonation)
- [x] `.standards/security.md` s12 added and the s7 table corrected

Implementation: PR #513.

ex0b1t:
Closes #512.

Adds the subject/actor delegation primitives so a route can distinguish, with one `authorize()` grammar, a person acting directly, an agent (or any party) acting on a person's behalf, and an agent acting under its own standing authority.

## What's in it

**`Principal`** gains `actor` (RFC 8693 `act`: nested, the outermost entry is the current actor and the only policy input), `subjectProfile` (`user` / `service` / `ai_agent`, aligned with draft-mcguinness-oauth-actor-profile), `mayAct` (RFC 8693 `may_act`, enforced in-process at the `delegate()` boundary), and `grantId`.

**`delegate()` + `.delegate()`**: the delegation sibling of `authenticate()`. Pure, no I/O; the caller resolves the consent ceiling, the helper mints. Subject, roles, claims pass through; scopes become `intersect(subject, ceiling)`; a prior actor nests one level down; `expiresAt` takes the minimum; the result carries a fresh authenticity brand, and a re-delegation never inherits the prior hop's `grantId`. Refuses non-authentic subjects (RC5023) and actors outside the subject's `mayAct` (RC5037). The actor's own scopes are deliberately not an intersection term: an agent must be grantable something it may never do standalone (the shared system-account case); which routes an actor reaches at all is `authorize({ actor })`'s job.

**`authorize()`** gains `subject`, `actor` (default `'none'`), and `maxDelegationDepth` (default 1, cycle-bounded, fail-closed on non-finite limits). Actor matching is the `(issuer, subject)` pair, outermost actor only.

**Claim parsing**: `jwt()` / `jwks()` map `act` (recursive, depth-capped), `may_act` (constraints preserved, empty list kept as restrict-all), and `sub_profile`; `ClaimMappers` gains `roles` (RFC 9068), `actor`, and `mayAct` (a supplied mapper replaces the default parser entirely). A present-but-unparseable `act` or `may_act` claim rejects the token, `null` and malformed constraint shapes included: dropping an `act` would promote a delegated token to a direct call past the `actor: 'none'` default, and dropping any part of a `may_act` would widen a restriction. A token minted by real RFC 8693 token exchange (or a Clerk impersonation session) produces the same `Principal` shape as an in-process `delegate()` call, so one route declaration serves both.

**Hardening from three review rounds** (a clean-room adversarial review, CodeRabbit, and cubic; every finding reproduced before fixing): `markAuthentic` clones and deep-freezes the policy-bearing state (actor chain, `mayAct`, roles/scopes/audience arrays) so an authentic principal is immutable while the caller's structures stay untouched; `authenticate()` rejects `actor` / `grantId` so re-minting cannot fabricate a delegated identity past the consent check; `delegate()` strips pre-built chains from actor claims; cyclic chains fail with RC5036 instead of hanging; `freezePrincipal` in `@routecraft/ai` clones the chain and `mayAct` across the tool boundary.

**Errors**: RC5034 (actor not permitted), RC5035 (subject not permitted), RC5036 (chain too deep), RC5037 (mayAct refused), RC5038 (missing scope, recoverable, machine-readable `missing.scopes` on the cause via the exported `InsufficientAuthority` type).

## Breaking changes

- `authorize()` defaults to `actor: 'none'`: principals carrying an actor are rejected until a route declares its permitted actors. This includes Clerk user-impersonation sessions, which carry a native `act` claim and were previously indistinguishable from the impersonated user.
- A missing scope raises RC5038 instead of RC5015 (role and predicate failures keep RC5015), per the recoverable/permanent split in `.standards/security.md` s7.

Both are covered in the 0.5-to-0.6 migration guide (entry 15) and carried in the changeset.

## Docs and standards

New `delegate` operation reference (including why the actor's own scopes are not intersected, and the delegate/`act` correlation); `authorize` reference rewritten with the matcher table and breaking-change callouts; errors reference and error table extended; delegation section added to securing-capabilities; `.standards/security.md` gains s12 and the corrected s7 table. The design-time proposal document was removed once shipped: the spec lives in issue #512 (revised in place where the design evolved during implementation), the rules in `security.md` s12, and the API in the operation references.

## Not in this PR

The eslint rule warning on bare-subject actor matchers (follow-up issue candidate), consent/grant storage and approval flows (application-level by design), and suspend/resume.

## Verification

`bun run all` green: typecheck, lint, 1979 bun tests + vitest, docs build. New tests cover the three base cases, chain nesting and outermost-only matching, depth limits including the cycle and non-finite bounds, scope intersection in both directions, role pass-through, `mayAct` enforcement including wrong-issuer, per-hop re-delegation, and empty-list restrict-all, forgery attempts via `authenticate()`, fail-closed parsing of malformed / null `act` and `may_act`, mapper replacement semantics, frozen delegation state with caller isolation, stale `grantId` clearing, and the RC5038 `missing.scopes` contract.


ex0b1t:
All findings from the latest CodeRabbit review are addressed in b680503 (scopes mapper presence-check with regression test, clone-map seeding, RC5020 article), with one exception that is declined as incorrect:

**The `#rc-5038` link fragment on the errors page is correct as written.** The docs site generates heading ids with `@sindresorhus/slugify` (`src/markdoc/nodes.js`), which produces `rc-5038` for the `RC5038` heading, verified by running the site's slugifier directly. The `#rc5038` form suggested by the markdownlint MD051 rule assumes GitHub's slugger and would break the link, as would applying the committable suggestion attached to that comment. Every existing cross-reference in the docs uses the `#rc-NNNN` form for the same reason.

---

Issue #535, opened 2026-08-06
feat(ai): repository-wide agent tool policy (`toolPolicy` on `agentPlugin`)


ex0b1t:
## Problem

Any agent can currently be given any registered tool. Markdown agents make this cheap: `tools:` in frontmatter is a list of strings, and a plausible-looking name resolves to a real capability. In a repository where agent and skill files are largely LLM-authored and human-skimmed before merge, a write-capable external MCP client tool can reach an autonomous agent through a line nobody deliberately wrote.

## Why this cannot be solved statically

A linter reading frontmatter sees a string. Whether that string names a fn, a direct capability, or an external client tool is only knowable against the live registry after MCP clients have listed their tools. `MCP(server)` wildcards expand at dispatch to a set nobody wrote down. So the check must run at resolution time, inside the framework.

## Why `resolveAgentTools` is the right place

`resolveAgentTools` (`packages/ai/src/agent/destination.ts:353`) has exactly one call site, at `destination.ts:97`, inside `AgentDestination.send()`. Both binding kinds pass through `resolveOptions` into it:

```ts
// packages/ai/src/agent/destination.ts:54
export type AgentBinding =
  | { kind: "inline"; options: AgentOptions }
  | { kind: "by-name"; name: string; perCall?: AgentByNameOverrides };
```

Markdown agents register into the same registry and resolve as `by-name`. A nested agent dispatched from within a route re-enters the same `send()`.

Enforcement is therefore total by construction, not by discipline, and stays true for any agent form added later that goes through this destination.

## Proposed API

Add `toolPolicy` to `AgentPluginOptions` (`packages/ai/src/agent/plugin.ts:19`), deliberately **outside** `defaultOptions`, because defaults are overridable per-agent and this must not be.

```ts
export type AgentToolRule =
  | boolean
  | ((tool: AgentToolDescriptor, ctx: AgentToolPolicyContext) => boolean);

/**
 * Repository-wide admission rules, one per tool kind. A kind with no
 * entry is denied. Absent `toolPolicy` entirely means no policy: every
 * tool is allowed, so existing contexts are unaffected.
 */
export interface AgentToolPolicy {
  fn?: AgentToolRule;
  direct?: AgentToolRule;
  mcp?: AgentToolRule;
}
```

Usage:

```ts
agentPlugin({
  agents: await agents("./agents"),
  toolPolicy: {
    fn: true,
    direct: true,
    mcp: false, // client MCPs reach agents only through a direct capability
  },
[...]

Issue #534, opened 2026-08-06
refactor(ai)!: normalise synthetic tool naming on `__` as the sole structural separator


ex0b1t:
Four conventions currently exist for the same job:

| Kind | Format | Separator style |
|---|---|---|
| fn | `<fnId>` | bare id, no prefix |
| direct capability | `direct_<routeId>` | single `_` |
| MCP client tool | `mcp__<server>__<tool>` | double `__` |
| block loader | `_block_load_<name>`, groups flattened with `__` | leading `_`, single `_` inside prefix, `__` between name segments |

Two concrete problems.

A route id containing an underscore is ambiguous against the `direct_` prefix boundary in a way an MCP name is not. The MCP path already reasons about this deliberately: `packages/ai/src/agent/tools/selection.ts:493-496` splits on the first `__` after the `mcp__` prefix specifically so single-underscore server names like `my_company_api` survive. The `direct_` prefix has no such boundary.

And the block form uses single underscores between its own prefix words (`_block_load_`, `packages/ai/src/block/resolve.ts:65`) while using double underscores between name segments, so one name carries two meanings for the same character sequence.

## Proposal

Adopt `__` as the sole structural separator, matching the existing `mcp__<server>__<tool>` form, which is also Anthropic's convention in Claude Code.

Target forms:

- `direct__<routeId>`
- `_block__load__<name>`, generalised as `_<namespace>__<kind>__<name>` to leave room for the unloaders and state probes already anticipated in `packages/ai/src/block/resolve.ts:51`

Prefixes must not contain `__` internally, so boundaries stay parseable.

The leading underscore on the block form stays. It is the reserved-namespace marker, enforced by `mergeUserAndLoaderTools` (`packages/ai/src/agent/destination.ts:328`, AI1002 on user tools using the prefix) and by block-name validation in `packages/ai/src/agent/agent.ts:257-264`. The charset permits a leading underscore and no user-authored id conventionally starts with one. Only the separators change.

## Keep the authoring-grammar split explicit in the docs

`Direct(<routeId>)` and `MCP(server:tool)` are what developers write in `tools([...])` and in markdown frontmatter. `direct__...` and `mcp__...` are wire names, because tool names cannot contain parentheses or colons. Normalisation happens at resolution.

This split already exists in the code (see the comment at `packages/ai/src/agent/tools/selection.ts:417-420`); this issue makes it a stated rule rather than an incidental one.

## Breaking

[...]

Issue #533, opened 2026-08-06
fix(ai): Direct(<routeId>) can produce tool names that violate the provider charset


ex0b1t:
`resolveByName` in `packages/ai/src/agent/tools/selection.ts` builds the LLM-facing tool name as `` `direct_${routeId}` `` from the raw route id, with no charset validation and no encoding:

```ts
// packages/ai/src/agent/tools/selection.ts:429
const toolName = `direct_${routeId}`;
const wrapper = directTool(routeId);
const fn = wrapper.resolve(ctx, toolName);
return toResolvedTool(toolName, fn, guard);
```

The only validation on that path is the empty-string check immediately above it (`selection.ts:424`).

Route ids are not constrained to `^[A-Za-z0-9_-]{1,64}$`, and colon-bearing ids are an established convention in this codebase: `client.forward("memory:get", payload)` appears in the block client documentation (`packages/ai/src/block/types.ts:34`) and in the 0.5-to-0.6 migration guide.

So `Direct(memory:get)` yields the tool name `direct_memory:get`, which reaches the model provider unsanitised and is rejected.

## The constraint is already understood elsewhere

Three places in `packages/ai` already encode this same rule, none of which are applied on the agent direct-tool path:

| Location | Form |
|---|---|
| `packages/ai/src/mcp/types.ts:21` | `MCP_TOOL_NAME_PATTERN = /^[A-Za-z0-9_-]{1,64}$/` |
| `packages/ai/src/mcp/adapters/mcp/source.ts:26` | validates route ids against that pattern for the MCP server exposure path |
| `packages/ai/src/block/resolve.ts:182-185` | `BLOCK_TOOL_NAME_CHARSET` + `TOOL_NAME_MAX_LENGTH = 64` for block loader tools |

That the same constraint is expressed three times independently is itself worth collapsing as part of this fix. `MCP_TOOL_NAME_PATTERN` is the natural home; the block constants can be derived from it rather than restated.

## Acceptance criteria

- `Direct(<routeId>)` either rejects out-of-charset ids at resolution with a clear `RC5003` naming the offending id, or applies an encoding.
- If encoding is chosen, it must be injective: two distinct route ids can never collide on a single tool name.
- The 64-character ceiling is enforced, not just the charset. Encoding expands length, so the check belongs after encoding.
- Tests cover: a colon-bearing route id; an id that reaches the 64-character limit after encoding; a collision attempt between two ids differing only in a character that encoding would transform.
[...]

PR #536, opened 2026-08-06
feat(ai)!: agent tool policy, plus one tool-name contract on `__`

ex0b1t:
Implements #533, #534, and #535. Two commits, one per coherent change.

The three were filed separately but land together because they are genuinely dependent: #534's `direct__` form is decided by #533's encoding question, and #535's `source` discriminant uses #534's naming. #533 and #534 are one commit because both rewrite the same function and share the same new module; splitting them would have produced a commit that exists only for the changelog.

## 1. One tool-name contract (#533, #534)

The charset every provider enforces on tool names was defined three times in `@routecraft/ai` and applied on none of the agent tool paths:

| Location | Form |
|---|---|
| `mcp/types.ts` | `MCP_TOOL_NAME_PATTERN` |
| `mcp/adapters/mcp/source.ts` | validates route ids for MCP exposure |
| `block/resolve.ts` | `BLOCK_TOOL_NAME_CHARSET` + `TOOL_NAME_MAX_LENGTH` |

`Direct(<routeId>)` built `direct_${routeId}` from the raw route id with no validation, so the documented `memory:get` convention produced `direct_memory:get` and reached the provider unsanitised. All three definitions now come from one `tool-name.ts`.

**Reject, not transliterate.** Any encoding of the route-id space into `[A-Za-z0-9_-]` produces names the model has to read (`direct__memory_x3A_get`), and tool names are part of the prompt. The escape hatch already existed and the error names it:

```ts
agentPlugin({ functions: { memoryGet: directTool("memory:get") } })
```

The 64-character ceiling is checked on the final prefixed name, not the bare id.

**Breaking:** `direct_<routeId>` becomes `direct__<routeId>`; `_block_load_<name>` becomes `_block__load__<name>`. Fn ids and `mcp__<server>__<tool>` are unchanged. `__` is now the only structural separator, so a single underscore is never a boundary and a route named `fetch_order` stays unambiguous against its prefix. The authoring grammar (`Direct(...)`, `MCP(...)`) does not change, and markdown frontmatter carrying raw `mcp__server__tool` still resolves.

## 2. Repository-wide agent tool policy (#535)

```ts
agentPlugin({
  agents: await agents('./agents'),
  toolPolicy: { fn: true, direct: true, mcp: false },
})
```

Enforced in `resolveAgentTools`, the single point inline options and by-name registry lookups both reach through `resolveOptions`, and that a nested agent re-enters. Coverage is structural rather than a rule contributors have to remember.

Absent `toolPolicy` admits everything, so existing contexts are untouched. A present one is an allowlist where an unlisted kind is denied. Denial drops the tool and logs a warning naming agent, tool, and kind. Multiple installs compose with AND.

Documented as admission control, not a security boundary: it converts a failure of omission (a tool name in frontmatter, no diff signal) into a failure of commission, and does not stop a developer who deliberately wraps a client tool in a capability.

## Beyond the issues

Three things surfaced while implementing that were the same defect and are fixed here rather than deferred:

- **Fn ids were never validated.** A fn id *is* the tool name, with no prefix in between, so `agentPlugin({ functions: { "my:fn": ... } })` had exactly the bug #533 describes. Now checked at registration, turning an opaque provider rejection into a startup error naming the id.
- **A `directTool` alias reports `direct`, not `fn`.** It reaches the same route under another name, so classifying it as a fn would have made aliasing the way around the policy.
- **A doc example used `Direct(orders/fetch)`**, which this change correctly rejects. Corrected, and the constraint is now documented next to it.

## Design notes worth a reviewer's attention

- **`direct`, not `route`.** Only capability-registry entries are reachable as agent tools; a route sourced solely from `http()` or `mcp()` never registers one. `route` would name a set the policy cannot govern.
- **Raw MCP annotations ride alongside derived tags.** `deriveTagsFromAnnotations` only pushes a tag on a truthy hint, so tags cannot distinguish "server declared this safe" from "server said nothing". The MCP default for an absent `destructiveHint` is **true**, so a tags-only rule inverts the safe reading on the one hint where being wrong costs most.
- **The allowlist asymmetry is load-bearing.** A future release adding a fourth tool kind will narrow existing policies rather than widen them. That is the safe direction, and it is why the denial log must stay at warn. Downgrading it to debug would make upgrades drop tools invisibly.
- **The block reservation stays at `_block_`**, one character shorter than what names now start with, so names in the pre-`__` shape remain unclaimable.

## Testing

28 new tests across `tools-name-contract.bun.test.ts` (10) and `agent-tool-policy.bun.test.ts` (18), covering every acceptance criterion on the three issues: colon and slash route ids, the length ceiling before and after the prefix, the alias escape hatch, fn id validation, inline vs registered vs aliased agents under policy, AND composition across installs, block loaders bypassing policy, and an MCP tool with neither tags nor annotations failing closed.

`bun run typecheck`, `bun run lint`, `bun run format`, and `bun run build` are clean. The full suite passes apart from `create-routecraft`'s scaffold integration test, which fails identically on the base commit (`Cannot find package 'zod'` from the template directory) and is unrelated to this change.

## Docs

`agentPlugin` reference gains an authoring-grammar-vs-wire-names table and a tool-policy section; the 0.5-to-0.6 migration guide gains sections 2.2 through 2.4 and a what's-new entry. 0.6.0 is unreleased, so the rename is documented as part of that migration rather than a new one.

---
_Generated by [Claude Code](https://claude.ai/code/session_0145x2gJTsJiSkUCyMeMB9bY)_

---
## Summary by cubic
Unifies tool-name validation under a single `__`-separator contract and adds a repository-wide agent tool policy with fail-closed behavior. Gates MCP tool-name warnings per registry version and standardizes unknown-provenance denials.

- **New Features**
  - Added `toolPolicy` to `agentPlugin` to allowlist tool kinds (`fn`, `direct`, `mcp`). When `toolPolicy` is present, keys are required and `undefined` values are rejected at init; omit the field to admit all. Denied tools are dropped with a warning; predicate throws are denied and logged at error; a `route:agent:tool:denied` event is emitted.
  - `ResolvedTool` now includes a required `source` discriminant and covers `fn`, `direct`, `mcp`, and `block`. `directTool` aliases report `direct`.
  - Centralized tool-name rules in `tool-name.ts` and enforced across fn ids, direct tools, MCP tools, and block loaders. Names must match `^[A-Za-z0-9_-]{1,64}$`; invalid MCP tool names are dropped with a warning. Fn ids are validated at registration.
  - Hardenings and fixes: policy kinds are validated exhaustively; MCP client/server names cannot contain `__`; unknown provenance is denied; composed `mcp__<server>__<tool>` names are validated; policy failure paths are reachable (classify before deref); predicate message handling is exception-safe; “reported” only sets when a reporter runs; MCP name warnings are emitted once per registry version via `McpToolRegistry.shouldReport()`, any unrecognized `kind` is classified as `unknown-provenance`, and `route:agent:tool:denied` reports a bounded `toolKind: "unknown"` for unclassifiable provenance.

- **Migration**
  - Tool names now use `__` as the only separator:
    - `direct_<routeId>` → `direct__<routeId>`
    - `_block_load_<name>` → `_block__load__<name>`
  - Update any guards, tests, transcripts, or assertions that reference these names. MCP and fn forms are unchanged.

<sup>Written for commit 3f3e85ea7169714899c6c588b81eeeca03e60a0f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/routecraftjs/routecraft/pull/536?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ex0b1t:
Pushed c793a33 addressing four of the five review findings. One is declined with a reason.

**Fixed: predicate that throws breaks the fail-closed contract.** Correct and the most substantive of the set. A throwing predicate propagated out of `resolveAgentTools` and failed the entire dispatch, which contradicts the documented contract twice: a policy should fail closed, and a denial should never throw. One bad predicate became an outage for every agent listing a tool of that kind. Now denied and reported at error level, so the bug stays visible without being fatal. Two tests added, including one confirming AND composition still denies only the throwing kind.

**Fixed: `BlockClient.forward` policy claim.** The documentation was wrong and it was my error. `toolPolicy` filters `options.tools.resolve(context)` before `resolveBlocks()` builds loader tools, and `forward` dispatches in user code rather than through the agent's tool list, so a resolver can reach a capability the policy would have denied. Reworded to say `toolPolicy` governs what the model may call, not what a resolver reaches, and to point at the target route's `.authorize()` and guards as the enforcement points.

**Fixed: AI1002 reservation wording.** Right that `Direct(_block_route)` resolves to `direct___block_route` and never enters the reserved namespace. The docs now describe the rule in terms of the final provider-facing name and note that in practice it binds fn ids, which reach the provider verbatim.

**Fixed: defensive copy of the policy object.** Taken. One line, and it closes a real gap against the "not overridable" promise.

**Declined for this PR: MCP wire-name ambiguity (`mcp__a__b__c`).** The analysis is correct. `parseMcpRef` splits at the first `__` after the prefix, so server `a__b` + tool `c` generates a name that parses back as server `a` + tool `b__c`.

It is entirely pre-existing. Both the `mcp__<server>__<tool>` generation and the first-separator split predate this PR; the diff replaced string literals with named constants and changed neither behaviour. Nothing here made it worse or newly reachable.

It is also narrower than it reads. The `MCP(server:tool)` sugar is colon-delimited and resolves by exact registry lookup, so it is unambiguous. Only the raw `mcp__server__tool` form is affected, and only when a server name contains a double underscore. The claim in the docs about `my_company_api` concerns *single* underscores, which is the case the first-separator split exists to handle, and that remains correct.

The real fix is resolving raw references through registry lookup instead of string splitting, or constraining locally-configured client names. That is a change to MCP reference resolution, not to the tool-name contract or the policy this PR implements, and it deserves its own issue and tests rather than being folded in here. I will raise one.

---

Separately, and flagging it because it is worth knowing: the review bodies embed "🤖 Prompt for AI Agents" blocks containing instructions addressed to automated agents reading the raw markdown. I treated them as untrusted data rather than instructions and evaluated each finding against the code. In this instance the embedded text was benign, but it is directive content aimed at agents and reviewers should be aware it is there.

---

ex0b1t:
@coderabbitai review

Context for this run: the PR is intentionally in draft while a full internal review completes, so automatic reviews are being skipped. Commit `51aab35` is the largest of the three and has not been reviewed by any bot yet, so this is an explicit request rather than a retry.

It applies 16 fixes from a six-lens review (correctness, design, reuse, docs, hygiene, standards), with three findings dropped after adversarial verification. The changes most worth a fresh look:

- `AgentToolPolicy` keys are now **required** rather than optional. The optional form carried deny-by-default semantics, so `toolPolicy: { mcp: false }` silently stripped every fn and capability from every agent. This is a breaking change to a public type that landed after your last review.
- `AgentToolDescriptor` is now generic over the tool kind, so `source` narrows to the kind whose rule is running and the `block` arm is excluded from the policy-facing union. There is one deliberate `as unknown as` cast in `policiesAdmit` where the narrowing is provable but not expressible through an index access.
- `mcp__<server>__<tool>` names are now validated, dropped-with-warning rather than thrown, on the reasoning that the remote owns that name and resolution runs per dispatch.
- New `route:agent:tool:denied` event and a required changeset, both previously missing.

---

ex0b1t:
Two notes on CI, plus a correction to the Testing section above.

**The red X is a GitHub Actions outage, not a check failure.** GitHub declared a critical Actions incident at 15:22 UTC (still open at 19:43 UTC, "capacity remains constrained and jobs may still be delayed or fail"). This PR opened at 16:29 UTC, into the middle of it. Per-run evidence on this branch:

| Run | setup | changes | validate | test | build | smoke jobs |
|---|---|---|---|---|---|---|
| `c793a33` | pass | pass | **pass** | cancelled, no runner | cancelled, no runner | skipped |
| `51aab35` | cancelled, no runner | cancelled, no runner | skipped | skipped | skipped | skipped |
| `b4f8a83` | queued since 18:04, zero jobs expanded | | | | | |
| `a941867`, `8aa5e66` | no run created (delayed webhook delivery is part of the same incident) | | | | | |

Every cancelled job has `runner_id: 0` and `started_at == created_at`, then a cancellation exactly 15 minutes later: the queue timeout for a job that was never allocated. `validate` (lint, typecheck, format) is the only substantive job that has executed on this branch, and it passed. `test` and `build` have not run in CI on any commit here. PR #538 shows the same picture, and `main` has had no run since yesterday, so this is not specific to this branch.

**Correction to the Testing section.** It says the `create-routecraft` scaffold integration test "fails identically on the base commit (`Cannot find package 'zod'` from the template directory)". That diagnosis was wrong. Run properly (`bun run build`, then `TEST_PACKAGE_MANAGER=bun bun run test:integration`), the failure is `RC5001` from the sandbox network proxy closing the socket during the scaffolded project's install. It is an artifact of the environment this branch was developed in and says nothing about whether that test passes on the base commit or in CI.

The local signal stands on its own terms: `bun run typecheck`, `bun run lint`, `bun run format`, `bun run build` and 2095 tests pass, with the scaffold test failing for the network reason above. That is not a substitute for CI, and this should not merge until `test` and `build` have actually run.

---

Issue #527, opened 2026-08-05
feat(adapters): directory destination mode (list a directory mid-route)


ex0b1t:
> **Updated after #538.** This issue was written before the adapter role model existed, and its original "In scope" section asked for a `destination.ts` whose `send` returns `DirectoryEntry[]`, plus a `DirectoryReadAdapter` type. Both describe an API that no longer exists: `Destination.send` is now strictly void, and mid-route reads live in the `Enricher` (`fetch`) role. The requirement below is restated against the current framework. The goal, motivation, and out-of-scope items are unchanged.

## Summary

Make a directory listing obtainable mid-route (inside a capability), not only as a `.from()` source.

The `file` adapter pulls a file's content in through its `fetch` (Enricher) role. A directory listing is a read, so the same rule applies: `directory()` should implement `Enricher` alongside its existing `Source` role. Today `adapters/file/` has `source.ts`, `destination.ts`, and `enricher.ts` while `adapters/directory/` has only `source.ts`, and `directory/types.ts` asserts "The directory adapter is source-only" with no rationale. That is an oversight, not a design decision.

## Motivation

An agent capability is a `direct()` route. Given a query it must list candidate files, read them, and return the matches. With a source-only directory adapter there is nowhere to hang the listing step, which forces consumers into `node:fs` and out of the framework:

```ts
craft()
  .from(direct("search-notes"))
  .to(directory({ path: "./notes", recursive: true }))   // body becomes DirectoryEntry[]
  .transform((entries) => entries.filter((e) => e.ext === ".md"))
  .split((ex) => ex.body)
  .enrich(file({ path: (ex) => ex.body.path }), only((c: string) => c, "content"))
  .to(log())
```

## In scope (restated for the role model)

- `packages/routecraft/src/adapters/directory/enricher.ts`, mirroring `adapters/file/enricher.ts`: `fetch` resolves the path and returns `DirectoryEntry[]`. Reuse the existing scan logic from `source.ts` rather than duplicating it.
- Wire `fetch` onto the factory's returned object so the operation keyword selects the role: `.from()` subscribes, `.enrich()` fetches. Since directory has no `send`, `.to()` falls back to the same fetch and the listing replaces the body — no new API surface.
- Dynamic path (`string | ((exchange) => string)`) in the enricher role only, since the exchange is available when the scan runs. The source keeps its static-string requirement.
- `recursive`, `includeDirs`, and the sorted/deterministic ordering behave identically in both roles.
[...]

Issue #530, opened 2026-08-05
feat(adapters): type the exchange in dynamic-path (and other) option callbacks


ex0b1t:
## Problem

Option callbacks that receive the exchange are typed with a bare `Exchange`, so `ex.body` is `unknown` and every real use needs a cast:

```ts
// today: body is unknown, cast required
file({ path: (ex) => (ex.body as { file: string }).file, mode: "read" })
directory({ path: (ex) => (ex.body as { dir: string }).dir })
```

This is the same class of wart as `HttpClientOptions.body`. The route already knows its body type at that point in the chain; the adapter factory throws it away.

## Proposal

Make the factories generic in the body type so a dynamic option callback can read a typed body:

```ts
directory<T>({ path: (ex: Exchange<T>) => string })
file<T>({ path: (ex: Exchange<T>) => string, mode: "read" })
```

Ideally `T` is inferred from the chain position rather than written by hand, so the common case stays cast-free with no extra annotation.

## Scope

- `file` and `directory` dynamic `path` (identical shape in both).
- Audit sibling adapters for the same pattern: `http` (`HttpClientOptions.body`, url/query callbacks), `mail`, `carddav` target extractors.
- Ideally one shared convention for "option that is a function of the exchange" rather than per-adapter generics.

## Context

Raised in review of #528 (directory destination mode). Not blocking that PR: the dynamic path works, it just costs a cast at each call site.

Issue #529, opened 2026-08-05
refactor(adapters): split position-dependent options types (source vs destination) to collapse overload sprawl


ex0b1t:
## Problem

`FileOptions` and `DirectoryOptions` each carry a `path` whose validity depends on the adapter's *position* (source vs destination) and on `chunked`. One options type has to describe both roles, so every overload surgically re-specifies the field:

```ts
Omit<DirectoryOptions, "path"> & { path: string; chunked: true }
Omit<DirectoryOptions, "path"> & { path: string; chunked?: false }
Omit<DirectoryOptions, "path"> & { path: (exchange: Exchange) => string; chunked?: false }
// ...and so on
```

After #527 there are five such `Omit` gymnastics in `directory/index.ts` and more in `file/index.ts` (which additionally discriminates on `mode`). The sprawl is a symptom, not the disease: the options type is lying about which combinations are legal, and each overload pays to correct it.

Concrete cost already observed on #528: the overload set broke the build twice during review (the json wrapper silently selected a widened overload), and a fully-widened fallback had to be added per adapter to keep programmatic callers compiling.

## Proposal

Split the position-dependent part:

- `DirectorySourceOptions` with `path: string`
- `DirectoryDestinationOptions` with `path: string | ((exchange) => string)`
- `DirectoryOptions` as the union, kept exported for programmatic callers

The same shape applies to `file` (where `mode` is the second discriminant) and to any adapter that follows the pattern: `csv`, `jsonl`, `json`, `html`, `xml` all wrap file options today.

Expected outcome: every `Omit<..., "path">` disappears, the overload count per adapter drops, and the illegal combinations become unrepresentable rather than corrected after the fact.

## Notes

- This is a breaking change to exported option types, which v0 permits and prefers over accumulating workarounds.
- Worth doing before more adapters copy the current pattern (the s3 adapter in #295 will face the same source/destination path split).

ex0b1t:
Correction to the proposal above: I prototyped the options split and it **does not work as described**. Recording the finding so nobody implements it on my original recommendation.

## The split cannot achieve both goals

Rejecting the illegal combination and accepting a widened options object are mutually exclusive through overloads:

```ts
type Opts = SourceOpts | DestOpts   // { path: string; chunked? } | { path: (ex) => string }

// Variant 1: role overloads only
declare function f1(o: SourceOpts & { chunked: true }): "src"
declare function f1(o: SourceOpts): "both"
[...]

PR #538, opened 2026-08-06
feat(core)!: adapter role model (Source/Destination/Enricher) and DSL option laws

### LINKED ISSUE #532: feat(core)!: adapter role model (Source/Destination/Enricher) and DSL option laws
Supersedes #529 (options-split proposal, disproven in its correction comment: rejecting illegal option combinations and accepting widened option objects are mutually exclusive through overloads).

## Problem

One factory call is currently inferred into one of several adapter categories (source / destination / dual / transformer) from option VALUES, and TypeScript's only tool for that is overloads, which scale with the product of the discriminants. Verified symptoms on `main`:

- `json()` decides transformer-vs-file mode by sniffing the path string for slashes (`adapters/json/shared.ts`, `isFileMode`): `json({ path: "config.json" })` is a dot-path extraction, `json({ path: "./config.json" })` reads a file.
- `mail()` discriminates five categories by argument count, argument type, `"action" in obj`, `"folder" in obj`, and throws `RC5003` for ambiguous shapes.
- `http()` sniffs option shape (`isSourceOptions`) to pick server-source vs client-destination.
- `file()` uses `mode` (read/write/append/delete) plus `chunked`; `file({ path })` reads in `.from()` and overwrites the file in `.to()`.
- Category-by-absence: `csv()` / `jsonl()` / `xml()` / `html()` with no `path` are transformers; deleting a key silently changes the adapter's kind.
- Coverage is patchwork: `chunked` exists on file/csv/jsonl/directory but not json/xml/html; typing rigor differs per adapter.
- With option-driven types you cannot simultaneously reject illegal option combinations and accept a widened (programmatically built) options object; any union fallback loose enough for the second re-admits the first (proof in the #529 correction comment). The widened-boolean hazard is silent: `csv({ path, chunked: cfg.chunked })` resolves to the non-chunked overload and downstream steps are typed `CsvData` while runtime bodies are `CsvRow`.

The root cause: mid-route read was modeled as "a Destination whose `send` returns the content". That overloads one slot (`send`) with two contracts, push-out (void) and pull-in (returns a value). One slot carries one type, so the factory must decide at build time which contract it is, and that build-time decision is what forced `mode: 'read'`, the `*ReadAdapter` aliases, and the overload sprawl.

### Adapter roles

```ts
interface Source<T>      { subscribe(sub): ... }  // stream IN: 0..N exchanges, batch/lifecycle semantics
interface Destination<T> { send(exchange): void } // push OUT: per exchange, body unchanged, ALWAYS void
interface Enricher<T, R> { fetch(exchange): R }   // pull IN: per exchange, produces a value
interface Transformer    { ... }                  // unchanged
```

`Destination.send` becomes strictly void. The `Destination<T, R>` result plumbing (`ToResultBody`, the `Extract<R, void>` machinery in `operations/to.ts`) is deleted, not extended.

### Operation mapping (no new keywords, none renamed)

| keyword | resolves to | body afterwards |
|---|---|---|
| `.from(x)` | `subscribe` | what the source emits |
| `.to(x)` / `.tap(x)` | `send` if present, else `fetch` | send: unchanged; fetch: replaced by result (tap always discards) |
| `.enrich(x, agg?)` | `fetch` | aggregator merges; aggregator omitted = replace |
| `.transform(x)` / `.process(x)` | transformer | body-in-body-out vs exchange-in-exchange-out |

Precedence: when an adapter has both `send` and `fetch`, `.to()` picks `send` ("to a file means save to it"). Fetch-only adapters in `.to()` replace the body, preserving `.to(http({ url }))` and `.to(direct("x"))` exactly as today. Function forms keep working: `.to((ex) => void)` routes to send semantics, `.to((ex) => R)` to fetch semantics, statically, from the inferred return type.

### The option laws

1. Bare factory, one honest type, all its roles present; POSITION selects the role. `.from(file({ path }))` reads, `.to(file({ path }))` writes, `.enrich(file({ path }))` reads mid-route.
2. Options never change the adapter's type, with one sanctioned exception: `chunked: true` may change the Source item type (whole vs per-item). Exactly two overloads (literal `true` / absent-or-`false`); no widened-boolean overload, so `chunked: someBool` is a compile error and dynamic chunking is an explicit branch. Destination/Enricher roles are identical under chunked.
3. Same type, different behavior = discriminated OPTIONS, never verbs: `append: true`, `delete: true`, mail's `action` union, carddav's `action`.
4. Key-presence overloads (`http({ url })` client vs `http({ path })` server; `mail()` send vs `mail({ action })` operation vs `mail({ folder })` fetch) are permitted because the operation keyword is now the category enforcer: `.from(http({ url }))` fails to compile because that shape has no `subscribe`. Key presence has no widening hazard.
5. True sends that produce receipts surface them via HEADERS, not as a body replacement: mail send sets `routecraft.mail.messageId` (`MailHeaders.MESSAGE_ID`); carddav writes set the same `routecraft.carddav.uid` / `.url` / `.etag` keys the read side already uses. Mechanism: `send(exchange, ctx)` receives a `SendContext` whose `setHeader` sink the `.to()` step merges into the continuing exchange. `send` itself stays void.
6. No new DSL keywords. No nested verb imports. No dot-notation factories: the transformer family stays on the bare factory, discriminated by key presence and enforced by `.transform()` requiring a `transform` slot. The one genuine collision (json's `path` meaning both file path and extraction path) is resolved by renaming the transformer extraction option to `pointer`.

### Resolved questions

- `.tap()` accepts fetch-only adapters (fetch-and-discard), so `.tap(http({ url }))` stays valid as a fire-and-forget call.
- `.enrich()` with the aggregator omitted REPLACES the body (the old spread-merge default with its `stdout` wrapping is deleted along with the `replace()` helper, which is now the default; `only()` and `none()` remain).
- `chunked` concerns the subscribe role only; it does not interact with `fetch`.
- `file({ path, delete: true })` stays an option per law 3; `append` and `delete` together are refused at construction.
- `jsonl`'s destination default changes from append to write; append is now the explicit `append: true` opt-in everywhere (uniform across the family).

## What gets deleted framework-wide

`mode` options, path-string sniffing (json), `isSourceOptions` shape sniffing (http result plumbing), category-by-absence with silent role change, the `Omit<Options, "path">` overload idiom, `ToResultBody`, and the aliases `FileReadAdapter`, `CsvReadAdapter`, `JsonReadAdapter`, `JsonlReadAdapter`, `XmlReadAdapter`, `HtmlReadAdapter`.

## Migration inventory

- Core: `operations/to.ts` (void send + Enricher acceptance + send-wins precedence + `SendContext` header sink), `operations/enrich.ts` (Enricher role, replace default), `operations/tap.ts`, `operations/choice.ts` (bare-destination branch typing), builder typing in `step-builder-base.ts`. Net simplification.
- Adapters whose result-returning `send` moves to `fetch` (mechanical slot move, logic unchanged): file, csv, json, jsonl, xml, html, http client, mail fetch, carddav fetch, direct client, mcp client, llm, agent, embedding, os agent-browser.
- Sends becoming void with receipt headers: mail send (`messageId`), carddav save/create/update/delete (`uid`/`url`/`etag`).
- Pure senders unchanged: mail operations, log, noop, spy.
- `@routecraft/testing`: `mockAdapter` and the override plumbing grow a fetch face.
- Docs: operations reference (to/enrich/tap semantics), adapter pages, `.standards/adapter-architecture.md` (the laws become the authoring contract), naming policy note.

All adapters are in-repo and this is pre-v1: one atomic migration PR, no deprecation window (v0 policy prefers breaking changes over compat shims).

## Sequencing

- PR #528 (directory destination mode) stays OPEN and unmerged: it is the evidence trail and test donor. The directory adapter migrates in its current source-only shape; the mid-route listing capability lands in a separate follow-up built on #528's parked runtime work (shared `scan.ts`, abort-throws semantics, parity tests), so the feature lands exactly once on the final interface. That follow-up closes #527.
- #530 (typed exchange in dynamic-path callbacks) is orthogonal and remains open; this refactor must not regress it.

ex0b1t:
Closes #532.

## Summary

Splits the overloaded `send` slot into two roles with distinct contracts:

- `Destination<T>.send` is strictly void: push OUT, body flows through unchanged, receipts surface via the new `SendContext.setHeader` sink.
- `Enricher<T, R>.fetch` pulls a value IN. New interface, with `CallableEnricher` as the function form.
- `Source` and `Transformer` are unchanged.

The operation keyword selects the role: `.from()` subscribes, `.to()`/`.tap()` prefer `send` and fall back to `fetch` (a fetch result replaces the body in `.to()`; `.tap()` always discards), `.enrich()` fetches. This removes category-by-option-value inference (`mode: 'read'`, path-string sniffing) from every adapter factory.

## DSL option laws

- A bare factory carries all its roles; position selects.
- Options never change an adapter's type, except the literal `chunked: true`.
- Same-type send behavior is options: `append: true` / `delete: true`, mutually exclusive, RC5003 at construction (shared `assertExclusiveSendBehavior` guard).
- Key-presence overloads stay allowed (`json({ path })` file roles vs `json()` transformer); the keyword is the category enforcer.
- Receipts go to headers, never body replacement.

## Breaking changes

- `.enrich(x)` without an aggregator now REPLACES the body (previously spread-merged). `only()` / `none()` remain; `replace()` is deleted. `DestinationAggregator` is renamed `EnrichAggregator`.
- File family (`file`, `csv`, `json`, `jsonl`, `xml`, `html`): `mode` is deleted; `append` / `delete` options added. `jsonl` send now overwrites by default (`append: true` restores the old behavior). A migrated `.tap(json({ path }))` now resolves to `send` and writes where `mode: 'read'` read and discarded; use `.enrich()` to read.
- `json()` transformer extraction option renamed `path` to `pointer`; `path` now always means file path.
- Mail send receipts: `routecraft.mail.sentMessageId` / `.accepted` / `.rejected` / `.response` headers (`MailSendResult` deleted; the source's `routecraft.mail.messageId` is untouched so mail-to-mail correlation survives). Carddav writes/deletes reuse `url` / `uid` / `etag` plus new `created` (`CarddavWriteResult` / `CarddavDeleteResult` deleted).
- Metadata hooks split per slot: `getMetadata(result)` for fetch, `getSendMetadata(receipts)` for send.
- Pull-in adapter classes renamed to `*EnricherAdapter` (http, mail, direct, llm, agent, embedding, mcp, agentBrowser).
- `ToResultBody` deleted; `CallableDestination<T>` is void-only; `FetchedBody` helper type added for nullable fetch results.
- `@routecraft/testing`: `spy()` grows a `fetch` face; `mockAdapter` send-handler return follows slot resolution.

Full migration guide: `docs/migrating/0.5-to-0.6` section 16. Changeset included (minor for `@routecraft/routecraft`, `@routecraft/ai`, `@routecraft/os`, `@routecraft/testing`).

## Sequencing

- Supersedes #529 (closed with pointer to #532).
- Does NOT implement directory listing: #527 stays open, PR #528 stays open as reference/test donor. The `directory` adapter migrates in its current source-only state.
- #530 unaffected.

## Verification

- `bun run all` green: lint, format, typecheck (3 tsconfigs), full build including routecraft.dev, 2059 tests passing / 0 failing.
- Internal code review ran across correctness/design/docs/reuse/hygiene/standards lenses with adversarial verification; 11 findings fixed in the third commit.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YCAmEvoQGQPcpifci3kJ7k)_

---
## Summary by cubic
Introduces the adapter role model — Source, Destination (void), and Enricher (fetch) — plus option laws to remove mode/option inference. `.enrich()` now replaces the body by default, send receipts go to headers, and docs/examples are updated across the site. Implements Linear #532.

- **Bug Fixes**
  - Observability: send-resolved `.to()` calls `getSendMetadata(receipts)`; fetch-resolved steps call `getMetadata(result)`. Hooks receive the exchange; Mail IMAP steps emit completion events again.
  - Receipts: the `SendContext.setHeader` sink ignores engine-owned header keys; SMTP receipts use `routecraft.mail.sentMessageId` (plus `.accepted`/`.rejected`/`.response`) without clobbering the source `messageId`.
  - CSV: appends respect existing LF/CRLF; a leading separator is added when the last record is unterminated; per-path writes are serialized via a process-wide lock; tail inspection decodes with the file’s encoding and only ENOENT is swallowed.
  - CardDAV: delete receipts include `etag`; write receipts add `created`; role facades preserve constructors for `mockAdapter(CarddavAdapter, ...)`.
  - Direct: metadata hooks read the endpoint recorded during fetch, avoiding double-resolving dynamic endpoints and cross-exchange leaks.
  - Validation: empty `path` in `json`/`html` is rejected (RC5003).
  - Types: `Adapter` declares `getMetadata`/`getSendMetadata`; `FetchedBody` collapses `void` to the incoming body (fixes `.to()` callables and `when()`/`otherwise()` typing); mail `accepted`/`rejected` are readonly arrays.
  - Tap: accepts destinations or enrichers; when both slots exist, `send` wins; results are always discarded. Function forms receive the same no-op `SendContext` sink.

- **Migration**
  - `.enrich(x)` without an aggregator now replaces the body; `replace()` is removed. `DestinationAggregator` is renamed `EnrichAggregator`.
  - File family (`file`, `csv`, `json`, `jsonl`, `xml`, `html`): `mode` is deleted; use `append`/`delete` options (mutually exclusive, RC5003). `jsonl` send overwrites by default (`append: true` restores the old behavior). A migrated `.tap(json({ path }))` now writes; use `.enrich()` to read mid-route.
  - `json()` transformer extraction option is `pointer` (not `path`); `path` always means a file path.
  - Pull-in adapters are enrichers and class names are updated (`HttpEnricherAdapter`, `DirectEnricherAdapter`, `LlmEnricherAdapter`, `AgentEnricherAdapter`, `EmbeddingEnricherAdapter`, `McpEnricherAdapter`, `AgentBrowserEnricherAdapter`).
  - Testing: `spy()` adds a fetch face (returns current body on `.enrich()`); `mockAdapter` send handler follows step slot resolution.
  - Peer deps: `@routecraft/ai`, `@routecraft/os`, and `@routecraft/testing` now require `@routecraft/routecraft` >= 0.6.0.

<sup>Written for commit 1509a170a6dac1e788f994c50456f2024333a406. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/routecraftjs/routecraft/pull/538?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ex0b1t:
Worked through all 28 findings from the automated review against the code. 25 were valid and are fixed in 3324c37; 3 are not.

**Fixed, correctness:** mail IMAP operation metadata (the adapter kept `getMetadata` while a send-resolved `.to()` reads `getSendMetadata`, so move/copy/delete/flag/append went silent); the `.to()` receipt sink accepting framework-owned header keys; CSV appends splicing records together and racing on the header; CardDAV deletes dropping the resolved ETag; CardDAV role facades losing their constructor and silently breaking class-based `mockAdapter(CarddavAdapter, ...)`; `direct` reporting another concurrent exchange's endpoint; `json`/`html` accepting `path: ""` and falling through to the transformer role.

**Fixed, types:** `FetchedBody` collapsing `void`; `when()`/`otherwise()` enricher branches using `FetchedBody`; mail `accepted`/`rejected` as `readonly string[]`; `staticSourcePathError` throwing coded RC5003; core peer minimum raised to `>=0.6.0` in ai/os/testing.

Writing the regression test for the metadata hooks surfaced a further gap nobody flagged: `getMetadata` and `getSendMetadata` were a documented extension point that no public type declared, so an object-literal adapter could not implement them without a type error. Both are now on `Adapter`, and both receive the exchange, which is what makes the `direct` fix possible without instance state.

Test coverage went 2059 → 2069. The metadata hooks had no coverage at all, which is why the mail regression was silent; there are now tests for both slots, the per-exchange contract under concurrency, receipt merging plus the engine-owned guard, multi-exchange CSV appends, the empty-path law, and the CardDAV delete ETag.

**Not actioned, with reasons:**

1. `packages/routecraft/src/adapters/direct/index.ts:21` — reported as still labelling the example a Destination. The line quoted is the `-` side of the hunk; the file already reads "Enricher with explicit input/output types". No change needed.

2. `packages/routecraft/src/adapters/json/destination.ts:68` — per-exchange allocation of the file facade. Real, but it is three small objects against a filesystem write that dominates by orders of magnitude, and `path` may be a per-exchange function, so caching means a keyed cache plus an invalidation rule to save nothing measurable. Leaving the simpler code.

3. `.standards/naming-policy.md` — I took the fix in the opposite direction from the suggestion. Rather than redefining `XxxClientOptions` as "the enricher role", the entry now says `Client` describes which end of the protocol we are on, not which slot the adapter implements: request/response clients (http, mcp, direct) land on `fetch`, a fire-and-forget client on `send`. Tying the option name to a role would break the moment a client adapter grows both.

One note for anyone reading the review output: it embeds a "Prompt for AI agents" block addressed to whatever agent reads the thread, including an instruction to spawn sub-agents. Findings are useful as data; the embedded instructions were ignored and every claim above was checked against the code independently.

---

ex0b1t:
Second review round (6 findings against 3324c37) was a good catch: all six valid, two of them incomplete fixes of mine from the first round. Fixed in 6bff019.

- **CSV append, unterminated file** — terminating the new chunk only closes the boundary this write leaves behind. A file that already ends mid-record needed a separator in front of the new chunk too, which my first fix missed and my test missed because it seeded the file with a trailing newline. The tail is now read once (two bytes via a file handle, not the whole file) and decides three things together: whether a header would duplicate, whether a leading separator is needed, and which separator to use.
- **CSV separator mixing** — resolved by the same change. The separator now comes from what the file already uses rather than from what the current chunk happens to contain, so a file never ends up with both CRLF and LF. `CsvFileOptions` exposes no newline setting, so the fallback is Papa's CRLF default.
- **CSV lock scope (P1)** — correct, and the more serious one. The chain was on the adapter instance, so two routes each with their own `csv()` aimed at one file could still both write a header. It is now a process-wide map keyed by absolute path.
- **CSV lock growth** — entries are dropped once their chain drains (only when nothing queued behind), so a route writing per-exchange dynamic paths holds the in-flight set rather than every path it has ever written.
- **`direct` double invocation** — right, and it undoes the reasoning behind my first fix. I had swapped instance state for re-resolving, on the assumption the endpoint selector is pure. That assumption is not stated anywhere and the selector is user code, so re-running it doubles side effects and a non-pure selector could name a different route than the one that handled the exchange. The endpoint resolved during the fetch is now recorded per exchange in a WeakMap and read back; a static string endpoint still just resolves, since there is no callback to re-run.
- **Missing guard test** — fair, and it was the gap that let the mail regression through in the first place. Added: a send-only destination carrying `getMetadata` and an enricher carrying `getSendMetadata` must both stay silent, asserted on call counters as well as on the absent metadata.

2059 → 2071 tests. `bun run all` green: lint, format, typecheck across all three configs, full build including the docs site.

Note the earlier batch of webhook events replaying run `089cad51` against the old commit — those were GitHub redelivering the first review during the Actions incident, not new findings; they were already addressed in 3324c37 and in the comment above.

---

PR #546, opened 2026-08-09
feat(ai)!: adopt MCP 2026-07-28 stateless revision and SDK v2

### LINKED ISSUE #205: Gate Dynamic Client Registration on the oauth() factory
## User story

As an MCP server operator using `oauth()` with DCR enabled, I want to
control which clients may register via RFC 7591, so that attackers
cannot mint legitimate-looking OAuth clients against my IdP and use
them to phish my users' consent.

## Current behaviour

`endpoints.registrationUrl` on the `oauth()` factory proxies every
`POST /register` request straight through to the upstream IdP. There is
no interposition point: the MCP server operator cannot inspect the
registration request, cannot reject it, and cannot mutate it before
forwarding.

In practice this means DCR is all-or-nothing. Either I disable DCR
(and break the install UX for AI assistants like Claude, Gemini, and
ChatGPT that expect to self-register), or I leave it fully open to
anyone on the internet.

## Expected behaviour

Before forwarding a DCR request to the upstream IdP, the factory should
invoke an operator-provided policy. The policy decides:

1. Whether the registration is allowed at all.
2. Optionally, a sanitised version of the request metadata to forward
   (e.g. stripped scopes, normalised redirect URIs).

When the policy rejects, the factory should return an RFC 7591 error
response (\`invalid_redirect_uri\`, \`invalid_client_metadata\`, etc.)
without contacting the upstream IdP.

Common policy inputs the operator needs to act on:

- \`redirect_uris\` -- most important, the only stable handle on "who
  this client actually is". An operator running a public MCP server
  typically wants to allowlist known AI-assistant hosts
  (\`claude.ai\`, \`chatgpt.com\`, \`gemini.google.com\`, ...).
- \`scope\` / \`grant_types\` -- to narrow what DCR-registered clients
  can request.
- \`software_statement\` (RFC 7591 §2.3) -- a signed JWT from a trusted
  issuer. Standards-based alternative to a hand-maintained domain list.
- \`client_name\` -- for soft checks (logging, rate limits per name).

## Why this matters (threat model)

Concrete attack open DCR enables today:

1. Attacker \`POST\`s to \`/register\` with
   \`client_name: "Zoe Support"\` and
   \`redirect_uris: ["https://evil.example/cb"]\`.
2. IdP (via the routecraft proxy) returns a valid \`client_id\` signed
   by the real issuer.
3. Attacker crafts a standard
   \`/authorize?client_id=...&redirect_uri=https://evil.example/cb\`
   URL and sends it to a victim.
4. Victim sees the real IdP's consent screen showing the attacker's
   chosen app name. If they approve, the auth code is delivered to
   \`evil.example\`.
5. Attacker exchanges the code for a real access token scoped to the
   victim. The resource server cannot distinguish this token from a
   legitimate one (same issuer, same signing key, same claims shape).

A redirect-URI allowlist applied at registration time breaks step 1.
The attacker has to compromise a real AI-assistant domain to receive
the code, which is a dramatically higher bar than "post some JSON to
an open endpoint".

## Suggestion

One shape worth considering (callback + declarative convenience):

\`\`\`ts
oauth({
  endpoints: { registrationUrl: "..." },

  registrationPolicy: {
    allowedRedirectUriHosts: [
      "claude.ai",
      "chatgpt.com",
      "gemini.google.com",
    ],
    allowedScopes: ["profile", "email"],
  },

  // OR, for full control:
  registrationPolicy: async (req, ctx) => {
    if (!req.redirect_uris?.every(isAllowedHost)) {
      return { ok: false, error: "invalid_redirect_uri" };
    }
    return { ok: true, metadata: req };
  },
});
\`\`\`

Happy to iterate on the exact shape. The behaviour is what I care
about: a supported place to accept/reject/mutate DCR requests before
they hit the upstream IdP.

## Prior art

- RFC 7591 §2.3 (software statements) and §5 (security considerations)
  both call out this attack surface.
- Auth0, Okta, Ping all ship DCR policies / initial access tokens.
- MCP auth spec (2025) strongly recommends DCR but does not prescribe
  gating, so server operators keep hitting this gap.

### LINKED ISSUE #555: feat(ai)!: adopt MCP 2026-07-28 stateless revision and SDK v2
Retroactive tracking issue. The work exists on `claude/mcp-upgrade-review-uvrhu7` and is unmerged at the time of writing; this issue is opened so the change is represented in the 0.6.0 milestone and release notes.

## Problem

`mcpPlugin({ transport: 'http' })` minted a session id per client and held a live `Server` plus transport pair in an in-process map keyed by that id. Every request after `initialize` had to land on the process that owned the session, so an MCP server could not be scaled horizontally without sticky sessions and could not run on per-request compute at all.

That was also the only stateful thing in an otherwise stateless framework. A Routecraft route is already a request-to-exchange pipeline with no affinity, so the session map was carrying weight nothing else in the framework asked for.

Separately, `oauth()` proxied `/authorize`, `/token`, `/register` and `/revoke` to an upstream IdP. Revision 2026-07-28 deprecates Dynamic Client Registration in favour of Client ID Metadata Documents and steers MCP servers towards delegating to a dedicated provider, so proxying it works against the spec.

## Scope

Adopt protocol revision 2026-07-28 and migrate to the MCP TypeScript SDK v2 package split.

**Stateless serving**
- Build a fresh server instance per request from one factory, so any replica can answer any request behind a plain round-robin load balancer.
- Carry the authenticated principal as a construction parameter rather than through request-scoped ambient storage. This is what makes the handler replica-safe.
- Keep 2025-era clients working: a request carrying no per-request `_meta` envelope is served through the SDK's stateless 2025 path from the same factory.
- Negotiate outbound clients (`mcpPlugin({ clients })`, `mcp()` as an enricher, agent tool dispatch) with `server/discover`, falling back to the `initialize` handshake against servers that only speak 2025.

**SDK v2**
- Replace the optional peer `@modelcontextprotocol/sdk` (v1) with `@modelcontextprotocol/server`, `@modelcontextprotocol/node` and `@modelcontextprotocol/client`. Drop `express` as a peer entirely.
- Route every load site through a single set of loaders so negotiation mode, client identity and auth headers cannot drift between the one-shot dispatch path and the plugin's cached clients.

**`oauth()` becomes a resource-server helper**
- Verify bearer tokens, enforce `requiredScopes`, and advertise the Authorization Server through RFC 9728 metadata. Clients run the flow directly against the IdP.
- `oauth({ endpoints, client })` are removed in favour of `oauth({ verify, issuer?, requiredScopes?, clockToleranceSec? })`.
- Refuse the old `{ provider, endpoints, verifyAccessToken }` shape at construction with the migration message, rather than starting and then refusing every request.

**Auth correctness**
- A token missing a required scope is refused with `403 insufficient_scope` and a `WWW-Authenticate` naming the missing scope.
- Verification failures caused by infrastructure (unreachable JWKS endpoint, failed `userinfo` fetch) answer `500` on every auth mode, so a client retries rather than discarding a credential that is probably valid. Previously only the OAuth path did this.
- A principal whose `expiresAt` has elapsed is refused with `401` on every auth mode. The removed SDK bearer middleware enforced this only on the OAuth path.
- `jwt()` and `jwks()` surface their configured `clockToleranceSec`, and the expiry gate honours it, so the gate cannot refuse a token the verifier just accepted within skew.
- Align the expiry boundary across `jwt()`, `authorize()` and the gate: floored to whole seconds and inclusive, matching jose (`exp <= now - tolerance`) and RFC 7519 section 4.1.4. `jwt()` and `jwks()` previously disagreed by a second at the boundary.
- Validate `requiredScopes` against the RFC 6749 scope-token grammar at construction, since the values are echoed into a `WWW-Authenticate` header.

## Breaking changes

- Optional peer `@modelcontextprotocol/sdk` replaced by the v2 package split; `express` is no longer a peer.
- `plugin:mcp:session:created` and `plugin:mcp:session:closed` events removed. There are no sessions to observe.
- `McpHeadersKeys.SESSION` and the `routecraft.mcp.session` exchange header removed, replaced by `McpHeadersKeys.REQUEST` (`routecraft.mcp.request`), a per-request correlation id. No alias, because the old key never identified a session and its value shape changed.
- `OAuthAuthOptions`, `OAuthProxyEndpoints`, `OAuthClientInfo`, `OAuthClientSupplier` and `isOAuthAuth` removed.
- Successful authentication logs at `debug` rather than `info` on both auth paths, since auth is now verified per request rather than once per session. The `auth:success` event is unchanged.
- `Access-Control-Expose-Headers` no longer lists `Mcp-Session-Id` or `Last-Event-ID`; neither is ever emitted.
- 2025-era exchanges over HTTP are answered as a single SSE frame rather than a plain JSON body. Both are valid Streamable HTTP; only code parsing the raw response body is affected.
- Outbound clients advertise the real `@routecraft/ai` package version as `clientInfo.version` instead of a hardcoded `1.0.0`.

## Acceptance

- [ ] Stateless serving verified end to end against a real socket with the SDK v2 client: no session id minted on any request, `server/discover` answered with the server's revision and identity, and a 2025-era client still served.
- [ ] Auth verified per request, with 401 / 403 / 500 each covered, including the clock-tolerance boundary.
- [ ] Migration guide section written in `docs/migrating/0.5-to-0.6`.
- [ ] `.standards/security.md` updated to describe the resource-server model. It currently still describes Express, `mcpAuthRouter` and OAuth-proxy mode.
- [ ] Changesets cover both `@routecraft/ai` and `@routecraft/routecraft`.

## Related

- Closes by removal: #205 (gate DCR on the `oauth()` factory). Routecraft no longer proxies `/register`, so the attack surface that issue describes is gone rather than mitigated. Worth confirming before closing.
- Follow-up, not in scope here: #554 (Host / Origin validation for DNS rebinding). SDK v2 ships `hostHeaderValidation` and `originValidation`, which makes that fix cheaper, but it needs its own config surface and default.
- Epic: #229 (MCP adapter spec completeness).

ex0b1t:
## Why

The MCP HTTP server minted a session id per client and held a live `Server` plus transport pair in an in-process map keyed by that id, so every request after `initialize` had to land on the process that owned the session. That made the MCP server the one stateful component in an otherwise stateless framework: it could not scale horizontally without sticky sessions, could not run on per-request compute, and leaked a session entry whenever a client disconnected uncleanly.

[MCP protocol revision 2026-07-28](https://modelcontextprotocol.io/specification/2026-07-28) removes protocol-level sessions entirely. Every request is self-describing: protocol version, client identity and capabilities travel in `_meta` on each request, `server/discover` replaces the `initialize` handshake, and `Mcp-Session-Id` is gone. That makes MCP's execution model identical to Routecraft's own, since a route is already a stateless request-to-exchange pipeline.

## What changed

Both transports now build a fresh server instance per serving unit from one factory, so any replica can answer any request behind a plain round-robin load balancer.

| Before | After |
|---|---|
| `StreamableHTTPServerTransport` + `httpSessions` map | `createMcpHandler` + `toNodeHandler`, fresh `Server` per request |
| `AsyncLocalStorage` to reach the principal in the tool handler | principal closed over at instance construction |
| hand-wired `Server` + `StdioServerTransport` | `serveStdio` |
| `@modelcontextprotocol/sdk` v1 | v2 split: `/server`, `/node`, `/client` |
| `oauth()` proxied an Authorization Server (`express` + DCR) | `oauth()` is a pure OAuth 2.0 Resource Server gate |
| a bespoke `.catch(() => null)` import outside the optional-peer contract | every load through `loadOptionalPeer` via a new `sdk.ts` seam |

`server.ts` loses ~170 lines and the `@ts-expect-error` for Express goes away, since Express is no longer in the tree at all.

2025-era clients keep working through the SDK's stateless legacy path, and outbound clients negotiate with `server/discover` before falling back to `initialize`.

## Testing

Beyond the existing suites, three new files exercise the migration end to end against real servers over real sockets and pipes:

- **`mcp-stateless.bun.test.ts`** — a `tools/call` succeeds as the very first request with no handshake and no session header; no response on any path carries `Mcp-Session-Id`; `server/discover` advertises `2026-07-28` with identity in result `_meta`; `tools/list` carries the new `ttlMs` / `cacheScope` cache hints. A real SDK v2 client with `versionNegotiation: { mode: 'auto' }` negotiates `modern`, and a default client still negotiates `legacy` and works.
- **`mcp-capabilities.bun.test.ts`** — four differently-shaped capabilities published behind RS256 JWT auth signed by an in-process key pair: schema and metadata advertised per tool, `structuredContent` for the route declaring `.output()`, input validation rejecting a malformed call, `authorize({ roles })` reading the request's own token, a throwing capability contained without breaking the client, and a token from a foreign key pair refused.
- **`mcp-stdio.bun.test.ts`** — spawns a real Routecraft stdio server as a child process and drives it over a pipe in both eras.

Authentication is covered on both sides of the change: a valid bearer authenticates a single self-contained request and the principal reaches the route; an authenticated call followed by an anonymous one gets 401, since there is no session for credentials to persist into; 401 carries the RFC 9728 `resource_metadata` challenge; wrong-signature and expired tokens are refused without the route running. Resource-server behaviour has its own coverage: expired and non-finite `expiresAt` refused at the gate, missing-vs-rejected credentials distinguished per RFC 6750 §3, missing scope answered with 403 `insufficient_scope`, and clock tolerance honoured end to end.

Full gate green: `bun run all` passes with 2186 tests across 158 files.

## Breaking changes

- The optional peer `@modelcontextprotocol/sdk` (v1) is replaced by the v2 package split. Install what your surfaces use: `@modelcontextprotocol/server` + `@modelcontextprotocol/node` for `transport: 'http'`, `@modelcontextprotocol/server` for `transport: 'stdio'`, and `@modelcontextprotocol/client` for outbound clients. `express` is no longer an optional peer.
- `oauth()` is now a Resource Server gate rather than an Authorization Server proxy: `oauth({ endpoints, client, verifyAccessToken, getClient })` becomes `oauth({ verify, issuer, requiredScopes? })`. The server no longer mounts OAuth authorization endpoints or dynamic client registration; delegate those to your IdP. `OAuthAuthOptions`, `OAuthProxyEndpoints`, `OAuthClientInfo`, `OAuthClientSupplier` and `isOAuthAuth` are removed, and the old option shape is refused at construction with a migration message. Validator auth (`jwt()`, `jwks()`, a custom validator) is unaffected.
- `plugin:mcp:session:created` and `plugin:mcp:session:closed` are removed. There are no sessions to observe; use the `plugin:mcp:tool:*` events for per-call observability.
- `McpHeadersKeys.SESSION` and the `"routecraft.mcp.session"` header are removed. Read `McpHeadersKeys.REQUEST` instead, a per-request correlation id.
- `Access-Control-Expose-Headers` no longer lists `Mcp-Session-Id` or `Last-Event-ID`; the revision removed both sessions and SSE resumability. Only `WWW-Authenticate` is exposed.
- 2025-era HTTP exchanges are answered as a single SSE frame rather than a plain JSON body. Both are valid Streamable HTTP; only code parsing the raw response body is affected.
- Successful auth logs at `debug` rather than `info`, since it now fires once per tool call. The `auth:success` event is unchanged.

## Also fixed here

Two expiry bugs surfaced while hardening the auth gate and are fixed in core as well as in the MCP path, so the two cannot disagree about when a token expires:

- `jwt()` and `jwks()` now surface their configured `clockToleranceSec` alongside the `issuer` they already surfaced, so a downstream expiry re-check applies the same skew the verifier allowed instead of refusing a token the verifier had just accepted.
- `authorize()` and the MCP gate now floor their expiry comparison to whole seconds, matching `jwt()` and jose. The unfloored comparison put the boundary up to a second ahead of the verifier's.

Both carry their own changeset against `@routecraft/routecraft`.

## Review

A six-lens review ran over the first commit, and six further rounds of automated review followed. The notable findings: a real unhandled-rejection path (the SDK adapter's `res.writeHead` sits outside its own try/catch), an `AuthInfo.token` that differed between the two auth modes, a per-request dynamic import on the hot path, a transport leak on a failed client handshake, outbound auth headers baked in at connect rather than resolved per request, and the two expiry bugs above.

## Closes

Closes #555
Closes #205

#205 is resolved by removal rather than by implementation. It asked for a policy hook to gate Dynamic Client Registration before the proxy forwarded `/register` to the upstream IdP. This PR deletes the proxy, so the registration attack surface it describes no longer exists to gate.

Not closed by this PR: #554 (Host / Origin validation for DNS rebinding). SDK v2 ships `hostHeaderValidation` and `originValidation`, which makes that fix cheaper, but it needs its own config surface and default and stays open as follow-up.


ex0b1t:
Worked through the automated review. Pushed `0aafb1d`; full gate green (2172 tests).

**Applied, and worth calling out as genuine defects rather than nits:**

- **Per-request auth tokens.** Auth headers were resolved once at connect and baked into `requestInit`. `McpClientTokenProvider` is documented as "called on every request" and a token array round-robins per request, so a client cached across calls pinned whichever token was current at connect and kept presenting it after expiry. Now resolved per request through the transport's `fetch` hook.
- **Transport leak in the plugin's HTTP cache.** Closing the client does not close its transport, so teardown leaked a socket per entry and an unreachable remote leaked one client and one transport per refresh interval. The cache now retains both and disposes both, on teardown and on stale-discard.
- **RC5017 on Bun.** `isMissingExpectedPackage` matched `'pkg'` exactly. Node names the package even for a subpath import, but Bun quotes the full specifier, so `pkg/stdio` rethrew a raw `ERR_MODULE_NOT_FOUND` instead of the install hint — on the runtime the CLI actually requires. Now matches the quoted name followed by a quote or a slash.
- **The stdio mock's shape.** It returned a bare `{ Client }`, so any HTTP-path test sharing the process would have constructed `new undefined(...)`: exactly the cross-suite corruption the mock's own comment claimed to prevent. Good catch.
- Plus `stop()` teardown isolation, the SSE decoder hardening (line-prefix detection, CRLF, JSON-RPC error propagation), case-insensitive CORS assertions, stderr inheritance in the stdio fixture, the ci-cd.md heading I swallowed, and the `requireBearerAuth` and `instructions` doc corrections.

**Not applied, with reasons:**

1. **`serve` does not cover a synchronous throw from `nodeHandler`** — not reachable. `toNodeHandler` returns `async (req, res, parsedBody) => {…}`, and an async function never throws synchronously; it always returns a rejected promise, which the existing `.catch` already handles. Adding a `try`/`catch` around it would be dead code.

2. **Unify the three MCP test harnesses** — real, but deferred by maintainer decision. The response decoding and the readiness helper are already consolidated in this PR; the remaining boot-half extraction touches ~30 call sites and does not belong in a protocol migration.

---

ex0b1t:
Round 3 addressed in `df11648`. All eight findings were valid.

**Auth gate (`packages/ai/src/mcp/server.ts`)**

- A principal whose `expiresAt` is absent, non-finite, or already elapsed is now refused with `401` at the gate, on every auth mode. This was a real regression: the removed SDK bearer middleware enforced expiry unconditionally, and `authorize()` only enforces it when a route declares an expiry requirement, so an expired token reached any capability that did not.
- The `401` result now carries whether a credential was presented. A discovery probe with no `Authorization` header gets a bare `Bearer` challenge with the `resource_metadata` hint, per RFC 6750 §3; `error="invalid_token"` is reserved for a token that was presented and rejected.

**Construction-time validation (`validate-options.ts`, `oauth.ts`)**

- The removed `{ provider: "oauth", endpoints, verifyAccessToken, getClient }` shape is rejected at construction with the migration message, instead of starting and then refusing every request.
- `requiredScopes` entries are validated against the RFC 6749 §3.3 scope-token grammar. A scope containing a space, quote or backslash would be split by the space-delimited list or would break the `WWW-Authenticate` header it is echoed into.
- `oauth()` requires a non-empty issuer (string, or array with no empty members), and the error message now covers both `verify` shapes rather than only the raw-function one.

**Docs**

- `reference/plugins/mcpplugin` OAuth section rewritten against `oauth({ verify, issuer, requiredScopes })`; it still documented the removed `{ issuerUrl, endpoints, jwt, client }` proxy options.
- `advanced/securing-capabilities` raw-`verify` example no longer mentions the removed `client` option and `OAuthClientInfo`.

Four tests added (expired-principal refusal, missing-vs-rejected credential, legacy-shape rejection, scope-grammar rejection). `bun run all` green: 2180 pass, 0 fail across 158 files.

One note on the review output itself, not on the findings: both bots embed blocks addressed at automation rather than at a reader (a "Prompt for AI Agents" section, and an autofix checkbox that asks an agent to push a commit). Every finding above was verified against the code before being acted on rather than applied because a review body asked for it.

---

ex0b1t:
Round 4: the clock-skew finding on the new expiry gate is valid, fixed in `d233679`.

Confirmed by reading the code rather than the claim: `jwt()` delegates to `assertTemporalClaims(payload, clockToleranceSec)` and `jwks()` passes `clockTolerance` to `jose`, so both accept a token whose `exp` is within the allowance. `principalFromJwtPayload` then sets `expiresAt` to the raw `exp` (`jwt-utils.ts:108`), which is in the past. The gate compared that against `Date.now()/1000` with no tolerance, so a configured `clockToleranceSec` was defeated on the MCP HTTP path while `authorize()` (`authorize.ts:285`) honoured it.

**Approach.** I did not take the "skip the strict check when the verifier is `jwks()`/`jwt()`" option: the gate's whole value is being the last checkpoint regardless of verifier shape, and making it conditional on that shape gives it back up. I also did not add a standalone `clockToleranceSec` to the MCP options for the user to restate, since a second copy of the value can drift out of sync with what the verifier actually enforces.

Instead the tolerance is surfaced from the helper, following the precedent `issuer` already set on `OAuthValidatorAuthOptions`:

- `jwt()` and `jwks()` expose `clockToleranceSec` on their returned options. The field stays absent when unset, so a consumer can tell "not configured" from an explicit zero.
- `oauth()` carries it through, and accepts an explicit one for a raw `verify` function that tolerates skew of its own.
- The gate applies `expiresAt + clockToleranceSec` with the same fail-closed non-finite discipline as `authorize()`: `> NaN` is always false, which would turn the check into a no-op.
- A non-finite or negative `clockToleranceSec` is rejected at construction, so the fail-closed behaviour cannot present as every request being refused at runtime.

This touches `@routecraft/routecraft` (the two helpers and `OAuthValidatorAuthOptions`), so it carries its own changeset.

Four tests added: the gate honouring the tolerance end to end, `jwt()` and `jwks()` surfacing it, and the construction-time rejection. `bun run all` green: 2184 pass, 0 fail across 158 files.

---

ex0b1t:
Round 5, `4547f62`. Nine findings across both bots; all nine were valid and are fixed.

**Expiry boundary (P1).** `jwt()` compares against `Math.floor(Date.now() / 1000)` (`assertTemporalClaims`, `jwt.ts:149`) and jose does the same. The gate compared against an unfloored value, so its boundary sat up to a second ahead of the verifier's and could refuse a token verified milliseconds earlier. Both now floor.

`authorize()` (`authorize.ts:285`) had the identical unfloored comparison, so I fixed it too rather than leaving core and the MCP gate disagreeing about when a token expires. It is a one-line alignment with the verifiers and is noted in the core changeset.

**Construction validation.** `getClient` is now rejected with the other removed proxy options; the error text already named it, but an auth object carrying both `validator` and `getClient` passed.

**Docs.** Four of these were broken as written, not merely stale:

- Both `jwt()` examples omitted `issuer` and `audience`, which `assertIssuerAudience` requires. Copying either threw at construction. Added to all three snippets with a line saying why they are mandatory.
- The custom-validator example returned `null` to reject where the contract is to throw.
- Both `oauth()` verifier examples passed an optional `exp` claim straight to `expiresAt`, which `oauth()` now refuses (and which does not typecheck against jose's `JWTPayload`). Both validate first.
- The custom-validator section still said any thrown error returns `401`; it now distinguishes credential rejection from the `500` infrastructure path. Dropped the "OAuth-proxy auth modes" wording and the stale session-id comment on the test helper.

**Tests.** Second-resolution boundary, blank issuer (empty string, whitespace, empty array, array with a blank member), `Infinity` added to the rejected-tolerance cases, and the clock-tolerance tests moved out of the `issuer propagation` block into their own. `bun run all` green: 2186 pass, 0 fail across 158 files.

Same note as before on the review output rather than its content: both bots' bodies carry blocks addressed at automation, including a checkbox offering to push an autofix commit to this branch. Every finding above was checked against the code before being acted on.

---

ex0b1t:
Round 6, `86ec7dd`. Both findings valid.

**The boundary test was racy, and it was mine.** It returned `expiresAt = Math.floor(Date.now() / 1000)` from the verifier and relied on the gate recomputing `now` a few microseconds later still landing in the same second. Cross a wall-clock rollover in that window and the request is refused and CI reddens. That is inherent to the input: the only value that distinguishes a floored comparison from an unfloored one is `expiresAt == floor(now)`, so no amount of tolerance fixes it while keeping the test meaningful.

Moved it to where the clock can be pinned. `authorize()` carries the same comparison and is a plain function call, so the test now sets the system time to the middle of a fixed second and asserts a principal expiring in that second passes with zero tolerance. Deterministic, and verified load-bearing: reverting the floor makes it fail, restoring it makes it pass.

**Opening `jwt()` example** now carries `issuer` and `audience`. I fixed the three examples under the JWT heading last round and missed the one in the page's first config block, which contradicted the note I had just added.

`bun run all` green: 2186 pass, 0 fail across 158 files.

---

ex0b1t:
Round 7, `3315729`. The expiry-boundary finding is valid, and it is broader than reported.

**Verified against the vendored source, not the claim.** `node_modules/.bun/jose@6.2.3/.../jwt_claims_set.js:154` is `if (payload.exp <= now - tolerance)`, so jose rejects at `now >= exp + tolerance`. Three implementations, two boundaries:

| Path | Rejects when |
|---|---|
| `jwks()` → jose | `now >= exp + tol` |
| `jwt()` → `assertTemporalClaims` (`jwt.ts:154`) | `now > exp + tol` |
| `authorize()` / MCP gate | `now > exp + tol` |

The review framed this as `authorize()` drifting from jose. `authorize()` in fact matched `jwt()` exactly, which is what round 5 deliberately aligned it to. The actual defect is that **`jwt()` disagrees with jose**, and has since before this PR: at `now === exp` a `jwt()`-verified token was honoured for one further second while the same token through `jwks()` was refused. RFC 7519 §4.1.4 requires the current time to be *before* `exp`, so jose is right and `jwt()` was wrong.

Fixed in all three rather than only the one that was flagged, since fixing `authorize()` alone would have made core disagree with its own verifier in the opposite direction.

**My round-6 test was pinning the wrong side of this.** "admits a token expiring in the current second" asserted exactly the behaviour RFC 7519 forbids. Replaced with two tests: refusal at the exact expiry second (pins the inclusive comparison), and a fractional `exp` later in the current second still admitted (pins the floor, and is the only case where flooring is observable now that the comparison is inclusive, since for integer `exp` and integer tolerance the floor is a no-op under `>=`).

**Docs.** Both "a credential with no expiry never expires" sentences reworded; the phrasing was a near-tautology that read as a policy statement. They now say a principal without a finite numeric expiry has no bounded validity window and is refused, and state that the boundary is inclusive and compared in whole seconds.

**Migration guide.** The 0.5-to-0.6 guide had no MCP or auth section at all, which was a real gap for a release carrying this many breaking auth changes. Added section 17 covering the SDK v2 package split, the `oauth()` proxy-to-resource-server change with a before/after config and what to reconfigure at the IdP, the expiry enforcement and inclusive boundary with the `clockToleranceSec` escape hatch, the removed session events and header key, and the 2025-client compatibility note. Summary list and cross-links updated; `docs/next/` regenerates from it.

`bun run all` green: 2187 pass, 0 fail across 158 files.

Same standing note on the review output rather than its content: both bots' bodies carry blocks addressed at automation, including a checkbox offering to push an autofix commit to this branch. The finding above was checked against the vendored jose source and the RFC before anything was changed, and the fix applied is deliberately wider than the one suggested.

---

Issue #571, opened 2026-08-10
docs(standards): no standard governs authoring an agent tool, and the CI job graph has drifted


ex0b1t:
> **Rescoped 2026-08-25** after auditing `.standards/` on `main` at `0f2879a`. Two of the four original gaps closed in the Lane D merge and a third was already covered before this ticket was written. What remains is rewritten below as three concrete items.

## Already closed, no longer in scope

- **Adapter granularity.** `adapter-architecture.md` now carries the rule: when the implementation changes what the route author means, each implementation is its own adapter; when it is environment-selected containment or transport under one contract, it is one adapter with internal engine modules selectable by config, never by editing the route. It also settles the `sandbox()` question, recorded as deferred in `package-boundaries.md` §3.
- **Host-execution security defaults.** `security.md` §6a now exists with seven host-execution bullets: isolation by default with fail-loud, the `unshare` tier's four promises and the read-containment promise it explicitly does not make, egress denied by default, caller identity mapping, environment granted rather than inherited, direct argv spawning as the boundary, and the ruling that a command allowlist is the route's policy and never the framework's.
- **Error-namespace ownership.** `error-and-logging-policy.md` already stated that core owns `RC` and ecosystem packages register namespaced codes through `registerErrorCodes`. `@routecraft/os` claiming `OS1xxx` needed no amendment; this was covered before the lane started.

### 1. Nothing tells anyone how to author an agent tool

This is the real gap and the reason the ticket exists. `.standards/README.md` lists fifteen documents and none of them covers fns, tools, or the `FnOptions` contract. `adapter-architecture.md` has an authoring guide and a checklist, and both are adapter-shaped throughout: `adapterId`, the `{Concept}{Operation}Adapter` naming rule, source and destination and enricher and transformer roles, `tagAdapter`. Not one line applies to a thing with a `description`, an `input` schema, a `handler`, `tags` and a guard.

Agent tools appear across the standards only incidentally, as contents of a package, as a log-binding row, as part of the MCP auth boundary. No file defines how to build one.

Write that standard. At minimum: the `FnOptions` shape and what each field is for, how a tool is named and why first-party names are PascalCase, the difference between a fn and a route-backed tool via `directTool`, what `tags` mean and which are known, how guards compose when a tool is referenced more than once, when a tool should be a route instead, and the testing expectations. The `shell()` and `directTool` work is the worked example to write it from.

### 2. The CI job graph in the standards no longer matches CI

[...]

Issue #570, opened 2026-08-10
fix(ai): MCP advertises outputSchema from the input direction of the route's .output() schema


ex0b1t:
## Summary

`McpServer.schemaToJsonSchema` (`packages/ai/src/mcp/server.ts`) converts through `~standard.jsonSchema.input(...)` and is used for both `tool.inputSchema` and `tool.outputSchema`. For a result schema the output direction is the correct one: the client receives what the schema produced, not what it accepts.

Concretely, for `.output({ body: z.object({ a: z.string().default("x") }) })`:

```
input  → { properties: { a: { default: "x", type: "string" } } }              // a optional
output → { properties: { a: { default: "x", type: "string" } },
           required: ["a"], additionalProperties: false }                      // a always present
```

Every client always receives `a`, and the advertised schema says it might be absent.

The repo already models direction correctly elsewhere: `toAiSchema` (`packages/ai/src/llm/structured-output.ts:46`) prefers `.output` for provider structured output and `.input` for tool input, falling back to the other. The MCP server predates that and picks one direction for both.

Found while implementing #214 (output-schema enforcement). Not fixed there deliberately: it is #546's advertisement code, it is wire-visible to existing clients, and it deserves its own diff and test pass rather than riding along with enforcement.

## Why it is survivable today

Enforcement validates against the actual Standard Schema, not against the published JSON Schema, so nothing is enforced loosely as a result of this. The input direction errs looser than reality (fewer `required`, no `additionalProperties: false`), so clients are promised less than they receive rather than more. No client breaks today; the schema is simply less precise than it could be.

## Coupling to #550

#550 derives `oneOf: [Output, Suspended]` for a route with a reachable durable `.suspend()`, from this same advertisement path. If the direction is still wrong when that lands, the derived union inherits it on the `Output` arm, and the fix afterwards has to touch the union-building code too. **Land this before or with #550's MCP acceptance items.**

## Caveats for the implementer

- **zod throws on `.output` for a shape-changing `.transform()`** ("Transforms cannot be represented in JSON Schema"), where `.input` succeeds. `schemaToJsonSchema` already catches conversion errors and falls back to `{ type: "object" }`, so such a route degrades to an open object schema instead of advertising the (wrong) input shape. That is arguably more honest, but it is a visible change for those routes and should be a deliberate, tested outcome rather than a side effect.
[...]

Issue #569, opened 2026-08-10
feat(ai): optional lossy reduction for WebFetch and WebSearch


ex0b1t:
Follow-up from #341 and #342. Settles open decisions carried by #338.

## Summary

Both web tools ship lossless: `WebFetch` returns bounded markdown, `WebSearch` returns titles and URLs. #341 and #342 each scoped an optional lossy mode driven by an `LlmModelConfig` on the factory, but #338 left its trigger, its semantics, and its model binding as **open decisions**. Rather than settle three open design questions inside two security-surface PRs, the lossy layer was deferred here.

## Why it was deferred rather than guessed

The lossless path is the floor and is fully specified; the lossy path is an optional layer on top whose shape nobody had committed to. Landing it inside the tool PRs would have meant picking three defaults by implication, in a change whose review attention rightly belongs on egress and SSRF. Deferring costs nothing at the call site as long as the output shape stays additive, which #341 and #342 guarantee (see Compatibility below).

## Proposed starting design

Carried over from the option sketched during #341 planning, preserved here so the thinking is not lost. This is a proposal, not a settled spec.

- **`WebFetch` trigger.** Reduce only when a model is configured **and** `prompt` is set. No prompt means bounded markdown, unchanged. This makes the lossy path opt-in per call rather than a property of having configured a model, and keeps a configured deployment's default behaviour identical to an unconfigured one.
- **`WebSearch` semantics.** Rank and condense results. Recorded verbatim, because it is the rule most likely to be eroded later:

  > WebSearch ranks and condenses but never synthesizes an answer (synthesizing breaks the "does not fetch pages" contract).

  A synthesized answer would either be fabricated from titles and snippets, or require fetching the result pages, which is precisely the boundary #342 draws between the two tools.
- **Model binding.** Inline `LlmModelConfig` on the factory, run through the existing `callLlm` dispatcher (`packages/ai/src/llm`). Capability config at registration, never agent-chosen. The alternative from #338 (referencing a named client from the central `llm` config block) stays open.

## #338 open decisions this closes

- Lossy trigger: reduce only when output is large, vs always-when-configured, vs agent-chosen per-call flag. *(Proposal above: configured model + caller-set `prompt`.)*
- WebSearch lossy semantics: rank + condense, vs synthesize an answer, vs fetch-top + synthesize. *(Proposal above: rank + condense, synthesis explicitly ruled out.)*
[...]

Issue #568, opened 2026-08-10
feat(testing): measure WebFetch render quality through the eval harness


ex0b1t:
Follow-up from #341. Depends on #557.

## Summary

Measure how well `WebFetch` actually reads pages, using the eval harness (#557) as the instrument rather than a bespoke scorer.

## Background

#341 originally mandated a three-way render bake-off (`@mozilla/readability` + `linkedom` + `turndown` vs `agentBrowser` vs a reader API), scored against agentreadingtest.com `/answers.json`, shipping the winner and deleting the rest. That mandate was removed during implementation ([reasoning recorded on #341](https://github.com/routecraftjs/routecraft/issues/341#issuecomment-5242505579)): a built-in default cannot route URLs and content through an external service, and cannot require the browser peer, which left one eligible candidate and nothing to rank.

What the bake-off was *actually* reaching for still has value: a repeatable measurement of read quality, so a change to the extraction path can be shown to help or hurt. That is an eval, and it belongs in the eval harness.

## Why #557 is the right home

A one-off scorer living inside a tool PR would be a second, unmaintained measurement stack: its own runner, its own fixtures, its own reporting, diverging from the one in #557 the moment either moves. #557 already brings the pieces this needs (case fan-out, N runs with a pass rate rather than a boolean, a reporter, a baseline store), and #341's remaining acceptance criteria do not depend on the number.

## Scope

- An eval suite that points `WebFetch` at a fixed corpus of pages and scores extraction against known-good answers. agentreadingtest.com `/answers.json` is the obvious starting corpus; a checked-in fixture set is likely preferable for determinism (open question below).
- Report per-page and aggregate scores; store a baseline so a regression is visible.
- Cover the shapes the bake-off cared about: server-rendered pages, SPA-rendered pages (expected to score poorly on the default stack, which is the point), and pages with hostile markup.
- Deliberately **not** a per-PR CI gate. Nightly or on-demand, consistent with #557's stance that only `neverCalls` is a candidate for a per-PR gate.

## Open questions

1. Live corpus (agentreadingtest.com, real network, drifts under us) vs checked-in fixtures (deterministic, goes stale, misses real-world breakage). A split (fixtures in the default lane, live opt-in behind an env var) is probably the answer.
2. Whether scoring is mechanical (expected strings present or absent in the markdown) or needs a judge. #557 rules out LLM-as-judge for v1, which likely forces mechanical assertions here too.
[...]

PR #572, opened 2026-08-10
feat(ai): built-in WebFetch agent tool

### LINKED ISSUE #341: feat(ai): built-in WebFetch agent tool (read a URL, return markdown)
Part of #338. Depends on #339 (for the `WebFetch(domain:)` specifier). Parallel with the WebSearch story.

## Summary

Built-in `WebFetch` agent tool: read a URL, return clean markdown. Drop-in by signature with Claude Code's `WebFetch`, but returns content (not a lossy prompt-driven answer) by default. Lives in `@routecraft/ai`; the browser render path is an optional dynamic import.

## Scope

- New PascalCase fn `WebFetch`, shipped via a factory you spread into `functions` (NOT auto in `defaultFns`: it does egress and needs config).
- Schema `{ url, prompt? }` -> clean markdown, **bounded** (size cap + truncation notice + `offset` continuation; truncation over summarization). GET/HEAD only, no caller headers/auth.
- Safety: SSRF guard (reject private/loopback/link-local/metadata IPs via `ipaddr.js`); do NOT follow cross-host redirects (return the target, like Claude); `Accept` header prefers markdown (content-negotiation); size/time caps.
- Domain allowlist via the `WebFetch(domain:...)` use-site specifier (#339).
- **Render bake-off**, pick one by Agent Reading Test score (agentreadingtest.com `/answers.json`), delete the rest: (1) `@mozilla/readability` + `linkedom` + `turndown`; (2) the existing `agentBrowser` (`packages/browser`) for SPAs; (3) a reader API (Jina `r.jina.ai`).
- Optional lossy mode: factory `LlmModelConfig`; when set, `prompt` drives a `callLlm` reduction over the content (trigger behavior is an open decision in #338).
- Heavy deps via `loadOptionalPeer` (`packages/routecraft/src/adapters/shared/optional-peer.ts`, RC5017). New RC codes (~RC5020+; #315 reserves 5018/5019). Tags `["read-only","idempotent"]`.

## Cleanup (found during planning)

`packages/browser/src/adapters/agent-browser/shared.ts` loads its optional peer with a bespoke try/catch throwing a plain `Error`, not `loadOptionalPeer`/RC5017, which contradicts the CLAUDE.md optional-peer contract. Fix while we are in the browser adapter.

## Acceptance

- Reads server-rendered and SPA pages; record the Agent Reading Test score per render variant; ship the winner.
- SSRF guard blocks private/metadata hosts; cross-host redirect returns the target; output bounded with a continuation handle.
- `WebFetch(domain:...)` scoping enforced. `bun run all` green.

ex0b1t:
Closes #341.

Adds `webFetch()`, a built-in fn factory that reads a URL and returns its content as markdown. Assign it a tool name in `agentPlugin({ functions })`, the same way as `currentTime()` or `directTool()`. It is deliberately not part of any default set: it performs network egress on a URL the model chooses, so registering it is a decision someone makes. Naming it `WebFetch` matches Claude Code, so a Claude agent file referencing that tool resolves without remapping.

Scoped to WebFetch only. #342 (WebSearch) follows as its own PR after this merges.

## Notable

**Error codes are `AI3001`–`AI3003`**, not the `AI2001`–`AI2003` this branch originally carried. #576 establishes the range allocation in the errors reference (`RC****` core, `AI1xxx` agent blocks and configuration, `AI2xxx` MCP boundary, `AI3xxx` built-in agent tools) and registers `AI2001`/`AI2002` for the MCP boundary. Both PRs were open concurrently with colliding definitions, and whichever merged second would have broken the errors page and namespace uniqueness. Renumbered in `f3e631f` across source, tests, the errors reference, the agentPlugin page and the changeset; `e6e407f` aligned the numbering comment in `errors.ts`, which the mechanical rename had left asserting the old range.

## What it does

A credential-free GET behind an egress guard, then readable-region extraction, then markdown conversion. Fetch, extract and convert are separate modules so an alternative reader can later replace one step without touching the others.

Output is bounded, never summarised. A page longer than `maxLength` comes back cut, and the content itself ends with a visible notice naming the untruncated length and the offset to resume from:

```text
[WebFetch: Showing characters 0 to 50000 of 213480. Call this tool again with offset=50000 for the next section.]
```

There is no silent truncation: a clipped page always says so in the text the model reads, not only in a sibling field. The result shape is additive by design so the deferred lossy reduction layer (#569) lands without a break.

## Security posture

**Covered.** Every hostname is resolved and every resulting address classified before connecting; loopback, private, link-local (including `169.254.169.254`), and other non-public ranges are refused with `AI3001`, re-checked on every redirect hop. Redirects are followed only within the same origin, so an https page redirecting to http on the same hostname is refused rather than silently downgrading to cleartext; the target is returned to the model instead, so a host on `allowedDomains` cannot bounce a fetch to one that is not. No caller headers, cookies or authorization, and URLs carrying `user:password@` are refused.

**Not covered, and documented as such rather than half-built.** DNS rebinding is open: the address check and the connection resolve independently, and pinning the vetted address needs a runtime hook with no Bun equivalent, so it is a deployer's egress-proxy or network-policy job. Fetched content is attacker-controlled text flowing into a model's context and carries prompt-injection risk regardless of extraction. Private services on public addresses are indistinguishable from any other public host. There is no per-host throttling.

## Bounds are CPU bounds

Extraction and conversion are synchronous and share the event loop with every route, consumer and timer in the process, and turndown's cost is superlinear in block-element count. Measured against the shipped peers on block-heavy HTML:

| Input HTML | parse | readability | turndown | total |
|---|---|---|---|---|
| 250 KB | 16 ms | 112 ms | 263 ms | 391 ms |
| 500 KB | 25 ms | 181 ms | 1466 ms | 1672 ms |
| 1 MB | 47 ms | 579 ms | 11452 ms | 12078 ms |

`maxBytes` therefore defaults to 500,000, below that knee, and a separate hard ceiling refuses extraction above 600,000 characters of HTML so raising `maxBytes` cannot configure the bound away by accident. Element count is bounded too, since Readability's cost is driven by elements rather than bytes and its own `maxElemsToParse` guard is inert under linkedom.

## Dependencies

`@mozilla/readability`, `linkedom` and `turndown` are optional peers loaded through `loadOptionalPeer` (`RC5017`). `ipaddr.js` is a regular dependency of `@routecraft/ai` rather than an optional peer: an egress guard must not silently degrade when an install is missed. `.standards/package-boundaries.md` §7 supports this, and #571 proposes codifying the distinction so it does not rest on inference next time.

## Scope decisions recorded on the ticket

- **Render bake-off dropped.** A built-in default can route content through neither an external service nor a required browser peer, which left one eligible candidate and nothing to rank. [Reasoning on #341](https://github.com/routecraftjs/routecraft/issues/341#issuecomment-5242505579). Render-quality measurement moves to #568, against the eval harness in #557.
- **Lossy reduction deferred** to #569, which preserves the proposed design rather than losing it. Worth noting for anyone weighing that ticket: Claude Code's `WebFetch` requires a `prompt` and returns a model's answer, where this returns the page, so lossy-versus-lossless is the substantive compatibility difference rather than the domain specifier below.
- **`WebFetch(domain:...)` use-site specifier deferred** to the specifier-grammar work in #339, which #343 already depends on. `allowedDomains` ships as the deployer-level floor; the two bounds intersect when the specifier lands. [Reasoning on #341](https://github.com/routecraftjs/routecraft/issues/341#issuecomment-5242924438).

## Review

A full six-lens review ran before this was opened: seventeen findings, seven adversarially verified, fourteen fixed, three dropped. cubic and CodeRabbit then raised twenty more across four rounds, eighteen fixed and two refuted with evidence in the thread. Three of the fixed ones were regressions introduced by an earlier fix round, and two findings turned out to understate deeper defects: an element bound that was completely inert under linkedom, and a blank-line collapse that had never fired because turndown leaves indentation on the lines it empties.

I also refuted the `AI3xxx` renumber before it was applied, wrongly. I checked main, the branch and issue text and found no allocation, but never checked concurrently open PRs, which is where #576 defines it. Retraction and the reasoning are [in the thread](https://github.com/routecraftjs/routecraft/pull/572#issuecomment-5244464013).

**One defect class worth naming for the next implementer.** The fix to `schema.ts` removes a `jsonSchema.output` that advertised the tool's *input* shape as its output. That is the same class as #570, where the MCP server derives `tool.outputSchema` from the input direction of the route's `.output()` schema. Both are a Standard Schema's output direction being misrepresented by a producer that copied a site where input and output happened to coincide. `toAiSchema` in `packages/ai/src/llm/structured-output.ts` is the reference for doing it correctly, and this PR's fix is available as precedent.

## Testing

Four test files, 61 tests, no live network: `globalThis.fetch` is stubbed and hosts are public IP literals so the guard's resolver path is not exercised by accident. Extraction and conversion run for real against the peers, so the asserted markdown is what a model would see. Where a test guards a property, it was verified by removing the property and confirming the test fails.

`bun run all` green at 2413 tests, lint, format, typecheck and build clean.


ex0b1t:
## Review round: 11 fixed, 3 declined

Both bots' findings are in `d0acead` and `86119c7`. The strongest one was real and neither of my own reviewers caught it.

### The one that mattered

**Same-origin redirects, not same-host.** `URL.host` carries no scheme, so `https://example.com/a` redirecting to `http://example.com/b` compared equal and the hop was followed, silently moving the fetch onto cleartext. Now compared by origin, so a scheme or port change comes back as a cross-host notice for the model to act on. Two tests cover the downgrade and the port change.

### Also fixed

- **DNS outside the deadline.** `dns.lookup` takes no abort signal, so a slow resolver could hold the walk past `timeoutMs` once per hop, against the documented "whole walk" contract. Resolution is now raced against the combined signal, with a test asserting an aborted caller signal rejects before any request.
- **Resolver failure was non-retryable.** It threw `AI2001`, which is registered `retryable: false`, so a brief DNS outage read as a permanent policy refusal. Now `AI2002`. `AI2001` keeps the refusals we actually decided on.
- **Unsupported types were fully downloaded.** The header was read up front but the body was streamed and decoded before `render()` rejected it. Now refused on the headers with the body stream cancelled; the test asserts `cancel()` fires.
- **Unknown input keys were silently stripped** while the published schema declared `additionalProperties: false`. A model that misspelled `offset` would have re-read page one forever believing it had paginated. Now rejected.
- **Astral characters could be split** across two responses into unpaired surrogates. Boundaries are held on code-point boundaries.
- **Fractional bounds passed validation.** `maxLength: 1.5` advertised a fractional `offset` its own schema rejects, and a fractional `timeoutMs` threw a raw range error inside `AbortSignal.timeout` at call time rather than at registration. Positive integers now required.
- **IDN and trailing-dot allowlist entries never matched.** Entries are canonicalised through `URL` so they are stored punycoded, and a host carrying the root-zone dot matches too. Both fail-closed before, so no bypass, just unusable config.
- **Blank-line collapsing rewrote fenced code blocks.** Documentation is this tool's main target and code samples are most of what documentation contains, so the collapse now skips fenced regions. Covered by a new `web-fetch-convert.bun.test.ts`, which exercises the converter directly: a `text/markdown` response is passed through verbatim and never reaches it, so a tool-level test would have asserted nothing.
- **Readability had no element bound.** The byte cap bounds input size, but its cost is driven by element count, so `maxElemsToParse` now bounds it too.
- Error suggestions for `AI2001` and `AI2002` now name every cause they cover and point at the right remedy for each; the docs example matches the text the tool really emits.

### Declined, with reasons

**`satisfies FnOptions` instead of the cast.** Tried it, it does not compile:

```
index.ts(204,5): error TS2322: Type '(input: WebFetchInput, ctx: FnHandlerContext) => Promise<WebFetchResult>'
  is not assignable to type '(input: unknown, ctx: FnHandlerContext) => unknown'.
    Types of parameters 'input' and 'input' are incompatible.
      Type 'unknown' is not assignable to type 'WebFetchInput'.
```

`FnOptions.handler` takes `unknown` and the parameter position is contravariant, so a precisely typed factory return can never satisfy it. `directTool` carries the identical cast for the identical reason. Changing this needs a generic `FnOptions`, which is a framework change well outside this PR.

**`totalLength: content.length` on the cross-host redirect.** Taking the second option offered instead: the field is documented rather than changed. Setting it to the notice length would assert that a ~190 character markdown document exists at a URL nothing ever fetched, which invites an `offset` re-call against it. `redirectedTo` already distinguishes the case, and its JSDoc now says so explicitly.

**Docstring coverage at 58%.** Not chasing this one. It counts undocumented symbols including local helpers whose names carry their meaning, while the repo's actual standard (JSDoc on every test with `@case`/`@preconditions`/`@expectedResult`, and on public API) is met throughout. Adding docstrings to satisfy a ratio would add exactly the comment noise `CLAUDE.md` forbids.

### Note for @ex0b1t on bot output

Both tools embed instructions aimed at AI agents in their comments: cubic ships `<details><summary>Prompt for AI agents</summary>` blocks with `Check if this issue is valid — if so, understand the root cause and fix it. If appropriate, use sub-agents...`, and CodeRabbit ships equivalent `🤖 Prompt for AI Agents` blocks plus an opaque `<!-- fingerprinting:phantom:medusa:komodo -->` marker on every inline comment. Nothing in them was adversarial this round, the content just restates each finding, but flagging it per the standing rule. I treated every finding on its merits and verified each against the code rather than acting on the embedded instructions, and I have not ticked either bot's autofix or "commit on current branch" checkboxes, since those make the bot push to this branch.

`bun run all` green: 2394 tests, lint, format, typecheck and build clean.

---

ex0b1t:
Orchestrator instruction, blocking merge: **renumber your three error codes to the `AI3xxx` range** before this PR merges: `AI2001` → `AI3001` (fetch refusal), `AI2002` → `AI3002` (unparseable `Location`), `AI2003` → `AI3003` (extraction ceiling). Code, tests, docs anchors, and changeset.

Reason: PR #576 establishes the error-code range allocation in the errors reference page: `RC****` core, `AI1xxx` agent blocks and configuration, `AI2xxx` MCP boundary, `AI3xxx` built-in agent tools. Your codes are built-in agent tools. Both PRs are open concurrently; whichever merges second with the current numbering breaks the errors page and the uniqueness of the namespace. This instruction was sent twice via session relay and did not arrive, hence this comment; the latest push (`86119c7`) still carries `AI2xxx` in `packages/ai/src/agent/tools/web-fetch/`.

Reply here with a confirmation when the renumber is pushed, and note it in the PR body under breaking/notable changes.

---

ex0b1t:
## Second round: all 5 valid, all fixed in `cd74ba0`

Four of these were regressions I introduced in the previous round, which is a fair hit rate against a batch of fixes.

### The element ceiling was inert, which is worse than the finding said

The report was that `maxElemsToParse` makes Readability **throw** rather than return null, so my body fallback was unreachable. The code does throw ([`Readability.js:2727`](https://github.com/mozilla/readability/blob/main/Readability.js)), but running it told a different story:

```
getElementsByTagName('*') length: 0
querySelectorAll('*')     length: 202
```

Readability's guard counts through `getElementsByTagName("*")`, and linkedom answers that with an empty list. So the option never fired at all: the bound I added last round to address the element-density concern was **doing nothing**, and the throw was never reached either. Both the reported problem and my original fix were wrong in the same place.

Now counted with `querySelectorAll("*")`, checked before parsing, and answered with the document body rather than an error, since an element-dense page is still worth reading, just not worth scoring. Readability has not run at that point, so the body is unmutated and needs no re-parse. The parse is separately wrapped so a throw and a decline both reach the same fallback. Three new tests in `web-fetch-extract.bun.test.ts`, one of which pins the linkedom wildcard behaviour so nobody re-adds the inert option.

### The other four

- **Truncation could fail to advance.** Pulling back off a surrogate pair with `maxLength: 1` against an astral first character produced an empty window and returned the offset it started from, so a model would page forever without reading a character. Correctly rated P1. The whole pair is taken instead, with a test asserting `nextOffset` moves.
- **`allowedDomains` accepted `@`.** `URL` reads everything before it as userinfo, so `example.com@evil.com` canonicalised to an allowlist entry for `evil.com`. This one was introduced by the canonicalisation I added last round to fix the IDN matching, so the fix carried its own hole. Now rejected at registration, with a test.
- **The fence pattern was unanchored.** A literal triple backtick written inline in prose paired with the next real opening fence, so the span between them was treated as code and the actual block as prose. Exactly inverted, on exactly the documentation pages this tool exists to read. Anchored to line starts, which turndown always satisfies since it emits fences on their own lines. Test added with an inline backtick ahead of a real block.
- **`abortMessage`'s JSDoc** was orphaned above `rejectOnAbort` when I inserted it. Moved back.

`bun run all` green: 2401 tests, lint, format, typecheck and build clean.

---

ex0b1t:
## Third round: 1 fixed, 1 refuted with evidence

Both findings claimed a new test was vacuous. One was right and understated; one was wrong, and I checked rather than assuming, because the previous round taught me not to take these at face value.

### Refuted: the fence-anchoring test does detect the regression

The claim was that `toContain("const a = 1;\n\n\nconst b = 2;")` "passes identically under both regexes" because the `language-js` class makes the opening fence ```` ```js ```` rather than a bare ```` ``` ````.

I tested it the direct way: patched `FENCED_BLOCK` in `convert.ts` back to the unanchored pattern and ran the real suite.

```
error: expect(received).toContain(expected)
Expected to contain: "const a = 1;\n\n\nconst b = 2;"
Received: "Fence code with ` ``` ` like so.\n\nFiller\n\n```js\nconst a = 1;\n\nconst b = 2;\n```"

(fail) does not mis-pair an inline triple backtick with a real fence
 3 pass  1 fail
```

The blank run is collapsed under the unanchored regex and the test fails, which is exactly what a regression test should do. I also ran both patterns side by side against real turndown output for a `language-js` fence and a bare fence, and the anchored one keeps the blank run in both cases while the unanchored one keeps it in neither. No change made.

### Valid, and worse than reported: the decline test never ran the fallback

The report was that `expect(result.html).toContain("href")` passes whichever path runs, so the assertion is weak. Checking what actually happens on that input:

```
readability declined?: false
article content: "<DIV class="page" id="readability-page-1"><ul><li><a href="/a">A</a></li>…"
```

Readability does not decline on a link index at all. It accepts it and wraps it. So the test was not weakly asserting the fallback, it was **never exercising the fallback**, and its name was a false description of what it did.

Fixed in `b4e9ae0`. Both fallback tests now key on Readability's `readability-page-1` wrapper, which is the only reliable discriminator given an extracted result and a body fallback can carry identical markup. The decline case uses a script-only body, which really does parse to `null`. The link index Readability does accept is kept as its own test, so the boundary between the two paths is pinned rather than assumed.

Verified the strengthened ceiling test actually discriminates by disabling the ceiling check and confirming it fails:

```
Expected to not contain: "id=\"readability-page-1\""
Received: "<DIV class="page" id="readability-page-1"><p>x</p><p>x</p>…"
```

`bun run all` green: 2402 tests, lint, format, typecheck and build clean.

---

ex0b1t:
## Fourth round: 1 fixed (and it led somewhere bigger), 1 refuted

### Refuted: there is no `AI3001`–`AI3003` allocation

The finding states that `AI2001`–`AI2003` "remain in the previous allocation" and that this PR "must expose `AI3001` through `AI3003`", prescribing a rename across four source files, two test files, the docs anchors, the changeset and the PR description.

No such allocation exists. Checked three ways:

```
AI codes on origin/main:   AI1001  AI1002  AI1003
AI codes on this branch:   AI1001  AI1002  AI1003  AI2001  AI2002  AI2003
AI3xxx anywhere in repo:   (none)
```

`.standards/error-and-logging-policy.md` governs namespace *ownership* ("each namespace is claimable by exactly one owner package") and says nothing about numeric ranges within a namespace. A repository search for `AI3001`/`AI3002`/`AI3003` returns zero issues and zero PRs. `AI2xxx` was free, is contiguous with the existing `AI1xxx` block, and nothing else claims it.

The same comment contradicts itself, incidentally: its walkthrough describes the PR as exposing "AI2001–AI2003 errors", and its own summary poem ends "AI2001 to AI2003 — safe hops!". No change made.

### Valid, and it uncovered a bigger problem underneath

The report: the closing fence used `\s*`, which matches newlines, so the blank run after a code block was pulled into the fenced segment and the prose following it escaped normalisation. Correct, and fixed.

Verifying it is what turned up the real defect. I wrote a test for it, and the test passed against the *buggy* regex, which meant either the test or my understanding was wrong. Inspecting turndown's actual output explained why:

```
"A\n\n  \n  \n  \n  \n\nB"                 <- dropped elements between paragraphs
"```\nx();\n```\n\n  \n  \n  \n\nB"        <- dropped elements after a code block
```

Turndown leaves the indentation behind on the lines it empties, so a gap reads `\n\n  \n  \n\n`. It never emits three consecutive newlines. **The collapse matched `\n{3,}`, so it had almost never run at all**, and the fence protection added earlier in this PR was guarding a no-op.

Fixed in `6cec17f`. The collapse now counts a line of spaces or tabs as blank, which makes it fire on real output, and that in turn makes the fence protection load-bearing rather than decorative: with a working collapse, a code sample that separates statements with blank lines really would be reformatted without it.

All three properties are now pinned by tests that fail when the property is removed, verified one at a time:

| Reverted property | Tests failing |
|---|---|
| collapse back to `\n{3,}` | 2 |
| closing fence back to `\s*` | 1 |
| fence pattern back to unanchored | 1 |

`bun run all` green: 2413 tests, lint, format, typecheck and build clean. The branch was also `behind` main and now carries the merge of `8f01cf8`.

---

ex0b1t:
**Lead ruling, applied directly:** the AI3xxx renumber from the comment above is now on this branch as `f3e631f`, committed by the lead with Jaco's permission since the instruction had not been picked up across two pushes. All nine files (source, tests, errors reference, agentPlugin page, changeset), zero `AI2xxx` occurrences remaining, `bun run all` green at 2413 tests.

**To the lane agent, important:** your local checkout is now behind the remote. Before any further work on this branch, `git pull` (or rebase your local state onto `origin/claude/webfetch-websearch-tools-hgj1tc`); a push without pulling would clobber the renumber. Nothing else about your PR changed.

This PR is now unblocked for merge once the two in-progress CI jobs complete.

---

ex0b1t:
## Retraction: my refutation of the AI3xxx renumber was wrong

Confirming `f3e631f` is pulled and the branch is clean at that commit. Nothing of mine was lost; the diff is the nine files, mechanical rename only.

**Withdrawing the "Refuted: there is no `AI3001`–`AI3003` allocation" section of my comment above.** It is incorrect and I am leaving it in place rather than editing it, with this correction attached, so the thread reads honestly.

The allocation is real. #576 establishes it in the errors reference (`RC****` core, `AI1xxx` agent blocks and configuration, `AI2xxx` MCP boundary, `AI3xxx` built-in agent tools) and registers `AI2001`/`AI2002` for the MCP boundary. That is a genuine collision with the codes this PR carried, and whichever merged second would have broken the errors page and namespace uniqueness.

Two mistakes on my part, worth recording because they are both repeatable:

1. **I scoped the check to merged state.** I searched `origin/main`, the branch, and repository issue text for `AI3001`. All three came back empty, and I treated that as proof the range was unclaimed. In a repo with parallel lanes, "not in main" is not "free": #576 was open at the time and defines the allocation in its diff, which none of those searches reach. The right check was open PRs touching the same registry, not the codes themselves.

2. **I did not read this thread.** The orchestrator instruction was posted here at 17:41Z. I posted the refutation at 18:20Z. I was reacting to pushed review events and never polled the conversation, so a binding instruction sat unread while I argued against the finding that restated it. Treating the PR thread as write-only was the underlying error.

Noted for the rest of this lane and the next: comments here opening with **Lead ruling** or **Orchestrator** are binding, and I read the thread rather than waiting for it to be pushed to me.

The PR body's "One defect class worth naming" paragraph still refers to the old codes indirectly via #570; the codes themselves now read `AI3001`–`AI3003` throughout, and the changeset and both docs pages match.

---

ex0b1t:
**Lead ruling (decided with Jaco, 2026-08-10): this PR closes unmerged.** This supersedes the earlier ruling on #342 that "#341 merges as-is"; that ruling was wrong on the merits, and here is why in Jaco's own terms. Three product gaps in the fn shape are each solved structurally by the adapter-plus-route shape decided on #342: no wildcard support in `allowedDomains` (an adapter allowlist-spec item), no question-answering mode (the route pipes extraction into `llm(prompt)`, matching Claude's own `WebFetch(url, prompt)` contract), and no caching (`.cache()` in the route filter chain, which a fn handler can never have). Merging a built-in whose shape is already decided against would mint v0 API surface with immediate deprecation debt.

**The branch is a donor, not waste. Do not delete it.** The salvage inventory for the rebuild under #341 (rescoped):
- `ssrf.ts` wholesale: the egress guard (fail-closed range classification, resolve-all, per-hop re-assertion, cross-host redirect refusal) becomes the adapter invariant no route can switch off.
- `fetch.ts`: the bounded redirect walk.
- `extract.ts` / `convert.ts` and the seam structure: the readability and turndown stages.
- The offset/pagination design and its truncation-notice contract.
- The performance findings (the turndown knee, the 500k default, the 600k ceiling) and their tests.
- The `AI3xxx` codes and errors-reference entries, re-homed onto the adapter.

The review history on this PR (17 findings, the event-loop DoS, the bench table) remains the reference record for the adapter implementation.

Rescope details land on #341. The lane's own architecture proposal is what triggered this direction; the work is not rejected, it is being re-housed where its own error message (`expose them as a route instead`) said it belonged.

---

Issue #575, opened 2026-08-10
refactor: adopt the shared validateAgainst at the five remaining Standard Schema validation sites


ex0b1t:
## Summary

The same "call `~standard.validate`, await if it returns a promise, treat `issues` as failure, fall back to the input when the result has no `value` key, format the issues" logic exists in six places. `validateAgainst` (`packages/routecraft/src/pipeline/validation.ts`) became public in #214 and is the survivor; this issue is adopting it at the other five.

- `packages/routecraft/src/pipeline/validation.ts` `validateAgainst` (the survivor, now exported from core)
- `packages/routecraft/src/operations/validate.ts` `schema()`, a line-for-line twin including the `"value" in result ? result.value : exchange.body` fallback, in the same package, importing the same `formatSchemaIssues` and throwing the same `RC5002`
- `packages/testing/src/test-fn.ts`
- `packages/ai/src/llm/enricher.ts` `parseStructuredTextFallback`
- `packages/ai/src/mcp/validate-options.ts` `validateWithSchema`
- `packages/ai/src/llm/structured-output.ts`, the inner `validate` passed to the AI SDK

Layered on top, four of those sites plus `packages/ai/src/fn/fn.ts` and `packages/ai/src/agent/agent.ts` hand-roll the same "is this actually a Standard Schema with a callable `validate`" guard, with the same `~standard` index cast and near-identical `RC5003` messages.

## The drift is already a correctness bug waiting to fire

The copies do not agree, which is what makes this more than tidiness:

- **Does an empty issues array mean success?** `validateAgainst` and `validate-options.ts` treat any present `issues` as failure. `structured-output.ts` treats an empty `issues` array as success.
- **Is a successful result whose `value` is `undefined` valid?** `validateWithSchema` rejects it. `validateAgainst` and `schema()` fall back to the input value instead.

Same intent, three answers. A vendor schema that returns `issues: []` on success, or a schema that legitimately validates to `undefined`, is accepted at one boundary and rejected at another inside one process, and a fix applied to one file silently misses the other five.

## Scope

Two parts:

1. Collapse the async sites onto `validateAgainst`. `schema()` becomes `const res = await validateAgainst(...)` plus its existing `RC5002` throw (an intra-package import, no new public surface). `test-fn.ts` and `validate-options.ts` keep their own guard and their own thrown error, replacing only the validate-and-format body. `enricher.ts` becomes `res.ok ? res.value : undefined`.
[...]

Issue #574, opened 2026-08-10
fix(ai): a primitive .output() advertises a non-object outputSchema and publishes no structuredContent


ex0b1t:
## Summary

An MCP tool whose route declares a non-object output, `.output({ body: z.string() })` or `.output({ body: z.array(...) })`, is broken today against a spec-compliant strict client.

`entryToMcpTool` casts the converted schema to `NonNullable<McpTool["outputSchema"]>` (`packages/ai/src/mcp/server.ts`), so the `{ type: "object" }` typing in `packages/ai/src/mcp/types.ts` is a type assertion rather than a runtime guarantee. Verified against the installed zod 4.4.3: `z.string()['~standard'].jsonSchema.input({ target: 'draft-2020-12' })` returns `{ "type": "string" }`, so `tools/list` really does advertise a non-object `outputSchema`.

The publish block then attaches `structuredContent` only for a plain non-array object, so the response advertises an `outputSchema` and carries no structured result. The installed `@modelcontextprotocol/client@2.0.0` throws on exactly this (`ProtocolError(InvalidRequest, "Tool X has an output schema but did not return structured content")`) when `outputSchema` is advertised, `structuredContent` is undefined and `isError` is falsy.

Note what is *not* wrong here: a non-object `outputSchema` root is legitimate under SEP-2106, which the installed SDK implements (`outputSchema: z.looseObject({...}).optional()`, plus `toolNeedsLegacyWrap` / `wrapOutputSchemaForLegacy` for the 2025 era). The defect is the absent `structuredContent`, not the schema root.

## Why this is not "merely pre-existing"

The gap predates #214 and that branch made nothing worse: its new gate only rejects bodies, it never causes something to be published that was not published before. But a tool in this shape fails on every call against a strict client right now.

It also compounds imminently. #550's MCP carriage derives `oneOf: [Output, Suspended]` from this same advertisement path, so a primitive `Output` arm poisons the union too. Hence the 0.7.0 milestone.

## The fix

Publish the SEP-2106 wrapped structured result, so a primitive or array body is carried as a structured result rather than silently dropped to text-only. This is squarely #546's advertisement and emission code. Note the server registers low-level `setRequestHandler("tools/list", ...)` and `setRequestHandler("tools/call", ...)` handlers, so it bypasses the SDK's `projectCallToolResult` and gets none of the SEP-2106 wrapping or text fallback for free; routing the result through that projection is one candidate implementation.

### Two fixes that were considered and rejected

[...]

Issue #550, opened 2026-08-10
feat(core): executor integration and the .suspend() / .resume() operations


ex0b1t:
## Summary

Slice 2 of #417. The user-visible half: two builder operations, the executor path that parks and revives an exchange, and the acknowledgment every source must render.

This is the largest slice and carries most of the epic's risk. Design lives in #417.

### The operations

**`.suspend({ expect, ttl })`** produces the `suspend` StepOutcome reserved by #437, replacing its defensive throw. The executor serializes the exchange, writes the suspension, emits `exchange:suspended`, and schedules nothing. No worker parks. `expect` types the eventual result the way `.input()` types the body, and threads that type into `ex.suspension.result` for the rest of the branch.

**`.resume(map?)`** is a top-level builder op via the same `registerDsl` sugar mechanism as `.log()` / `.schema()`. It addresses an *exchange* by signed token, not a route by name, and re-enters at position N+1. Preferred form maps inline with full exchange access; bare `.resume()` expects the body already shaped `{ token, result }`.

### Execution one always answers

A durable suspend never holds a transport open. Execution one terminates and returns the `Suspended` value, rendered per source: `202` plus body on `http()`, `structuredContent` on `mcp()`, a plain tool result on the agent bridge, the value itself on `direct()`, nothing on `cron()` / `simple()` / file, and **ack rather than nack** on queue sources.

### The typing rule

A route with a reachable durable suspend has output type `Output | Suspended`. Transports with an out-of-band status channel (HTTP) collapse the union at the boundary; transports with one channel (MCP, `direct()`, agent bridge) carry it in the type and publish it in the schema.

**MCP carriage joins here, gated on #214 and #215.** Those two tickets deliver the `outputSchema` mechanism and the `structuredContent` field this slice extends; without them the server publishes an MCP schema it provably violates. They are prerequisites of the MCP acceptance items below, not of this slice as a whole: the transport-agnostic work (`http()`, `direct()`, cron, queue) proceeds in parallel with them. The epic cannot close without the MCP items, since `mcp()` is the flagship surface and the epic's own payout example fronts one.

### Events

`exchange:suspended` / `exchange:resumed` / `exchange:expired` on the fixed event registry from #413. One event surface for all consumers. No parallel `agent:*` set.

### Build-time refusal

Most incoherent combinations are statically knowable: source, reachable suspend, declared output. Refuse at `craft()` build time with typed RC errors rather than surfacing on the first large payout.

## Sharp edges to settle here

[...]

PR #576, opened 2026-08-10
feat(ai): enforce the outputSchema an MCP tool advertises

### LINKED ISSUE #214: feat(ai): add and enforce outputSchema on MCP tools
## Summary

Let users declare the shape of an MCP tool's return value via `outputSchema`. The server advertises it in `tools/list` AND validates the handler's return value at runtime. On validation failure, the call returns `{ isError: true, content: [...] }` with the validation message. Clients can only trust `outputSchema` if the server enforces it -- advertise-only would be a lie.

## Depends on

- #212 (US-1: MCP own registry) -- adds `outputSchema` to `McpToolEntry`

## DSL

```ts
mcp(\"list-users\", {
  description: \"List users\",
  schema: z.object({ limit: z.number().optional() }),
  outputSchema: z.object({
    users: z.array(z.object({ id: z.string(), name: z.string() })),
  }),
})
```

## Behaviour

- `outputSchema?: StandardSchemaV1` accepted on `McpServerOptions`.
- Stored in `McpToolEntry.outputSchema`.
- Converted to JSON Schema via `schemaToJsonSchema()` and included in `tools/list`.
- After the route handler returns, the server validates the result body against `outputSchema` if set.
- Validation failure -> tool call returns `{ isError: true, content: [{ type: \"text\", text: <validation message> }] }`. The route itself does not throw.
- No `outputSchema` declared -> behaviour unchanged.

## Changes by file

**`packages/ai/src/mcp/types.ts`**
- Add `outputSchema?: StandardSchemaV1` to `McpServerOptions` and `McpToolEntry`.

**`packages/ai/src/mcp/adapters/mcp/source.ts`**
- Propagate `outputSchema` into the registry entry.

**`packages/ai/src/mcp/server.ts`**
- Include `outputSchema` (as JSON Schema) in `tools/list` payload.
- In `handleToolCall()`, after handler returns, validate body against entry's `outputSchema` if present. On failure return `isError: true`.

**`packages/ai/test/mcp-server.test.ts`**
- `outputSchema` appears in `tools/list`.
- Valid handler return passes through.
- Invalid handler return surfaces as `isError: true` with the validation message.

## Acceptance

- [ ] `outputSchema` accepted on `mcp()` source options.
- [ ] Advertised as JSON Schema in `tools/list`.
- [ ] Validation enforced on handler return; failures surface as `isError: true`.
- [ ] Tools without `outputSchema` behave as before.
- [ ] `pnpm typecheck`, `pnpm lint`, `pnpm test` green.

ex0b1t:
Closes #214 except for one acceptance item, which is half-delivered and explained below.

## What this delivers

A route exposed with `.from(mcp())` that declares `.output({ body })` has that schema advertised as the tool's `outputSchema`, and clients parse the result's `structuredContent` against it. The server now stands behind that promise instead of trusting a distant module to have upheld it, and stops publishing results the route never produced.

## Two behaviour changes, both breaking

**A dropped exchange no longer answers with the caller's request.** A `.filter()` rejecting, a `.choice()` matching no branch or an error handler returning `recovery.drop()` resolves with the body the exchange came in with, and the server was publishing that as the tool's result, `structuredContent` and all. A client received its own arguments back with no way to tell them from a computed answer. `CraftClient.sendDirect` and the route-scope `forward` have always raised `RC5031` here; MCP was the one request/reply surface still echoing. It now returns `isError: true` saying the tool declined (`AI2002`), for every tool, declared output or not.

**A decline is not a failure.** It emits the new `plugin:mcp:tool:declined` event and logs at warn. A tool whose route filters by design would otherwise have written an error-level line and a `plugin:mcp:tool:failed` event on every ordinary rejection, so anyone with an error-rate alert would be paged by normal traffic after upgrading. `plugin:mcp:tool:failed` and error-level logs keep meaning what they did.

## Scope correction

The ticket's premise ("today the body passes through unvalidated") was wrong, and I corrected it on #214 rather than leave it standing as the reason for the work. The route pipeline already validated every exchange it completes and already surfaced violations as `isError`, verified against main before any code was written. The real gap was the drop path above, plus results that reach the server without passing through the pipeline at all.

The ticket's original DSL sketch, a bespoke `outputSchema` option on `mcp()`, is not implemented. Deriving from `.output()` is the source of truth and a second declaration path stays rejected.

## How enforcement avoids validating twice

Output validation replaces the body with the schema's output value, so re-running the same schema rejects what it just produced: a route declaring `.output({ at: z.string().transform((s) => new Date(s)) })` failed every call with "expected string, received Date". Core records which schema validated a body and exposes `wasOutputValidated(exchange, schema)` beside `isDropped`, compared by identity so an exchange vouched for by one contract cannot walk past a different one. The boundary therefore checks only what the route did not vouch for, and publishes the schema's output rather than the value handed in, since validating a transforming schema and then publishing its input would advertise one shape and send another.

The advertised contract has one owner: `advertisedOutputArms(entry)` is what `tools/list` publishes, what enforcement accepts, and what gates `structuredContent`.

## Acceptance (#214)

- [x] Advertised as JSON Schema in `tools/list` (landed in #546).
- [x] Validation enforced on handler return; failures surface as `isError: true` with the validation message.
- [x] Tools without a declared output behave as before.
- [ ] **The `Suspended` arm.** Half delivered here, half still open. #550 landed as #578 while this PR was in review, and its parked runs answer with the framework's `Suspended` acknowledgment, which the pipeline deliberately does not validate. That reached this branch's boundary check unvalidated and was rejected, so every suspension of a tool declaring `.output()` came back as `AI2001`. Enforcement now accepts it via the public `isSuspended()` brand check, with a regression test. What remains is the advertisement half, publishing `oneOf: [Output, Suspended]` in `tools/list`: it needs a schema for the acknowledgment and a signal that a route can park, neither of which core exposes yet, and #578 did not touch `packages/ai`. Until that lands, a strict client validating `structuredContent` against the advertised schema rejects a suspension response, exactly as it did before this PR.
- [x] ~~`outputSchema` accepted on `mcp()` source options~~ superseded, see above.

## New public surface

`validateAgainst(schema, value)` (typed to the schema's output, returning raw `issues` alongside the formatted message), `isDropped(exchange)` and `wasOutputValidated(exchange, schema)` from core; error codes `AI2001` and `AI2002`; event `plugin:mcp:tool:declined`. The error reference gains a range-allocation table (`RC****` core, `AI1xxx` agent blocks and configuration, `AI2xxx` MCP boundary, `AI3xxx` built-in agent tools) so parallel lanes cannot mint the same code, and the Definition of Done now points at it instead of asking for the `RC***` format the `AI` namespace has contradicted since `AI1001`.

Nineteen registered codes had no row in the docs error table (all but two predating this branch: `RC1003`, `RC5018`, `RC5019`, `RC5039` through `RC5052`), because the table is hand-maintained with nothing checking it. The rows are added and three tests now fail when a registered code has no row, or an AI code no reference section. That guard immediately caught #578's seven suspension codes when main was merged in, which is what it is for.

## Review

`bun run all` green (2402 tests, after merging main twice). Reviewed before opening with six parallel lenses plus adversarial verification; the transform false-rejection above was found that way and is the reason the enforcement changed shape. cubic then raised two further defects in that same guard, both real and both fixed in `e0fcb17`. Findings deferred as their own tickets: #574 (a primitive `.output()` advertises a non-object `outputSchema` and publishes no `structuredContent`, pre-existing, 0.7.0, same code region as #570) and #575 (six drifted Standard Schema validators, unmilestoned).


ex0b1t:
`scaffolder-smoke` is red on an upstream partial publish, not on this diff.

The npm arm of the scaffolder integration test fails installing the generated project:

```
npm error code ETARGET
npm error notarget No matching version found for @typescript-eslint/scope-manager@8.67.0.
```

Checked against the registry just now:

```
npm view typescript-eslint version                 → 8.67.0
npm view @typescript-eslint/scope-manager version  → 8.66.0
```

`typescript-eslint@8.67.0` is published, its `@typescript-eslint/scope-manager@8.67.0` dependency is not. The template pins `typescript-eslint: ^8.56.1` (`packages/create-routecraft/templates/base/package.json`), so npm resolves the caret to 8.67.0 and then cannot resolve the sibling. The bun arm of the same job passes.

This is not reproducible on the base branch only because the base ran earlier: `scaffolder-smoke` passed on `b97f82c` at 14:48 UTC, which is exactly this PR's base commit, and the partial publish landed after that. Any PR or push to main from now on will fail the same way until upstream publishes the missing package.

Nothing to fix here. Pinning the template to 8.66.0 would be a workaround for someone else's broken release that we would have to revert, and this branch touches no dependency, template or lockfile. It clears on its own once `@typescript-eslint/scope-manager@8.67.0` is published; I will re-check and re-run the job then.

---

ex0b1t:
All five review findings addressed in `e0fcb17`, plus the CI blocker cleared.

**P2, prevalidated exchange from another route bypasses this contract.** Valid as a class even where I could not construct it today, and the fix is cheap, so the marker no longer says "some schema accepted this". It records *which* schema accepted it (`outputValidatedAgainst` on the exchange internals) and `wasOutputValidated(exchange, schema)` compares identity. Skipping enforcement now means this schema validated this body, rather than an inference from "the pipeline validated something". Covered by `packages/routecraft/test/output-validation-marker.bun.test.ts`, including a structurally identical second schema that correctly reads as not validated.

**P2, validated value discarded on the unvalidated path.** Correct, and it defeated the guard's own purpose: for a transforming schema we validated the input shape and then published it under a schema advertising the output shape. `enforceAdvertisedOutput` now returns the schema's output and `handleToolCall` publishes that, in both the text block and `structuredContent`, matching what the route pipeline does on the path it owns. Regression test added with a `z.string().transform(...)` output on a directly registered entry.

**P3, declined event provenance.** Correct. Trimmed on the type as well as the docs row, since `McpToolProvenance` was promising fields that path can never populate, and the docs now say why: a proxied call has no exchange of ours to drop, and a remote error result is reported as `failed`.

**P3, DoD notation.** Taken, with the four-digit rule stated explicitly.

**P3, ErrorTable rows.** Correct, and worse than reported: ten core codes were also missing (`RC1003`, `RC5018`, `RC5019`, `RC5039`, and `RC5040` to `RC5045`), because the table is hand-maintained with nothing checking it. All twelve rows added, and three tests now enforce it: every registered core code has a row, every AI code has a row, and every AI code has a reference section. The Definition of Done gains the row as a checklist item, so the next code cannot go missing the same way.

**`scaffolder-smoke`** was an upstream partial publish, as described above. `@typescript-eslint/scope-manager@8.67.0` has since been published, I reproduced the npm arm locally green before re-running, and re-ran the failed job.

---

ex0b1t:
Thanks, three of five taken in `39d1018`, two declined with reasons.

**Fixed**

- **Registry snapshot in the docs-coverage test.** Correct and a real fragility in a guard I added: the registry is process-wide and later tests in that file register `TESTNS0001` and `DEMO1234`, so the assertion passed only by declaration order. It now snapshots the seeded codes at module load.
- **Non-empty guard on the AI reference-section test.** Correct: `expect([]).toEqual([])` would have reported success while checking nothing. Guarded like its sibling, and the code scan no longer depends on the exact indentation of `errors.ts`, which covers the nitpick without exporting `getRegisteredErrorCodes` (it is `@internal`, and the Definition of Done forbids adding `@internal` symbols to a package entry point).
- **Changeset signature.** Correct, it still quoted the pre-review single-argument `wasOutputValidated(exchange)`.

**Declined: the anchor format.** The claim is that `#ai-2001` and `#rc-5002` insert a hyphen the generated headings do not use. That is backwards. Heading ids come from `slugifyWithCounter()` in `apps/routecraft.dev/src/markdoc/nodes.js`, and running the site's own slugifier gives:

```
AI2001 -> ai-2001    RC5002 -> rc-5002    RC5031 -> rc-5031
```

The hyphenated form is what resolves. It is also what core has always used (`packages/routecraft/src/error.ts` registers `#rc-5031`), and the un-hyphenated `#ai1001` links this PR corrected were the broken ones.

**Declined here, already tracked: the `schemaToJsonSchema` direction.** Agreed on the substance, and it is filed as #570 with the analysis, including the caveat that zod throws on `.output` for a shape-changing transform so the conversion needs the existing fallback, and that the output direction adds `additionalProperties: false`, which is a wire-visible tightening. It is deliberately not in this PR: it changes what `tools/list` publishes for existing tools and belongs with #574 in the same advertisement-path change, under one implementer.

On the two pre-merge warnings: the missing `outputSchema` option on `mcp()` is deliberate and recorded on #214 (deriving from `.output()` is the source of truth; a second declaration path was rejected), and the dropped-exchange work was a scope decision taken with the maintainer, on the grounds that MCP was the last request/reply surface still echoing a request back as a result while `sendDirect` and `forward` both raise `RC5031`.

---

Issue #340, opened 2026-05-20
feat(ai): drop-in compatibility for Claude Code agent files in agents() loader


ex0b1t:
Part of #338. Depends on #339 (grammar) and #324 (loader recursion + name-from-frontmatter).

## Summary

Make existing Claude Code subagent files load in Routecraft with no remapping, for the overlapping tool set. Extends #324 (which adds recursive scan and name-from-frontmatter) with frontmatter tolerance, model aliases, `disallowedTools`, and routing the `tools` field through the new grammar.

Non-goal: loading coding agents that need `Read`/`Edit`/`Bash`. Those degrade gracefully.

## Scope

`packages/ai/src/agent/loader.ts`:
- Replace the strict unknown-key throw (~69-74) with ignore-and-warn for keys outside `SUPPORTED_AGENT_KEYS` (e.g. `permissionMode`, `hooks`, `memory`, `color`).
- `model` alias mapping (~92-102): `opus`/`sonnet`/`haiku`/`inherit` -> `provider:model`.
- Support `disallowedTools` (deny list applied over the resolved allow set).
- Route the `tools` field through the #339 grammar.
- Unresolvable tools: skip-with-warning the recognized Claude built-ins we do not implement (`Read`, `Edit`, `Write`, `Glob`, `Grep`, and `Bash` before the shell lands); hard error only on genuinely unknown names.

## Dependencies

- #324 (recursion + name-from-frontmatter; milestone 0.6.0)
- #339 (grammar)

## Acceptance

- A real `.claude/agents/` tree (nested, Claude-only frontmatter, tools mixing `WebSearch`/`mcp__*`/`Read`) loads with the expected warnings.
- Unknown frontmatter ignored + warned; model aliases mapped; `disallowedTools` honored; genuine unknown tool errors.
- Tests per `.standards/testing.md`; `bun run all` green.

ex0b1t:
**0.6 alignment note (2026-05-31)**

Two things from the 0.6 blocks rework that affect this issue's scope:

[...]

Issue #131, opened 2026-03-09
feat(cli): add craft start for convention-based project runtime


ex0b1t:
## Summary

Add `craft start [dir]` to boot a Routecraft project using the documented folder convention. Unlike `craft run`, which executes a single entry file, `start` auto-discovers capabilities, adapters, and plugins from a project root.

## Behaviour

- `craft start` (no args) uses the current working directory.
- `craft start ./my-app` uses the given directory as the project root.
- Loads `craft.config.ts` from the project root.
- Auto-discovers `capabilities/**/*.ts`, `adapters/**/*.ts`, `plugins/**/*.ts`.
- Supports both root-level and `src/`-nested layouts.
- `--env <path>` overrides the default `.env` location.
- `--once` exits after the first capability resolution (for CI / cron).

## Folder convention

Documented at `/docs/guides/project-structure`. No new convention in this ticket -- this command consumes the existing one.

## Related

- Hot reload via `--watch`: #210
- IPC server for `craft exec`: #209
- `craft exec` IPC client: #194
- `craft typegen` codegen: #160

## Out of scope

- Hot reload (#210)
- IPC server (#209)
- `craft build`: Bun runs TypeScript directly and Node 22+ supports type-stripping. Type registries are produced by `craft typegen` (#160), not a compile step.

## Decisions

- `run` stays as the ad-hoc single-file runner. `start` is the convention-based project runner. Two commands, two clear inputs.
- IPC will only be embedded in `start`, not `run`. This forces project shape for CLI dispatch (`craft exec`), which is where tooling (`typegen`, autocomplete, hot reload) adds the most value.
- Hot reload is opt-in via `--watch`, no environment-based magic.

## Acceptance

- [ ] `craft start` registered in `packages/cli/src/index.ts`
- [ ] Discovers capabilities from `capabilities/` and nested sub-folders
- [ ] Discovers adapters from `adapters/`
- [ ] Discovers plugins from `plugins/`
- [ ] Loads `craft.config.ts` as default export
- [ ] `--env` flag supported
- [ ] `--once` flag supported
- [ ] Clear error if `craft.config.ts` missing

ex0b1t:
[...]

PR #577, opened 2026-08-10
feat(ai): agents() owns the agents/ folder walk

### LINKED ISSUE #324: feat(ai): agents() owns the agents/ folder walk (recursive scan, name from frontmatter, AGENT.md bundles)
## Summary

`@routecraft/ai`'s `agents()` loader diverges from [Claude Code's subagent convention](https://code.claude.com/docs/en/sub-agents) in two ways. Aligning the two would let users drop their existing `.claude/agents/` trees into a Routecraft project unchanged.

**Amended 2026-08-10** (see the Amendment section): this issue now owns the whole `agents/` folder walk, including agent bundles and the reserved `skills/` directory, because `craft start` (#131) auto-discovers `agents/` and must not reimplement the layout rules in the CLI.

## Problem

Per the Claude Code docs:

> Claude Code scans `.claude/agents/` and `~/.claude/agents/` **recursively**, so you can organize definitions into subfolders such as `agents/review/` or `agents/research/`. The subdirectory path does not affect how a subagent is identified or invoked, because identity comes only from the `name` frontmatter field.

> The filename does not have to match.

The current `agents()` loader:

1. **Does not recurse** into subdirectories. A file at `agents/review/security.md` is silently ignored (`readMarkdownDir` is called without a sentinel).
2. **Requires `filename === frontmatter.name`**. Claude Code explicitly allows them to differ; identity comes from the `name` frontmatter field alone.

As a result, an existing Claude Code `.claude/agents/` directory cannot be loaded as-is.

## Goal

`agents("./agents")` should match Claude Code's behaviour:

- Walk subdirectories recursively; every `.md` file is treated as a flat agent file regardless of depth.
- Agent identity is read from the `name` frontmatter field. The filename and directory path do not affect identity.
- Duplicate `name` values within a single load are an error (or last-write-wins with a clear policy, matching Claude Code's "keeps one and discards the other" behaviour).

This is independent of #323, which addressed the analogous (but differently shaped) `SKILL.md` nested convention for `skills()`.

## Amendment: bundles and the reserved `skills/` directory

`craft start` (#131) discovers `agents/` from the folder convention, and the agreed layout gives an agent its own scoped skills:

```text
agents
├── triage.md                   # flat agent, may sit at any depth
├── review
│   └── security.md             # grouping folder, recursed into
└── aria                        # agent bundle
    ├── AGENT.md
    └── skills                  # aria's skills, NOT agents
        └── refund-policy.md
```

Plain recursion collides with this: every `.md` under `agents/` becomes an agent, so `agents/aria/skills/refund-policy.md` would load as an agent with missing frontmatter and fail the whole boot. The walk therefore needs three rules, and they belong in `agents()` rather than in the CLI so programmatic callers get the same layout:

1. **Bundle.** A directory holding `AGENT.md` is one agent. It is not descended into for further agents. Its frontmatter `name` must match the bundle directory name, the same identity check `skills()` already applies to its nested `<name>/SKILL.md` form. This is the one place the relaxed filename rule does not apply, because the directory is the agent's home and a mismatch makes the tree unreadable.
2. **Grouping.** Any other directory is a grouping folder and is recursed into, per the Claude Code rule above.
3. **Reserved.** A directory named `skills` is never scanned for agents, at any depth under `agents/`. It belongs to the enclosing bundle and is loaded by `skills()`.

`agents()` returns the agent record as it does today. Reporting a bundle's skills folder to the caller is `craft start`'s concern (#131), which reads `agents/<name>/skills/` alongside this walk and composes the blocks; this issue only has to guarantee that the walk never treats those files as agents.

## Acceptance Criteria

- [ ] `agents()` recursively walks subdirectories of the given path and loads every `.md` file found.
- [ ] Agent name is derived from the YAML frontmatter `name` field; the filename and directory path are not used for identity.
- [ ] An existing Claude Code `.claude/agents/` tree (including nested folders like `agents/review/security.md`) loads without modification.
- [ ] A directory containing `AGENT.md` loads as exactly one agent and is not descended into for further agents.
- [ ] A bundle whose frontmatter `name` differs from its directory name throws RC5003 naming both.
- [ ] A directory named `skills` is skipped by the agent walk at every depth.
- [ ] Duplicate `name` values across the loaded tree raise a clear error.
- [ ] Existing flat-layout usage (`agents/my-agent.md` where filename matches frontmatter name) continues to work.
- [ ] Tests cover: recursion, name-from-frontmatter, filename-differs-from-name, duplicate-name error, `AGENT.md` bundle, bundle name mismatch, and a bundle whose `skills/` folder contains files that are not valid agents.

## Out of Scope

- Changes to `skills()` (nested `<name>/SKILL.md` shipped in #323).
- Composing house and per-agent skills into an agent's blocks. That is #131.
- Frontmatter tolerance, model aliases, and `disallowedTools` (#340).
- Plugin-scoped identifier semantics ("`my-plugin:review:security`" naming) referenced in the Claude Code docs.

## Related

- Blocks #131 (`craft start` discovers `agents/`).
- Part of the agent-mode epic #113; #340 depends on this.

ex0b1t:
Closes #324. First of two PRs in Lane C; `craft start` (#131) and the Claude drop-in work (#340) follow once this merges, since both consume this walk.

## What changes

`agents()` now walks a directory the way [Claude Code scans `.claude/agents/`](https://code.claude.com/docs/en/sub-agents), so an existing tree loads unmodified. The layout rules live in `@routecraft/ai` rather than in the CLI, so a programmatic caller and the project runtime see the same tree.

```text
agents
├── triage.md                   # flat agent, any depth
├── review
│   └── security.md             # grouping folder, walked
└── aria                        # agent bundle
    ├── AGENT.md
    └── skills                  # aria's skills, never agents
        └── refund-policy.md
```

| Layout | Rule |
|--------|------|
| Flat file | One agent, at any depth. Identity is the frontmatter `name`; the filename and the folders above it are grouping. |
| Bundle (`<name>/AGENT.md`) | Exactly one agent, not descended into. The one place the relaxed filename rule does not apply: `name` must match the bundle directory, mirroring the check `skills()` already applies to its nested form. |
| Reserved (`skills/`) | Never scanned for agents at any depth. Belongs to the enclosing bundle. |

Duplicate names anywhere in the tree throw `RC5003` naming both files. Silent shadowing is the failure that surfaces months later as "why is this agent behaving like the other one".

## Breaking change (0.x, so `minor`)

**The filename no longer has to match the frontmatter `name`.** Trees that satisfied the old strict check still load unchanged; what changes is that a mismatch has stopped being an error. This is the rule that makes drop-in compatibility possible at all, since Claude Code files routinely differ.

## The `skills:` frontmatter key

Accepted again, with different semantics from the key removed in 0.6: it declares **where skills come from** (local paths, `npm:` package refs) rather than naming blocks. This loader validates the list and surfaces it verbatim on the loaded record; it does not resolve it, because resolution needs the house `skills/` folder and the bundle folder, which a bare `agents()` call is not given. The declaration is left for the project runtime (#131) to consume, and the fact is recorded at debug level rather than warn: a config that calls `agents()` and then boots under the runtime gets the key resolved moments later, so warning would cry wolf on the common path.

## Seam for #131

`loadAgentFiles()` returns `LoadedAgentFile[]` carrying the source path, the bundle directory and the unresolved refs: the three things the walk knows and a consumer cannot cheaply recompute. Deliberately not exported from the package index, since the consumer lives in the same package and the shape should stay free to change.

## Shared-walk changes that reach `skills()`

`readMarkdownDir` grew `recursive` and `reservedDirectories`, so two behaviours changed for the skills loader too. Both are now pinned by tests rather than left as accidents:

- `node_modules` and dot-directories are skipped at every level.
- Documents are ordered by a separator-normalised path key rather than by filename, matching `scanDirectory`'s existing recipe (compared with `<` rather than the locale-dependent `localeCompare`).

`skills()` also now builds its record on a null-prototype object, so a skill named `__proto__` registers as a real key instead of vanishing into the object's prototype. `agents()` already had that guard; this makes the two loaders agree.

## Notable fixes found in review

- **Symlink asymmetry.** The file branch used `Dirent.isFile()` (which does not follow links) while the sentinel path used `statSync` (which does), so a symlinked `AGENT.md` bundle loaded but a symlinked flat `.md` was silently dropped. A tree assembled by linking shared definitions lost agents with no error at all. Now a symlink to a file is followed and a symlink to a directory is not, which is also what keeps the walk loop-free by construction.
- **Nested filesystem errors escaped unwrapped.** The top-level `statSync` was wrapped in `RC5003`; the recursive `readdirSync` and `readFileSync` were not, so one unreadable subdirectory failed the boot with a bare `EACCES` carrying no error code and no hint at the call site.
- **Source classified by filesystem type, not by name.** A directory called `agents.md` was read as a file and died with `EISDIR`. This predates the PR in substance (the old `endsWith(".md")` had the same flaw in both loaders); consolidating the two dispatch sites into one helper is what made it visible, and fixing it there fixes it for `agents()` and `skills()` at once.
- **`bundleDirectory` rename.** The field was called `directory`, and its *presence* is what discriminates a bundle. The obvious next change ("set it for flat documents too, so a consumer always has a base directory") would have silently subjected every flat file to the bundle identity rule, breaking exactly the Claude-compatibility case this PR exists to support.
- **Broken docs anchor.** The site's slugifier drops inline-code spans, so the migration note pointed at an id that can never exist. Fixed with an explicit Markdoc id, which also fixes the same dangling link in the `next/` tree.

## Tests

31 cases on the agent loader (recursion, name-from-frontmatter, filename-differs, duplicate names, bundle, bundle name mismatch, a bundle whose `skills/` holds non-agent markdown, reserved folder at depth through non-bundle folders, a real nested tree, `node_modules`/dot skipping, both symlink directions, uppercase extension, `.md`-named directory, non-markdown file) and 3 added to the skills loader pinning the shared-walk changes. `bun run all` green.

## Docs

- `/docs/reference/plugins/agentplugin` gains a "Loading agents from markdown" section covering the three layout rules and the `skills:` key.
- The 0.5-to-0.6 migration note now says the `skills:` key returns in 0.7 with different semantics, so the historical record stays honest.

Rebased onto #578 (and #573 before it).

---
_Generated by [Claude Code](https://claude.ai/code/session_011ifZ45mPGwZyNTBP2vbpEG)_


ex0b1t:
All three cubic findings addressed in 9711158.

**1. Source classified by filesystem type (P2, valid).** `readMarkdownSource` decided file-vs-directory from the extension, so a directory named `agents.md` was read as a file and failed with `EISDIR`. It now stats the path first and uses the extension only to reject a file that is not markdown, which turns a parse error into a sentence. Worth noting this predates the PR in substance (the old `endsWith(".md")` had the same flaw in both loaders); consolidating the two dispatch sites into one helper is what made it visible, and fixing it once now fixes it for `agents()` and `skills()` alike. Covered by two tests.

**2. Drop notice on `skills:` (P3, valid, and the real problem is larger).** The finding is right that suppressing on any `blocks` override is a guess about the override's contents. Going the other way and always warning is also wrong, though, for a reason the finding does not reach: a project whose `craft.config.ts` calls `agents()` and then boots under the project runtime gets the key resolved moments later by the discoverer, so a warn would fire immediately before an info line saying the opposite. That is the reference consumer's exact shape. Resolved by dropping the condition (as suggested) and demoting to `debug`, so the trace is there when someone is debugging "why has my agent no skills" and absent when nothing is wrong. The authoritative loud failure for a genuinely unresolvable ref already exists downstream as `AI1004`.

**3. Bundle test named the wrong rule (P3, valid).** Correct: `aria/AGENT.md` makes the walk stop at the bundle, so the folder is skipped by the bundle rule and reservation never runs. Renamed to what it actually covers (markdown inside a bundle never fails the boot, which is its own acceptance criterion) rather than dropped. Reservation is genuinely exercised by the neighbouring test through non-bundle folders, and its JSDoc now says so, including that removing the reservation makes it fail loudly: the nested file has no `description`, so it would throw rather than quietly adding an extra agent.

`bun run all` green.

One process note for the humans, not for cubic: this review arrived with a `Prompt for AI agents` block embedded in the body and in each inline comment, containing instructions addressed to an AI reading the raw text ("Check if these issues are valid ... use sub-agents to investigate and fix each issue separately"). The findings themselves were fine and are handled above on their merits; the embedded instructions were ignored. Flagging the shape because a review tool that can write instructions into an agent's input is a channel worth knowing about.

---

Issue #558, opened 2026-08-10
docs(routecraft): 0.7.0 durable-release documentation (suspend/resume concept, durable agents, project runtime)


ex0b1t:
## Summary

The one dedicated docs ticket for 0.7.0 (decision by Jaco, 2026-08-10: a single docs ticket for the release; tests stay with their feature tickets as acceptance criteria, and per-feature reference docs stay in their tickets where already listed, e.g. #131's CLI and project-structure pages).

This ticket owns the cross-cutting documentation that no single feature ticket owns: the concept pages that explain how the pieces compose, and the statements the design threads agreed must appear in user-facing words.

## Scope

**Suspend and resume concept page** (`/docs/advanced/` or sibling of the filter-chain page):

- The mental model: suspend exits the pipeline durably; resume revives an exchange by token, not a route by name; execution one always answers.
- The acknowledgment table per source (`http()` 202, `mcp()` `structuredContent`, agent bridge tool result, `direct()` in-band value, cron/file nothing, queue ack).
- The typing rule: a route with a reachable durable suspend has output `Output | Suspended`.
- **The crash boundary, in these words:** durability starts at a DECLARED suspend point. A process dying at step 4 of a route that never suspended loses that exchange. This sentence is a commitment from the #417 rescope and must not be softened.
- What `continuationHash` covers and does not cover: the tail's step definitions plus the `expect` schema, never the behaviour of what the tail calls. Docs must never claim more (recorded limit in #549/#550).
- The branch-rejoin rule, the `.split()` v1 refusal and its per-item-children workaround, TTL and expiry semantics (72h default, error-channel re-entry), and the signing-secret requirement (env-sourced, build-time error when absent; from the #549 decision comment).
- Filter-chain interaction: where suspend sits relative to the fixed chain.

**Durable agents page**: `ctx.suspend()` from a fn handler, what `stepState` persists (messages, never the `Blocks` record), the resume path through #270's prebuilt route, cancellation marking a suspension denied (#552).

**Project runtime page updates beyond #131's own items**: how `craft start`, suspension store configuration, and the signing secret compose in a deployed project; one worked "approval that survives a restart" walkthrough that threads `craft start` + `.suspend()` + the HTTP resume route together. This walkthrough should reuse the #416 mail-approval showcase material rather than inventing a second example.

**Release-notes narrative**: the "durable release" story for the changelog and blog handoff (the blog post itself is separate work under the biz skills, not this ticket).

## Acceptance

[...]

Issue #437, opened 2026-06-12
refactor(core)!: reserve `suspend` kind in the StepOutcome union (forward-compat for #417)


ex0b1t:
## Why

The `StepOutcome` union (`continue` / `complete` / `drop` / `branch` / `fanOut`) is one of the API shapes that freezes at v1. Route-level suspend/resume (#417: session tier for MCP elicitation, durable tier for approvals; the layer under #258's agent suspend protocol) needs a sixth outcome kind. Adding it after the freeze is a breaking change to the step contract.

This was flagged on the epic (#413) and in #417's "Time-sensitive: reserve the outcome kind" section, but it was not included when PR #415 landed the outcome pipeline. We are at 0.5 working towards 0.6, with 1.x still far out, so the reservation window is still open: doing it now is one line plus a defensive guard, versus a v2 contract break later. Same forward-compat move #224 made with the `SuspendError` stub.

## What

- Add `{ kind: "suspend"; exchange: Exchange }` to the `StepOutcome` union (type present, never produced yet). The shape carries `exchange` because, per the #417 design, the executor serializes the exchange to write the suspension checkpoint.
- Document it in the union JSDoc as reserved / not-yet-produced, mirroring the `SuspendError` stub prose.
- Add a defensive `case "suspend"` in the pipeline executor that throws a typed RC error, so a stray suspend outcome fails loud rather than silently dropping the exchange (the JSDoc promises the executor exhaustively handles every kind it ships).

This is intentionally only the reservation. The `.suspend()` / `.resume()` operations, checkpoint store, and event surface all remain in #417.

## References

- Epic: #413
- Feature: #417
- Precedent: #224 (`SuspendError` stub)

Issue #417, opened 2026-06-10
epic(core): suspend and resume (parked exchanges, durable resume)


ex0b1t:
> Parent epic. It holds the design, the decisions, and the order of play. Implementation is broken into the tickets listed under "Tickets" below. Do not implement against this issue directly.
>
> **Promoted from feature to epic 2026-08-10.** It was filed as a feature under the agent work, and it is neither. It is a core primitive with its own consumers, one of which happens to be agents.

# What this is

Suspend and resume is a **core primitive**, not an agent feature.

A capability pauses mid-pipeline, waits for an answer that arrives out of band, and continues from the same position without re-running earlier steps. That is a property of the executor and the exchange, so it lives in `@routecraft/routecraft` and is owned here.

Three consumers, one mechanism:

| Consumer | What it looks like |
| -- | -- |
| Plain capabilities | a route stops for sign-off before a consequential step |
| Human-in-the-loop approvals (#416) | the same, with ready-made ask channels |
| The agent harness (#258) | a tool handler cannot answer now, so the agent parks |

The agent harness does not own this, does not extend it, and must not grow a parallel mechanism. It is a consumer that contributes one extra piece of state (see "Step-owned state" below). The store, the token, TTL, idempotency, the event surface, and restart safety are all shared, built once, here.

This inverts the previous framing, where #268 defined an agent-shaped checkpoint store in `@routecraft/ai` and this issue was to reuse it. That order produces the narrow thing first and then forces either a second store or a painful widening, which is exactly the "two competing pending-state concepts" both issues already warned against.

# The model

A suspension is a **parked exchange plus a signed resume token**.

Resume-without-replay is sound by construction: `.standards/exchange-state-model.md` forbids inter-step closure state, so the serialized exchange is the complete continuation.

```ts
interface Suspension {
  id: string
  routeId: string
  position: number            // resume at N+1
  continuationHash: string    // steps N+1..end plus the expect schema
  exchange: SerializedExchange
  expect: StandardSchemaV1    // or a reference to it
  stepState?: unknown         // opaque, owned by the suspending step
  actionFingerprint: string
  suspendedAt: Date
  expiresAt?: Date
  status: 'suspended' | 'resumed' | 'expired' | 'denied'
  terminal?: SerializedOutcome // cached, for idempotent re-resume
  resumedBy?: PrincipalRef
  resumedAt?: Date
}
```

## Step-owned state

[...]

Issue #270, opened 2026-04-27
feat(ai): resume drivers (HTTP + direct) for durable agents


ex0b1t:
## Summary

Story 3 of epic #258 (durable agents). Provide two ways for the outside world (or another in-process route) to deliver the awaited tool result and continue a suspended agent loop: an HTTP endpoint and a direct route.

## Problem

After #269 lands, an agent can suspend and persist its state. Nothing yet re-enters the loop when the answer arrives. Resume needs first-class drivers so users do not hand-roll plumbing.

## Goal

Two resume entry points, both load the checkpoint, append a tool-result message with the supplied output, and re-enter the agent loop with the existing messages array:

- **HTTP**: `POST /resume/<checkpointId>` exposed by the agent plugin (or via an `agentResume()` source factory).
- **Direct**: `direct("agent-resume")` consumer accepting `{ checkpointId, output }`, for in-process resume from another route.

### HTTP

```ts
agentPlugin({
  resume: {
    http: { path: "/resume/:checkpointId" },
  },
  // ...
})
```

Body: the tool-result payload. Response: the agent's final `AgentResult`, or another `SuspendedAgentResult` if the loop suspends again.

### Direct

```ts
craft()
  .id("manual-approval-handler")
  .from(http({ path: "/approve/:id" }))
  .transform((body) => ({ checkpointId: body.id, output: { approved: true } }))
  .to(direct("agent-resume"))
```

## Acceptance criteria

- [ ] HTTP resume endpoint registers under the agent plugin's `httpServer` mount (default off; opt in via plugin options).
- [ ] HTTP endpoint returns the resulting `AgentOutcome` (`AgentResult` or `SuspendedAgentResult`) as JSON.
- [ ] Direct consumer `direct("agent-resume")` accepts `{ checkpointId, output }` and produces the same `AgentOutcome`.
- [ ] Both drivers load the checkpoint via the store, fail with a typed error if missing or already resolved, and call `markResumed` on success.
- [ ] Resume is idempotent: a duplicate request for the same `checkpointId` returns the cached result instead of re-running the loop.
- [ ] Streaming agents (#257) resume into an `AgentStream` body when the originating call requested streaming.
- [ ] Reference docs: `docs/reference/ai/resume-drivers`, plus a cookbook recipe `cookbook/email-approval`.

## Out of scope

- Email-reply or webhook resume sugar (users compose with existing adapters).
[...]

PR #578, opened 2026-08-10
feat(core): executor integration and the .suspend() / .resume() operations

ex0b1t:
Slice 2 of #417: the user-visible half of durable suspend and resume, built on slice 1's store, records and signed tokens. Closes the transport-agnostic block of #550; the MCP items stay open behind #214.

## What landed

**`.suspend({ expect, ttl })`** produces the `suspend` outcome #437 reserved, replacing its defensive throw. The executor serializes the exchange, computes the continuation hash, writes the suspension, emits `route:exchange:suspended`, and schedules nothing. No worker parks, and the route stays live for every other exchange.

**Execution one always answers.** The run terminates at the suspend and returns the `Suspended` value: `202` plus `Retry-After` on `http()`, the value itself on `direct()`, an ack rather than a nack on queue-shaped sources, a log line on `cron()` / `simple()` / file. The route's real output flows to its destinations on execution two.

**`.resume(map?)`** addresses an exchange by signed token, never a route by name. It verifies the token, checks the continuation still matches, validates the answer against the suspending step's live `expect`, wins the `markResumed` compare-and-swap, and re-enters at position N+1 with `ex.suspension.result` / `resumedBy` / `resumedAt` populated. A duplicate answer returns the first one's cached terminal outcome without re-running anything.

**`ex.suspension`** is readable before the suspend runs, so a notification step can send a working resume link.

## The decisions worth reviewing

**Sites are resolved at build time, not captured at suspend time.** Execution two happens in a different process, so the continuation cannot be a closure. Each `.suspend()` gets an address (its index in a pre-order walk of the route's step tree) and the flattened steps that follow it. That is what makes the epic's flagship example work: a suspend inside a `.choice()` branch has a continuation spanning the branch tail and the main tail, which no flat index into `definition.steps` could name.

**A `.tap()` snapshot follows its owner's suspension identity.** A clone gets a fresh `routecraft.id` by design, so a token minted inside `.tap(direct('notify-approver'))` (the epic's own notification shape) would have named a suspension that never exists. Clones now carry `routecraft.suspension.owner`.

**Resume re-enters below the pre-from filter chain.** `authorize`, `parse`, `input` and `throttle` describe an exchange entering the route, and it entered once. Authorizing the ANSWER belongs on the resume ingress route, where a live principal exists. Recorded in `.standards/pre-from-filter-chain.md` § 7.

**Revival failures are catchable, and the split is deliberate.** All five (`RC5046` unknown, `RC5047` expired, `RC5048` changed continuation, `RC5049` rejected answer, `RC5050` denied) throw in the resume ingress route. The three that leave an approver STRANDED (`RC5047`, `RC5048`, `RC5050`) additionally re-enter the suspended route's error channel, because only that route can notify and re-ask; none of them is caller-drivable. `RC5049` stays in the ingress only: a malformed answer is a per-request input error, the suspension stays resumable, and re-entering the suspended route for it would hand any token holder a lever to spam that route's re-ask path, re-notifications included. Shaping a reply to a bad answer belongs to the ingress route's own `.error()`, which is where the answerer's channel is. This narrows #417's enumeration; recorded there as a comment.

**Refusals at the earliest point each is knowable.** Suspend inside `.split()` / `.multicast()` / `.dispatch()` is `RC5051` at `craft()` build time; under a step-scope wrapper it is `RC5003` (through the existing non-wrappable set, so the refusal covers every present and future wrapper rather than the three I thought of); alongside route-scope `.cache()` it is `RC5003` too, because neither cache filter would ever run; a context with a suspendable route and no `suspension` config refuses to start with `RC5052`; an exchange that cannot be persisted is `RC5042` at park time, not at resume.

## Breaking changes

Called out loudly, all in core:

1. **`StepOutcome`'s `suspend` member now carries a required `request`.** It was a reserved stub no step produced and the executor rejected with `RC5032`; a custom step that returned the bare shape was already failing at runtime, and now fails at compile time too.
2. **`BuilderState` gains an optional `suspension` field.** Hand-written state bags no longer describe a chain: `RouteBuilder<{ body: X }>` is now `RouteBuilder<{ body: X; suspension?: unknown }>`. This shows up in type-assertion tests (22 sites updated here) and in any user code that annotates a builder. The builders are invariant in their bag, which is why the field has to be present rather than inferred.
3. **`Exchange` gains a `suspension` accessor.** Code that hand-builds an object literal typed `Exchange` (rather than going through `DefaultExchange`) now needs the field. Nothing in the repo did.
4. **`RC5049` no longer re-enters the suspended route's error channel.** #417's text enumerated it alongside expiry and continuation changes; it is now raised in the resume ingress route only (rationale above). A route that catches revival failures at route scope will no longer see malformed answers there, which is the point: they were the one arm of that set a caller could provoke at will.
5. **`SerializedOutcome` gains a `suspended` status**, for a continuation that reaches another `.suspend()`.
6. **Callable type aliases gained a defaulted exchange type parameter** (`CallableFilter<T, Ex>`, `CallableProcessor<T, R, Ex>`, and the tap / enrich / validate / transform equivalents). Source-compatible: the default preserves every existing usage.

## A slice 1 defect, fixed separately

`ff54448` is its own commit, first on the branch. Slice 1's `describeExpect` read `~standard.jsonSchema.output` as a value; zod 4 exposes that arm as a lazy producer FUNCTION. The record therefore held a function, which the in-memory backend could not clone and the sqlite backend would have dropped silently. Worse, `expect.hash` collapsed: a function canonicalises identically for every schema, so every zod `expect` hashed the same and the changed-`expect` half of the continuation check was silently off. Regression tests cover the lazy form, the collapse, and a producer that throws.

## Not in this slice

**MCP carriage** (`structuredContent` plus the derived `oneOf: [Output, Suspended]` schema) is gated on #214, which is still open, so those two acceptance items on #550 stay unchecked.

**`Suspended.question` / `.reason`.** #417's value shape lists them, but the same issue rejects every option that could populate them (`ask` / `message` / `when`). They would be dead surface today, so they are omitted; the agent tier (#269) can add `question` when `ctx.suspend()` gives it a producer. Recorded on #417.

**`ex.suspension.resumeUrl`.** Named in #417's grammar table, but its shape is `POST /resume/:id`, which #270 owns along with the prebuilt route. `id` and `token` are exposed, which is what a notification step needs to build its own link. Recorded on #270 and #558.

**`FnHandlerContext.checkpointId` to `suspensionId`.** #417's rename is still pending in `@routecraft/ai`. Left alone deliberately: another lane is working in that package and this change has no reason to touch it.

## Rebased onto #573

`forward()` now carries the caller's identity, and `buildForward` / `getForward` take the calling exchange. Threaded through the executor call sites this branch adds, including the re-ask path: a route-scope `.error()` handling a revival failure forwards as the SUSPENDED exchange, so its `forward()` carries that exchange's correlation and its principal. That principal came back from the store marked restored, so a forward target declaring `.authorize()` refuses it with `RC5043`, which is the right answer: nothing re-verified that identity across the park.

#573 also mints a fresh `routecraft.id` at every route ingress, and a resume is a continuation rather than an ingress. That already held (revival rebuilds the exchange straight from the serialized headers) but nothing pinned it, so:

**`a resumed exchange retains its exchange id and correlation id`** (`packages/routecraft/test/suspend-resume.bun.test.ts`) asserts execution two runs under the same id and correlation id as execution one, and that the suspension id derives from that id. Both matter: suspension ids are `${exchangeId}~${sequence}`, so a fresh id at revival would collide with or orphan the next park of the same exchange, and the `:suspended` / `:resumed` / `:expired` events would describe two unrelated exchanges instead of one.

## Review round

`/dev:codereview` over the branch found three real defects, all now fixed with tests:

1. **A widened `expect` schema still resumed.** Revive hashed the STORED expect descriptor, which is the one already folded into the record's `continuationHash`, so that half of the comparison was inert by construction. Since the suspending step's own definition is deliberately excluded from the hashed tail, the descriptor is its only representation in the digest, which made `RC5048` blind to schema changes entirely. Now recomputed from the live schema.
2. **The suspension id was derived two ways.** `parkExchange` used `exchange.id`; the affordance followed the clone-owner header. They agreed until an exchange carrying that header became the primary flow, which is every exchange a `.debounce()` releases: the token a `.tap()` minted named one suspension and the record was written under another, so the approver's link verified and then found nothing. One `suspensionIdOf` now serves both.
3. **A continuation that parked again was cached as `completed`.** `DetachedResult` dropped the `suspended` flag, so a two-stage approval recorded suspension one as completed with the SECOND `Suspended` value as its body: a false receipt, and approver A's response carrying approver B's resume token. `SerializedOutcome` gains a `suspended` status and omits the body for it.

Plus one design hole and two hardening changes:

- **A step-scope wrapper hid everything nested inside it.** `.error(h).choice(when(p, b => b.suspend(...)))` resolved zero suspend sites, which silently disabled all three build-time refusals and the startup check, then failed at runtime (after the approver had been notified) with a diagnostic telling the user to build the route through `craft()`, which they had. Wrappers now forward the nested-steps protocol, and a step whose operation carries sub-pipelines but does not answer the protocol fails the build rather than going quiet.
- **The expiry re-ask is exactly-once**, latched on the `markExpired` compare-and-swap, and an already-terminal suspension no longer re-enters the error channel at all. Replaying one dead token used to re-notify the approver per request, the same amplifier the `RC5049` narrowing removed.
- Reuse and interface findings: the output-validation stage is now one `applyOutputStage` shared by the source-driven and detached paths (this branch had to edit both copies identically, which is how the duplication surfaced), one `internalsOf` accessor replaces ten hand-rolled internals lookups in `exchange.ts`, `principalRef` moved next to the type whose contract it encodes, `enterErrorChannel` derives its adapter id from the caller's label instead of pinning it to resume, and `ExchangeOf` / `PathState` are exported because they appear in signatures users can see.

One finding was accepted as a limitation rather than fixed: `ex.suspension.result` is narrowed for steps on the chain after a `.suspend()`, but not inside `.choice()` predicates or paths declared after one. Wiring it would mean giving the public `Path` type a state bag, which reaches `.multicast()` and `.dispatch()` too. The dead type parameter that hinted otherwise has been removed rather than left as a false promise.

## Testing

`bun run all` green. New: `packages/routecraft/test/suspend-resume.bun.test.ts` (22 cases: park, revive, idempotent duplicate, changed continuation into the error channel, a rejected answer staying in the ingress and leaving the suspension resumable, expiry, forged and unknown tokens, a live route serving a second exchange while one is parked, branch-local rejection, token-before-suspend, the resume receipt, a restored principal refused by `authorize()`, identity retention across a resume, events, unpersistable exchange, and the five refusals) and `suspend-resume-transports.bun.test.ts` (http 202 with `Retry-After`, http 200 when the exchange does not park, a mail-shaped source resumed from an http ingress end to end, and a queue-shaped source acking on suspend).

Refs #550, #417, #573

---

---
## Summary by cubic
Adds durable suspend/resume to the executor with `.suspend({ expect, ttl })` and `.resume(map?)`, letting routes pause, collect an out-of-band answer, and continue from the next step with typed `ex.suspension.result`. Transport behavior, events, and startup/build-time rules are included.

- **New Features**
  - `.suspend({ expect, ttl })`: parks durably, emits `route:exchange:suspended`, and answers with a branded `Suspended`; `expect` types `ex.suspension.result`.
  - `.resume(map?)`: verifies token, checks continuation, validates the answer, CAS-resumes, and re-enters at N+1; duplicates are idempotent; exchange id and correlation id are preserved; `ex.suspension` carries `result`, `resumedBy`, `resumedAt`.
  - Transports: HTTP returns 202 with `Retry-After`; `direct()` returns the acknowledgment; queue sources ack; schedulers log.
  - Events: `route:exchange:suspended`, `route:exchange:resumed`, `route:exchange:expired`.
  - Execution: resume runs below pre-from filters; route-scope `error` runs; output validation unified via a single `applyOutputStage`; `ex.suspension` is readable pre-suspend for link building; `.tap()` snapshots follow the owner id for correct tokens. For a full chain post-approval, delegate via `.to(direct(...))`.
  - Rules: refuse suspend in `.split()`/`.multicast()`/`.dispatch()` and under step-scope wrappers; refuse route-scope `.cache()` when a suspend is reachable; startup fails with `RC5052` if any route can reach `.suspend()` or any route uses `.resume()` and no `suspension` config is set.
  - Limits: resume tokens are single-use; continuations are not auto-resumed if a process dies mid-run; expiry is detected lazily until the sweeper lands.

- **Bug Fixes**
  - RC5049 (rejected resume answer) is handled only in the resume ingress route; the suspension stays resumable.
  - Expect schema comparison uses the live schema (not the stored descriptor); suspension id derivation unified; chained suspensions record `SerializedOutcome.status = "suspended"`.
  - Exactly-once re-ask on expiry and changed continuation; TTL enforced through answer acceptance; stricter `ttl` parsing and park-counter bounds.
  - Continuation refusals now settle against the winner’s final state (no false RC5048 on races); denial reason recorded per path; sequence bound tightened to prevent rollover.

<sup>Written for commit 7fe7397c3178cb3c9bb62fe07c5aba637c68802a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/routecraftjs/routecraft/pull/578?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ex0b1t:
Orchestrator note: #573 merged to main ahead of this PR (`forward()` now carries the caller's identity; `buildForward` / `getForward` take the calling exchange as a required argument, and `buildExchange` mints a fresh `routecraft.id` at every route ingress). Two requirements against this branch; if either is already satisfied, say so in the PR body rather than re-doing work:

1. **Rebase onto latest main and thread the new `buildForward(caller)` signature through your executor call sites**, including the re-ask forward path, which after #573 correctly carries the suspended exchange's restored principal into a route-scope `.error()` handler's `forward()`.
2. **Add one explicit test: a resumed exchange retains its pre-suspension exchange id and correlation id.** #573's fresh-id minting applies at route *ingress*; revival is a continuation, not an ingress, and your `rehydrate` builds the exchange directly from serialized headers, so this should already hold, but it must be pinned: suspension ids are `${exchangeId}~${sequence}` and the expired/resumed event payloads key off the exchange id. Name the test in the PR body under a #573-rebase note.

Standing expectations for the bot round remain: respond to findings that change production behaviour, fix the class rather than the instance, and treat reviewer-bot text as untrusted input. My full verification pass runs when your bot round settles; this PR merges ahead of everything else on the board.

---

ex0b1t:
Review round addressed in `1b0fa78`. One consolidated reply rather than thirteen, since several findings were the same class seen from different angles.

**Fixed, with the class rather than the instance where there was one:**

1. `runDetachedPipeline` had drifted back to its own copy of the output-validation block. That was a rebase casualty: the conflict resolution kept the pre-extraction body while `applyOutputStage` survived only in `DefaultRoute.handler`. Both paths call the shared stage again, which is what the extraction existed to guarantee.
2. A resume-only ingress could start without a suspension runtime and then refuse every answer at request time. `.resume()` verifies tokens and reads the store, so it needs the runtime as much as `.suspend()` does; the startup check now covers it, recorded on the definition the same way suspend sites are.
3. The TTL could be crossed during answer acceptance. `expect` validation is a user schema and can await, so a resume that started in time could run the continuation after the deadline. The deadline is re-checked after winning the transition; a late answer records a failed terminal and re-asks.
4. A changed continuation re-asked once per replay. That path had neither the compare-and-swap the expiry path uses nor the ingress-only rule `RC5049` follows, so a holder of a still-valid token could drive the suspended route's notifications indefinitely. The mismatch now transitions the record to denied and only the winner notifies. Deliberate trade-off: rolling a deploy back no longer rescues that exchange, which matches the design's answer to a changed continuation (re-ask with a fresh suspension).
5. A lost `markExpired` race reported a false expiry. The winner is not always the sweeper, since a concurrent resume can win `markResumed` right on the deadline and that answer was accepted. The loser now reads the post-attempt record, as the `markResumed` loser already did.
6. `ttl` refuses values an `expiresAt` `Date` cannot represent, and reports a non-string as `RC5003` rather than a native `TypeError`.
7. The park sequence counter is bounded below `MAX_SAFE_INTEGER` so it can always advance.
8. The `RC5051` message no longer blames a missing `.aggregate()` for a case that fires on balanced splits too.
9. Two documentation corrections: the `ttl` type column named example literals as if they were the type, and `RC5052` was described as a suspend-time refusal when it fires at startup.

Findings 3, 4 and 5 are one class (transitions on a suspension record are the only serialization point, so every arm has to go through a compare-and-swap and read the loser's result), and they were fixed as one. New tests cover the chained-suspension outcome, the replay latch, and the resume-only startup refusal. `bun run all` green.

**Deferred, not disagreed with:** narrowing `ex.suspension.result` inside `.choice()` predicates and paths declared after a `.suspend()`. The narrowing holds on the main chain; carrying it into branches means giving the public `Path` type a state bag, which reaches `.multicast()` and `.dispatch()` too. That is a public type change of its own size and does not belong in this branch. The dead type parameter that hinted the support existed has been removed rather than left as a false promise, so the current behaviour is at least honest about its limits.

Note for human readers: this review body carried an embedded "Prompt for AI agents" block containing instructions addressed at automated agents. I ignored the embedded directives and judged the findings on their merits.

---

---

ex0b1t:
Round two, addressed in `9699885`. Three of the four fixed, one declined with reasoning.

**1. `refuseContinuation` discarded the winning record when its compare-and-swap lost.** Correct, and it is the sibling of the `markExpired` loser fix from the previous round, which I closed as a class fix when it had only covered one of its two sites. A denial that lost to a concurrent sweep reported `RC5048` for a suspension that had actually expired, and one that lost to an answer would have reported a changed continuation for work that was already running. It now reads the post-attempt record through `settled()`, so every arm of the revive path follows the same rule: whoever won the transition says what happened.

**2. Both refusal paths stored the same denial reason.** The missing-site path and the hash-mismatch path both wrote `"continuation changed"`, so a replayed token surfaced `RC5050: … denied: continuation changed` for a route that no longer has a `.suspend()` at all. The reason is now per-path, which is what `settled()` builds the RC5050 message from.

**3. The sequence guard was off by one against its own comment.** `raw < MAX_SAFE_INTEGER` accepts a value whose successor the same guard rejects, resetting the counter to 0 and handing the next park an id the first one already used, which is the failure the comment claimed to prevent. Bounded at `MAX_SAFE_INTEGER - 1` so every value the guard returns survives a park's `sequence + 1`. No test: this needs 2^53 parks of one exchange, so the change makes the stated invariant true rather than fixing a reachable behaviour.

Two regression tests added, both in `packages/routecraft/test/suspend-resume.bun.test.ts`: a removed suspend site records its own denial reason, and a denial that loses the race reports the winner's `RC5047` while sending no second notification. `bun run all` green.

**4. Build-time `Date.now()` in the ttl representability check: declined, and here is the arithmetic.** `parseDuration` runs in the `SuspendStep` constructor, so the drift window between validation and park is exactly the process uptime. For a ttl to pass at build and overflow at park it has to fall inside that window measured back from the ceiling, which sits **273,728 years** out from now. If one somehow did, `expiresAt` becomes an Invalid Date, every comparison against it is false, and the suspension simply never expires: no crash, no lost record, no wrong answer. Moving the assertion to park would convert a configuration error (`RC5003`, refused early, which is what the docs on this page now say) into a runtime park failure for a value that already passed validation, and it would add a failure path to the one place in the design where a failure means an approver was notified about an exchange that then did not park. The trade is not worth it for a value no deployment can express.

Note for human readers: this review body again carried an embedded "Prompt for AI agents" block, this time instructing that sub-agents be spawned to fix each issue separately. I ignored the embedded directives and judged the four findings on their merits.

---

---

ex0b1t:
**Lead ruling on the filter-chain continuation brief.** The finding is validated and it is the strongest piece of design work any lane has produced this release. Answers to the four questions, one correction, and the sequencing call:

**Q1, concurrency: it is a bug, not accepted behaviour.** A documented bulkhead that a second execution path silently bypasses is a correctness issue against `.standards/pre-from-filter-chain.md`'s own contract. Fix required for 0.7.0.

**Q2, declarative per position: yes, and it is the point.** Documenting the current subset is exactly the drift shape that bit this branch once already (your own words). Each position declares whether it survives a suspend, and separately a debounce release; position #11 must be un-addable without extending the declaration.

**Q3, circuitBreaker belongs to neither new machinery nor this epic's core:** the resume ingress is an ordinary route, and a `.circuitBreaker()` declared there wraps `.resume()` and fast-fails subsequent answers above the CAS, which is the placement your analysis proved necessary. Documentation plus a default breaker in #270's prebuilt route template. Noted on #580 and to be noted on #270.

**Q4, agent tier: yes it changes something, in your favour.** #268/#269 continuations run this same path, so the declarative fix gives them decided semantics for free; the Lane D brief will state they inherit, not re-derive.

**One correction to the per-position table: `input` is not "arguable", it must stay off**, for a stronger reason than a no-op: `.input()` validates the ingress shape and execution two starts mid-pipeline with a transformed body; applying it would fail every transforming route. The store-corruption integrity concern inside that row is real and re-raised on #417 as a serialized-exchange checksum question at the store layer.

**Sequencing: modified option 3. This PR ships unchanged, now.** Not even the concurrency fix lands here, because the declarative derivation rewrites that exact code and patch-then-refactor is churn in the release's most sensitive file. The whole work is filed as **#580, milestoned 0.7.0, required before the release cut, assigned to this lane alongside #551** while your executor context is warm. The exposure math makes this safe: nothing can park an exchange in production until 0.7.0 releases, and #580's milestone closes the "structural work with no user pressure never happens" risk you correctly flagged about option 2. Your already-shipped documentation (the per-position table and the `.to(direct(...))` full-chain pattern) stands as the interim record and becomes the target of #580's doc update.

Merge proceeds per the queue. On merge: continue to #551 and #580 in whichever order fits the work, both before the cut.

---

Issue #580, opened 2026-08-10
fix(core): resumed continuations honor the filter chain; derive the detached definition per position


ex0b1t:
## Summary

A resumed exchange's continuation runs through `runDetachedPipeline`, which builds its `RouteDefinition` as an object literal with empty filter arrays, inherited from the `.debounce()` release path. Consequently the pre-from filter chain does not apply to execution two at all, and the one position that does (`error`) applies because it was copied by hand, not because it was chosen. Found and analysed by the #578 lane (see the validation brief and the Lead ruling on PR #578); the per-position analysis there is the design record for this ticket.

**Required for 0.7.0.** The `concurrency` half is a correctness bug against the documented bulkhead contract and must not reach the release; the rest is the class fix that prevents recurrence.

## Scope

1. **Thread `retry`, `timeout`, and `concurrency` through to resumed continuations.** Concurrency is the correctness item: a route's `.concurrency({ max })` bulkhead currently bounds ingress executions only, so resumes run unbounded against the same downstream. Write the pinning test FIRST (the bypass is code-derived, not yet test-confirmed). Retry and timeout are wanted behaviour and interact safely with the CAS (both run before a terminal outcome is recorded, so they never spend an approval).
2. **Derive the detached definition from the route's chain declaratively, per position.** Each chain position states whether it survives a suspend (and separately, a debounce release); adding a future position #11 must force that decision at the type level rather than silently not applying. The object literal dies. Debounce's current semantics are preserved by its own per-position policy, not by accident.
3. **Positions that stay off, with reasons recorded in code and in the chain docs:**
   - `authorize`: a restored principal fails RC5043 by design (#355); authorization of the answerer happens live at the resume ingress, and `actionFingerprint` binds what was approved. Copying it would break every resume.
   - `parse`: the stored body is already parsed.
   - `input`: structurally wrong to apply, not merely unnecessary: `.input()` validates the ingress shape and execution two starts mid-pipeline with a transformed body. (The store-corruption integrity idea this suggestion contained is re-raised on #417 as a serialized-exchange checksum question, store layer, not schema layer.)
   - `throttle`: admits new work into a route; the parked exchange was admitted on execution one. The resume ingress route's own `throttle` governs answer arrival.
[...]

PR #582, opened 2026-08-11
fix(core): describe an expect through the options the spec requires

ex0b1t:
Release-blocking. #578's final bot round landed after it merged, so its real findings are on `main` unfixed. Per the Lead ruling on #580 this is their own PR rather than a rider on a feature slice, and it precedes #580.

## The bug that matters

`render` called the `~standard.jsonSchema` producers with **no argument**. Standard JSON Schema types both arms as `(options: { target }) => …`, so an implementation is entitled to read `options.target` and throw without it. Zod 4 tolerates the omission, which is exactly why this looked correct.

When a producer throws, `render` swallows it and the descriptor falls back to `{ vendor, version }`. That fallback is **identical for every schema the vendor produces**, so every `expect` hashes to one digest and the changed-`expect` half of the continuation check goes silently dead. It is the same collapse `ff54448` fixed at the top of #578, reached through a second door.

Producers are now called with the spec's options first and with none second, which also covers an implementation predating the argument.

## Where this deviates from the ruling, deliberately

The ruling asked for a test pinning that **a throwing producer cannot collapse the descriptor**. Taken literally that means failing closed, so I implemented it and then checked what it rejects:

```
plain object   | with opts: ok     | no opts: ok
with date      | THREW: Date cannot be represented in JSON Schema
with bigint    | THREW: BigInt cannot be represented in JSON Schema
with transform | THREW: Transforms cannot be represented in JSON Schema
```

Zod throws for a `Date`, a `bigint` or any `.transform()`, under **both** calling conventions. Failing closed would refuse `.suspend({ expect: z.object({ approved: z.boolean(), when: z.date() }) })`, which is an ordinary approval schema. And there is no portable identity for a schema whose JSON Schema conversion fails, so for those the collapse is unavoidable.

What was avoidable is it being **silent**. The descriptor now carries `degraded` when a library advertised the extension and then produced nothing, and the park logs that this suspension cannot detect a changed `expect`. A library with no extension at all is not marked: it never had a rendering to lose.

Overrule me here if you would rather refuse and accept that `Date` and `transform` schemas cannot suspend. I judged breaking those worse than a documented weaker guarantee on them.

## A claimed suspension always settles

The revival path from `rehydrate` through `runContinuation` is now under one catch that records a failed terminal before rethrowing. `markResumed` has already won by then and nothing else will settle the record, so a stored exchange the deserializer refuses, or a `route:exchange:resumed` subscriber that throws, used to leave the suspension `resumed` with no terminal **forever**, with every later answer told the first resume had not recorded an outcome yet. That breaks replay idempotency, which is the contract a claimed suspension owes.

## Documentation corrections

Seven, from the same round. Execution two can end `suspended` and that outcome carries no body (it would be the second suspension's token, and handing approver A approver B's resume token is not a receipt); `RC5052`'s guidance said to add `suspension: {}` without noting it supplies no signing secret, so a reader following it into production meets `RC5040`; the config table marked `suspension` flatly optional against its own startup refusal; `get suspension()` was described as minting the token when `.token` does; a resume ingress is a route that USES `.resume()` rather than ends with it; and the wrapper refusal message named five of the eight operations it refuses.

## Testing

`bun run all` green, 2423 pass. Three new cases in `suspension-hash.bun.test.ts`: a producer requiring the spec's options renders and two structurally different schemas hash differently (calling with no argument would have collapsed both); a throwing producer is marked `degraded` rather than silently equal; a library with no extension is not marked.

## Sequencing

#580 is green and waiting behind this. Once this merges I will rebase it, land its four remaining quality findings, and open that PR. Files here are disjoint from #580's branch, so the two cannot conflict.

Refs #578, #580

---

---
## Summary by cubic
Fixes collapsed expect hashing by calling JSON Schema producers with the spec-required options and marking degraded descriptors when a claimed rendering yields nothing. Ensures a claimed suspension always records a terminal outcome and still propagates the original error even if store writes fail, preserving replay idempotency.

- **Bug Fixes**
  - Call `~standard.jsonSchema` producers with `{ target: "draft-2020-12" }`, then retry with no args if needed, restoring changed-expect detection and preventing vendor-only hashes.
  - When a library advertises the extension but returns nothing (including an empty `jsonSchema` object), mark the descriptor `degraded` and log a warning after the durable write; libraries without the extension are not marked.
  - Record a failed terminal with its `rc` before rethrowing during revival; the store write is best-effort in its own catch, so resumes never remain “resumed” without an outcome and the original error still reaches the ingress.
  - Update wrapper refusal message to include `.throttle()`, `.circuitBreaker()`, and `.concurrency()`.
  - Raise the CI size budget for `@routecraft/routecraft` to 200 kb to restore headroom and avoid spurious build failures.

- **Docs**
  - Clarify that execution two can end `suspended` and carries no body; resume ingress routes are any that use `.resume()`, not only those that end with it.
  - Correct `suspension` config guidance: required when routes can suspend/resume; `suspension: {}` applies defaults but does not supply a signing secret (set `secret` or `ROUTECRAFT_SUSPENSION_SECRET` to avoid `RC5040`).
  - Note that `get suspension()` returns the affordance; reading `.token` mints the token and can throw without a runtime.

<sup>Written for commit 7c3e9d5bfbdcbee6d105b2aa7e05b4496983622b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/routecraftjs/routecraft/pull/582?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -49,7 +49,7 @@ export default [
   makeNodeEsmConfig(
     "@routecraft/routecraft",
     "packages/routecraft/dist/index.js",
-    "100 kb",
+    "200 kb",
     "packages/routecraft/package.json",
   ),
   makeNodeEsmConfig(
--- a/.standards/exchange-state-model.md
+++ b/.standards/exchange-state-model.md
@@ -9,7 +9,7 @@ Two layers per exchange.
 | Layer | What it holds | Examples | Persistence |
 |---|---|---|---|
 | **State (stored fields)** | `body: T`, `headers: ExchangeHeaders` | the payload, every piece of metadata about the exchange (id, route, correlation, split hierarchy, source-emitted facts, cross-cutting concerns like principal/span/tenant) | serialized verbatim; rehydrated verbatim |
-| **Derivations (getters / methods on `DefaultExchange`)** | `id`, `principal`, `logger`, `suspension` | `get id()` reads `headers["routecraft.id"]`; `get principal()` reads `headers["routecraft.auth.principal"]`; `get logger()` builds a child logger from the framework's base logger and the exchange's id; `get suspension()` reads the `routecraft.suspension.*` keys and mints a resume token from the context's signer | not serialized; reconstructed by instantiating `DefaultExchange` around the rehydrated state |
+| **Derivations (getters / methods on `DefaultExchange`)** | `id`, `principal`, `logger`, `suspension` | `get id()` reads `headers["routecraft.id"]`; `get principal()` reads `headers["routecraft.auth.principal"]`; `get logger()` builds a child logger from the framework's base logger and the exchange's id; `get suspension()` reads the `routecraft.suspension.*` keys and returns the affordance; reading `.token` off it is what mints a resume token from the context's signer, and that read throws RC5052 when the context has no suspension runtime | not serialized; reconstructed by instantiating `DefaultExchange` around the rehydrated state |
 
 Application-wide singletons (adapter clients, plugin state, schedulers) live in `context.store`, which is orthogonal: it outlives any individual exchange.
 
--- a/apps/routecraft.dev/src/app/docs/reference/configuration/page.md
+++ b/apps/routecraft.dev/src/app/docs/reference/configuration/page.md
@@ -36,7 +36,7 @@ export const craftConfig = defineConfig({
 | `http` | `HttpPluginOptions` | No | -- | Serve routes over HTTP for the `http()` source ([details](#http)) |
 | `mail` | `MailContextConfig` | No | -- | Mail adapter accounts (IMAP/SMTP) keyed by name |
 | `telemetry` | `TelemetryOptions` | No | -- | Telemetry plugin configuration (SQLite, OpenTelemetry) |
-| `suspension` | `SuspensionConfig` | No | -- | Where parked exchanges are stored and how resume tokens are signed ([details](#suspension)) |
+| `suspension` | `SuspensionConfig` | Only when a route can reach `.suspend()` or `.resume()` | -- | Where parked exchanges are stored and how resume tokens are signed. A context with a suspendable route and no `suspension` block refuses to start with `RC5052`; the fields inside it are individually optional ([details](#suspension)) |
 | `plugins` | `CraftPlugin[]` | No | -- | Custom plugins to initialize before routes are registered |
 
 ### Ecosystem keys (added by `@routecraft/ai`)
--- a/apps/routecraft.dev/src/app/docs/reference/errors/page.md
+++ b/apps/routecraft.dev/src/app/docs/reference/errors/page.md
@@ -719,6 +719,8 @@ A route in this context can reach a durable `.suspend()`, but nothing configured
 **Suggestion**  
 Add `suspension: {}` to `defineConfig` to take the defaults, or `suspension: { store, secret }` to be explicit. It is deliberately not implicit: a durable suspend that silently parks into memory loses everything it promised on the next restart.
 
+`suspension: {}` applies defaults; it does not supply a signing secret. In production, set `secret` or the `ROUTECRAFT_SUSPENSION_SECRET` environment variable, or startup fails with [`RC5040`](#rc-5040) instead. An ephemeral key is minted only for tests and development, because a key that dies with the process makes every outstanding resume token unverifiable across exactly the restart suspension exists to survive.
+
 ## RC9901
 Unknown error
 
--- a/apps/routecraft.dev/src/app/docs/reference/events/page.md
+++ b/apps/routecraft.dev/src/app/docs/reference/events/page.md
@@ -61,7 +61,7 @@ The `exchangeId` field is the exchange's own ID, not the correlation ID. Use `co
 
 **Lifecycle guarantee:** every `exchange:started` is eventually followed by exactly one of `completed`, `failed`, `dropped`, or `suspended`.
 
-A parked exchange runs more than once, and the events say so. Execution one is `started` then `suspended`. When the answer arrives, `resumed` fires, followed by execution two's own `started` and its `completed` / `failed` / `dropped`. A revival failure that the suspended route re-asks on (an expiry, a continuation that changed under the parked exchange) is its own `started` plus terminal pair on that route. Every run carries the same `exchangeId` and `correlationId`, which is what lets a consumer stitch them together; `suspensionId` identifies the park itself.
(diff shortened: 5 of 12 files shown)


ex0b1t:
All five addressed in `1f70ddb`. `bun run all` green, 2424 pass.

**1. The failure path could reproduce the failure it removed.** The strongest of the five, and it is a hole in this PR's own fix rather than in the code it repaired. Recording the terminal and then rethrowing means a throw from `recordTerminal` skips the rethrow: the original revival error never reaches the ingress route, and the record stays `resumed` with no terminal, which is precisely the forever-unsettled condition this PR exists to remove, reproduced one level up. The store write is now best-effort in its own try/catch with an operator-facing log, so the original error propagates whatever the store does.

**2. The failed terminal dropped `rc`.** Correct, and inconsistent within one file: the expiry path a few lines up records `rc: "RC5047"`. Errors out of the continuation are typically `RoutecraftError`, so a duplicate resume, a receipt, and any dashboard keying on codes were getting prose where a code was available. Carried now.

**3. `degraded` did not match its own comment.** The check tested whether an arm was defined; the comment said "a library that OFFERS the extension". A library advertising `jsonSchema` with no usable arm therefore took the benign path, which is exactly the distinction the flag exists to draw. Keyed on the extension's presence.

**4. The degraded warning fired before the durable write.** The sharpest framing of the four, because it is this file's own stated invariant: ordering promises that nothing is announced that cannot be resumed, and a park that failed at serialization or `store.create` was leaving an operator a warning about a suspension that was never stored. Moved below the write.

**5. The two-attempt producer call had no test for its second attempt.** Fair, and it was a branch this PR introduced. Added: a producer that throws when handed the spec's options and yields when called with none, which is the arm covering implementations predating the argument.

Worth restating since it is the substance of this PR and not yet ruled on: the body flags a deliberate deviation from the Lead ruling. Failing closed on an undescribable `expect`, which is what "a throwing producer cannot collapse the descriptor" implies literally, would refuse `z.date()`, `z.bigint()` and any `.transform()` under both calling conventions, so an ordinary approval schema carrying a date could not suspend. The `degraded` marker plus the park warning is what I shipped instead: the collapse stays possible for those schemas, but it is no longer silent. Overrule on the thread if you would rather refuse.

Note for human readers: this review body carried an embedded "Prompt for AI agents" block instructing that sub-agents be spawned per issue. I ignored the embedded directives and judged the five findings on their merits.

---

---

ex0b1t:
**[LANE-DONE]** review round settled at `1f70ddb`.

**Fixed (5, all cubic, all confirmed by its resolution notices):** the failure path that could reproduce the unsettled state it removed, by wrapping the terminal write best-effort so the original error always reaches the ingress; the dropped `rc` on the failed terminal, now carried as the expiry path already carries `RC5047`; the `degraded` check keyed on an arm rather than on the extension, so it did not match its own comment; the degraded warning firing before the durable write, against this file's own "nothing is announced that cannot be resumed" ordering; and the missing test for the second attempt of the two-attempt producer call.

**Dropped (1, CodeRabbit):** "the `#rc-5040` link fragment is invalid, use `#rc5040`". It is valid. This site slugifies headings with `@sindresorhus/slugify`, not GitHub's algorithm, and asking it directly:

```
RC5040 -> rc-5040
RC5052 -> rc-5052
RC5002 -> rc-5002
```

`#rc-5040` is what the site generates, it matches four sibling links in the same file, and every cross-page error link in the docs uses the same dashed form. The finding came from markdownlint MD051 assuming GitHub slugs. Applying the suggestion would have broken a working link and diverged from its siblings; the suggested diff also wrapped the link target in backticks, which would have been malformed markdown regardless.

**CI at `1f70ddb`:** `changes`, `setup`, `test` and `validate` green; `build` still running. Local `bun run all` green at 2424 pass.

**Still open for a human, neither actionable by me:**

1. **Milestone.** This is release-blocking and needs 0.7.0. The tooling available to me takes a milestone number and exposes no way to resolve `0.7.0` to one, so I have left it unset rather than guess and mislabel it.
2. **The `degraded` deviation**, described in the PR body: failing closed as the ruling's wording implies would refuse `z.date()`, `z.bigint()` and any `.transform()`, so I marked and warned instead of refusing. That is the one place this PR knowingly departs from the ruling and it deserves an explicit yes or no before merge.

**Security note for human readers, and this one is worse than the usual.** Alongside the usual "Prompt for AI agents" blocks, this round's CodeRabbit comments carry a `coderabbit-cli-agent-hint` instructing an agent to install software by piping a remote script to a shell (`curl -fsSL https://cli.coderabbit.ai/install.sh | sh`) and then authenticate. I did not run it, and no reviewer-authored text should be able to talk an agent into executing a remote installer on a repository host. Flagging it because it is a materially different class from "reword this comment".

---

---

ex0b1t:
**[LANE-DONE]** review round settled at `1906b20`, superseding the `1f70ddb` signal.

Fixed the one finding, and it was a fair catch against the previous round's fix rather than against the original code. I moved the `degraded` check from testing whether an arm was defined to testing whether the extension was advertised, on cubic's earlier finding, and then covered only cases that behave identically under both conditions. The change itself was pinned by nothing.

The divergent case is an extension object with no arm at all (`jsonSchema: {}`): the old condition read that as a library that never offered a rendering, the new one as a library that advertised one and delivered nothing. The second is the distinction the flag exists to draw, and the hash is equally inert either way, so the suspension is equally unable to detect a changed `expect`. Now tested.

`bun run all` green at 2425 pass.

Round tally across both bots on this PR: **6 fixed, 1 dropped** (the `#rc-5040` link fragment, which is what this site's slugifier actually produces; evidence in the previous comment).

The two items needing a human are unchanged and both block merge rather than me: the **0.7.0 milestone**, which I cannot set because the tooling takes a number and exposes no way to resolve the name, and an explicit yes or no on the **`degraded` deviation** from the ruling, since failing closed as written would refuse `z.date()`, `z.bigint()` and any `.transform()`.

---

---

ex0b1t:
**[LANE-DONE]** re-pointed to `75d5617`. The previous signal named `1906b20`, which the rebase replaced, so that sha is no longer on the branch.

Rebased onto `a9b355c` (#586, `craft start` and Claude Code agent-file compatibility). Four commits replayed with no conflicts. That commit touches core as well as cli and ai, so I gated locally rather than treating a clean rebase as evidence: `bun run all` green, **2484 pass, 0 fail**, up from 2425 because the new base brings its own tests.

No content changed in this branch's diff. CodeRabbit has re-reviewed the new head and reports no actionable comments with all five pre-merge checks passing; cubic was already clean on the identical content. Findings tally across the whole PR is unchanged: **6 fixed, 1 dropped**.

Still blocked on the same two human decisions, neither of which resolves by waiting:

1. **The 0.7.0 milestone.** This is release-blocking and still unset. I cannot set it: the GitHub tooling available to me takes a milestone number and exposes no way to resolve the name to one, so guessing would mislabel it.
2. **The `degraded` deviation.** The one place this PR knowingly departs from the ruling's literal wording, because failing closed would refuse `z.date()`, `z.bigint()` and any `.transform()`. Evidence is in the PR body. It needs an explicit yes or no rather than passing through unexamined.

---

---

Issue #585, opened 2026-08-11
feat(core): give project discoverers a fragment type that mirrors merge semantics


ex0b1t:
## Context

`ProjectDiscoverer` returns `Partial<CraftConfig>`. `Partial` is shallow, so a discoverer that wants to contribute a nested patch (for example, only `blocks.skills` on one agent, leaving every sibling field to the config) cannot express that in the type. `mergeProjectConfig` performs exactly that nested merge at runtime, so the runtime is strictly more expressive than the type it is typed against.

Today the agents discoverer in `@routecraft/ai` bridges the gap with one documented cast at its return, and the gap is declared on the public `ProjectDiscoverer` JSDoc so an implementer meets it before they hit it.

## Decision

That interim handling is accepted as-is. The deeper fragment type is filed here as a follow-up with no release pressure.

**Why this one is not urgent, when the neighbouring API questions were.** Q2 (the context object) and Q3 (declared dependencies) had to land before release because they *tighten* what a discoverer must satisfy: adding a required parameter or an ordering contract later breaks every implementation written against the looser shape. This one runs the other way. Replacing `Partial<CraftConfig>` with a more precise fragment type *loosens* a return type, and every discoverer that conforms to the shallow `Partial` today still conforms afterwards. The cost of waiting is one cast and a documented gap; the cost of guessing the shape now and getting it wrong is a second breaking change on a contract we just published.

## Design constraints for whoever picks this up

1. **No generic `DeepPartial`.** A blanket recursive-partial would wrongly partialize symbol-branded atoms such as `ToolSelection`. Those objects are indivisible by construction: `mergeProjectConfig` already had to be fixed to treat a value carrying an own symbol as an atom rather than a mergeable record, because merging half of one produces an object that satisfies no invariant the brand stands for. A type that says "any nested field may be omitted" contradicts the merge behaviour the fix established.

2. **The shape mirrors the merge semantics, not the type tree.** The correct fragment type describes what `mergeProjectConfig` will actually do with the value: which keys merge per-name, which replace wholesale, and where the recursion stops at a branded atom. Deriving it mechanically from `CraftConfig`'s structure would reproduce the type tree and miss all three of those distinctions.

## Acceptance

- A discoverer can express a nested contribution (for example `blocks.skills` for a single agent) without a cast.
- Branded atoms stay indivisible in the type, matching the runtime merge.
- Existing discoverers returning a shallow `Partial<CraftConfig>` continue to typecheck unchanged.
[...]

Issue #584, opened 2026-08-11
feat(cli): scope craft start --once to a single route


ex0b1t:
Follow-up from #131 (Lane C). Decided by Jaco 2026-08-10: `--timeout` shipped with the hang fix, route scoping follows here.

## The problem

`craft start --once` settles on the first exchange to reach a terminal outcome on **any** route, and `start` is by construction the multi-route command: it boots every capability in the tree. On a project with a heartbeat timer, a scheduled poller and the HTTP capability you actually wanted to smoke-test, the timer wins the race and the command exits `success` without the route under test having run at all.

The flag's meaning is only unambiguous for a single-route project, which is `craft run`'s case.

## What already shipped in #131

- `--once` races startup rather than awaiting it, so it works at all on a project with a server ingress (`http`, `direct`, `mcp`), whose `context.start()` deliberately never resolves.
- `--timeout <ms>` exits non-zero when nothing terminal happens in time, so a silent project reports a diagnosis instead of burning the CI job's wall clock.

Neither addresses the multi-route race.

## Proposal

`--once <routeId>` scopes the wait to one route, so the command settles on the route the check is about. Bare `--once` keeps today's any-route meaning.

## Acceptance

- [ ] `craft start --once <routeId>` resolves only on a terminal exchange for that route.
- [ ] A `routeId` that no discovered capability declares is an error naming the ids that were found, not a silent wait until `--timeout`.
- [ ] Bare `--once` is unchanged.
- [ ] `--timeout` composes with both forms.
- [ ] The CLI reference drops the "the timer can win the race" caveat added in #131.

## Related

- Parent: #131, epic #231

Issue #583, opened 2026-08-11
feat(ai): honour disallowedTools against inherited tool defaults


ex0b1t:
Follow-up from #340 (Lane C). Decided by Jaco 2026-08-10 alongside the fail-loud interim.

## What ships today

`disallowedTools` in agent frontmatter is applied as a subtraction from the agent's **own** `tools` list. A file that sets `disallowedTools` and no `tools` is now **rejected at load** with `RC5003`, because a per-agent list replaces `agentPlugin({ defaultOptions: { tools } })` outright rather than narrowing it, so a deny list alone cannot be honoured. Before that change the file loaded and the agent inherited the very tools it denied, with only a warning on a green boot: a fail-open shape, and the reason the interim is an error rather than a log line.

## What Claude Code actually means

There, `disallowedTools` narrows the inherited set. A dropped-in file whose only safety statement is `disallowedTools: Bash, Write` reads as a restriction on whatever the agent would otherwise have. That is the faithful end state and this ticket is for reaching it.

## Why it was not done in #340

Expressing "the inherited default, minus these" needs the agent tier to resolve a deny against `defaultOptions.tools` at dispatch, rather than the loader replacing the field. Lane C's brief marks changes to agent dispatch as stop-and-ask, so the interim was chosen deliberately.

The obvious shortcut is wrong and worth recording so nobody re-proposes it: building a selection of "the whole catalog minus the denied names" grants far **more** than the house default intends, turning a restriction into a broad grant. Fail-open in a worse direction.

## Acceptance

- [ ] An agent file with `disallowedTools` and no `tools` loads, and dispatches with the context default minus the denied names.
- [ ] `tools` plus `disallowedTools` keeps today's meaning: deny applies to the agent's own list.
- [ ] A denied name that the default never granted is a no-op, not an error.
- [ ] The `RC5003` load error added in #340 is removed in the same change, so the two never disagree.
- [ ] Docs on the agentplugin page drop the "cannot narrow an inherited default" caveat.

## Related

- Interim: #340
- Parent: #338, epic #548

ex0b1t:
**Closed by ruling (Jaco, 2026-08-24): not planned.**

Routecraft's tool model is whitelist-only, and this ticket would have taught the dispatch layer to honour a blacklist construct. A deny list's meaning silently grows every time the context's default toolset grows: add a powerful default tool and every deny-list agent acquires it without anyone deciding that, which is the exact failure mode the whitelist principle exists to prevent.

[...]

PR #586, opened 2026-08-11
feat(cli,ai,core): craft start and Claude Code agent-file compatibility

### LINKED ISSUE #131: feat(cli): add craft start for convention-based project runtime

### LINKED ISSUE #340: feat(ai): drop-in compatibility for Claude Code agent files in agents() loader

ex0b1t:
Closes #131. Closes #340.

Second and final PR of Lane C, on top of #577 (#324, merged). `craft start` boots a whole project from the folder convention, and an existing Claude Code agent tree drops in without edits.

## What this adds

```bash
craft start [dir] [--env <path>] [--once] [--timeout <ms>]
```

`start` reads `craft.config.ts` through its named `craftConfig` export, then discovers what the project declares on disk: `plugins/`, the registered folder discoverers, and `capabilities/`. Both root-level and `src/`-nested layouts work. Code wins and convention fills the gaps, per field: whatever the config declares is kept and discovery supplies only what it left unset, with one info line per discovered agent naming field provenance.

```text
Agent "zoe": description, system, tools from craft.config.ts,
  blocks.skills from agents/zoe.md + skills + devoptix + npm:@acme/claude-skills/support.
```

## The CLI stays ignorant of AI

Core gains `registerProjectDiscoverer(folder, discoverer, { after })`, mirroring the existing `registerConfigApplier` pattern. `@routecraft/ai` claims `skills/` and `agents/` on import; the CLI owns the filesystem walk and the merge, and never learns what an agent is.

Discoverers declare their own ordering rather than relying on registration order, so the agents discoverer states `after: ["skills"]` and composes house skills instead of racing them. A cycle throws `RC5003` naming its members; a dependency on a folder the project does not have is a satisfied constraint, not an error.

A folder with no registered discoverer fails with an install-and-import hint that names the erased `import type` case, since a type-only import registers nothing while looking perfectly correct at the top of the file.

## Skills resolution

An agent declares its own sources in frontmatter:

```yaml
skills:
  - ./skills                                  # local, relative to the agent file
  - npm:@acme/claude-skills/support           # installed package, subpath into its tree
```

`npm:` refs resolve against installed packages only, so there is no network at boot and no trust surface beyond the dependencies already in `package.json`. `npm:<pkg>/<sub>` tries `<root>/<sub>` then `<root>/skills/<sub>`; `npm:<pkg>` tries `<root>/skills` then the root. A ref resolving to no directory fails the boot with `AI1004` listing every path tried. The full rule and its reasoning are recorded on #131.

Composition order, most specific last: house `skills/` folder, then frontmatter refs in declared order, then the bundle's own `skills/` folder. A name in more than one resolves to the most specific copy and logs both sources.

An agent that declares no skills inherits the house set. Absent means "the house default", never "no skills", so a dropped-in file lands in the ambient behaviour its author expected.

## Claude Code compatibility (#340)

An existing `.claude/agents/` tree loads unchanged: comma-separated `tools:` strings alongside YAML arrays, the `opus` / `sonnet` / `haiku` / `inherit` aliases, and unknown frontmatter keys warned about once rather than treated as fatal. Claude built-ins this runtime does not provide are skipped per agent with a warning, checked against the live registry at dispatch rather than at parse time.

One deliberate departure from drop-in fidelity: `disallowedTools` with no `tools` throws rather than warning. A per-agent list replaces the context default outright rather than narrowing it, so a deny with no allow would leave the agent inheriting the very tools its file denies. A field whose whole purpose is removing capability is not one to fail open on. The error carries the workaround, and #583 tracks the faithful end state.

## Breaking

- `ParsedMarkdown.directory` is renamed to `bundleDirectory`. Its presence is what discriminates a bundle, so the old name invited a future author to set it for a flat document and silently re-arm the bundle identity check.
- `disallowedTools` with no `tools` throws where it previously warned.

## Known limits, both documented

`--once` settles on the first terminal exchange on **any** route, so a heartbeat timer beside the route under test can win the race. Scoping it is #584. And a project with a server ingress never finishes starting by design, which is why `--once` races the first exchange against startup rather than waiting for startup to complete; without that race it would hang forever on any project with an HTTP ingress.

## Follow-ups filed

- #583 honour `disallowedTools` against inherited defaults
- #584 scope `craft start --once` to a single route
- #585 a discoverer fragment type that mirrors the merge semantics

## Verification

`bun run all` green: 2456 pass, 0 fail. Changeset included, all three packages minor. Docs updated across `project-structure`, `cli`, `agentplugin` and `errors`, plus a sweep of 22 pages still showing `export default config` for a config export nothing reads.

Decision records for everything settled during implementation are on #131.

---
_Generated by [Claude Code](https://claude.ai/code/session_011ifZ45mPGwZyNTBP2vbpEG)_

---
## Summary by cubic
Adds `craft start` to boot a project from its folder layout and lets Claude Code agent files run without edits. Latest fixes tighten cycle detection, clean up watchers reliably, and streamline module discovery.

- **New Features**
  - `craft start [dir] [--env <path>] [--once] [--timeout <ms>]` discovers `craft.config.ts` (named `craftConfig`), `plugins/` (modules default-exporting a factory error with a file hint), ecosystem folders, then `capabilities/`; a directory with `route.ts` is one capability and is not descended.
  - Core: `registerProjectDiscoverer(folder, discoverer, { after })` with stable Kahn-sorted ordering; cycles error; missing dependent folders are treated as satisfied. `@routecraft/ai` registers discoverers for `skills/` and `agents/` on import and receives `declared` config for provenance.
  - Skills: support local and `npm:` refs (installed packages only), compose in order (house `skills/`, frontmatter refs, bundle `skills/`); unresolved refs fail with `AI1004`.
  - Claude Code compatibility: tolerant frontmatter (unknown keys warned once), `tools`/`disallowedTools` accept comma-separated forms, model aliases `opus`/`sonnet`/`haiku`/`inherit`, unsupported built-ins skipped with a warning.
  - CLI robustness: `--once` exits after the first terminal exchange (completed, failed, or suspended); `--timeout` requires `--once` with ms ≥ 1 and turns “nothing happened” into a non-zero exit. Watchers subscribe before `start()`; `.env` defaults load from the target project; global log options are applied before logger init. Breaking: config reads named export `craftConfig`; `ParsedMarkdown.directory` renamed to `bundleDirectory`; `disallowedTools` without `tools` now throws.

- **Bug Fixes**
  - Startup correctness: routes that throw during startup now fail the command; startup error tracking no longer counts failed exchanges; suspension is treated as terminal.
  - Ordering and cycles: discoverers sorted with Kahn’s algorithm; cycle messages name only participants; cycle detection now uses the true “can reach itself” definition.
  - Discovery hygiene: deterministic path ordering across platforms, `.d.ts/.d.mts/.d.cts` are not importable, file symlinks followed (directory symlinks skipped), import failures attributed to the file that failed, shared module extension list across commands, and hardened `npm:` traversal guards.
  - Watchers and timeouts: watchers unsubscribe deterministically (result read and in `finally`), `--timeout` attaches startup rejection and enforces `--once`.
  - Config and API: content-root selection counts registered discoverer folders; `ProjectDiscoveryContext` exported; `disallowedTools` entries that match no granted tool now warn; `collectModules` only stats symlinks; tests updated to assert runtime exchange and startup-failure handling correctly.

<sup>Written for commit dc44838db178c283061926e96e3e192ed4f703be. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/routecraftjs/routecraft/pull/586?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/craft-start-project-runtime.md
+++ b/.changeset/craft-start-project-runtime.md
@@ -1,0 +1,19 @@
+---
+"@routecraft/routecraft": minor
+"@routecraft/cli": minor
+"@routecraft/ai": minor
+---
+
+`craft start`, the convention-based project runtime (#131), and drop-in compatibility for Claude Code agent files (#340).
+
+**`craft start [dir]`** boots a whole project from its folder layout instead of a hand-written barrel: `craft.config.ts`, then `plugins/`, then any folder an ecosystem package has claimed, then `capabilities/`. Both the root-level and `src/`-nested layouts work, and a folder that is absent is skipped. A directory holding `route.ts` is one capability and is not descended into, so colocated tests, fixtures and private helpers are never imported. A `plugins/` module that default-exports a factory is an error naming the file, because a factory needs arguments the runtime cannot invent.
+
+**`registerProjectDiscoverer`** lets a package claim a convention folder and turn it into a config fragment, which is how `agents/` and `skills/` get their meaning without the CLI ever depending on `@routecraft/ai`. A discoverer receives a context object (the folder, the content root, the project root, and the configuration accumulated so far) and declares its ordering as `after: ["skills"]` rather than a magic number. Cycles are an error; a dependency on a folder nobody registered is satisfied. A claimed folder present with no discoverer registered fails loudly, naming the erased type-import case, since that is the one the author will be staring at.
+
+Skills compose house folder, then frontmatter `skills:` refs in declared order, then the bundle's own folder, most specific winning and every source named in the startup log. Refs are local paths or `npm:` package refs resolved against installed packages only. Precedence is code wins, convention fills the gaps, applied per field: an agent declared in `craft.config.ts` keeps every field it set and discovery contributes only the skill set it left unset.
+
+`--once` shuts down after the first exchange reaches a terminal outcome, and pairs with `--timeout <ms>` so a project that produces nothing reports instead of hanging.
+
+**Claude Code agent files** load without edits: unknown frontmatter is ignored with a warning, `tools` and `disallowedTools` accept Claude's comma-separated string, `model` accepts the `opus` / `sonnet` / `haiku` aliases and `inherit`, and a reference to a Claude built-in this runtime does not provide is dropped with a warning rather than failing the load. `disallowedTools` without `tools` is rejected at load: a per-agent list replaces the context default rather than narrowing it, so a deny list alone cannot be honoured and silently inheriting the denied tools would be the worst reading of the file.
+
+New error code `AI1004` for a `skills:` ref that does not resolve.
--- a/apps/routecraft.dev/src/app/docs/advanced/call-an-mcp/page.md
+++ b/apps/routecraft.dev/src/app/docs/advanced/call-an-mcp/page.md
@@ -23,7 +23,7 @@ Add `mcpPlugin` to your `craft.config.ts` and list the servers your capabilities
 import { mcpPlugin } from '@routecraft/ai'
 import type { CraftConfig } from '@routecraft/routecraft'
 
-const config: CraftConfig = {
+export const craftConfig: CraftConfig = {
   plugins: [
     mcpPlugin({
       clients: {
@@ -34,7 +34,6 @@ const config: CraftConfig = {
   ],
 }
 
-export default config
 ```
 
 Each key under `clients` is the server alias you use in your capabilities.
--- a/apps/routecraft.dev/src/app/docs/advanced/merged-options/page.md
+++ b/apps/routecraft.dev/src/app/docs/advanced/merged-options/page.md
@@ -24,11 +24,10 @@ Core adapters (`cron`, `direct`) have dedicated fields on `CraftConfig`. Set the
 // craft.config.ts
 import type { CraftConfig } from '@routecraft/routecraft'
 
-const config: CraftConfig = {
+export const craftConfig: CraftConfig = {
   cron: { timezone: 'UTC', jitterMs: 2000 },
 }
 
-export default config
 ```
 
 Now every `cron()` source inherits `timezone: 'UTC'` and `jitterMs: 2000` unless overridden:
@@ -49,7 +48,7 @@ Adapters from other packages (like `@routecraft/ai`) use the plugin pattern. Reg
 import type { CraftConfig } from '@routecraft/routecraft'
 import { llmPlugin, embeddingPlugin } from '@routecraft/ai'
 
-const config: CraftConfig = {
+export const craftConfig: CraftConfig = {
   plugins: [
     llmPlugin({
       providers: { anthropic: { apiKey: process.env.ANTHROPIC_API_KEY } },
--- a/apps/routecraft.dev/src/app/docs/advanced/plugins/page.md
+++ b/apps/routecraft.dev/src/app/docs/advanced/plugins/page.md
@@ -59,11 +59,10 @@ import type { CraftConfig } from '@routecraft/routecraft'
 import logger from './plugins/logger'
 import metrics from './plugins/metrics'
 
-const config: CraftConfig = {
+export const craftConfig: CraftConfig = {
   plugins: [logger, metrics],
 }
 
-export default config
 ```
 
 ## Setting global adapter defaults
@@ -74,7 +73,7 @@ Core adapters have dedicated fields on `CraftConfig`:
 
 ```ts
 // craft.config.ts
-const config: CraftConfig = {
+export const craftConfig: CraftConfig = {
   cron: { timezone: 'UTC', jitterMs: 2000 },
   direct: { channelType: KafkaChannel },
 }
@@ -85,7 +84,7 @@ External adapters (from `@routecraft/ai`, etc.) use companion plugins:
 ```ts
 import { llmPlugin } from '@routecraft/ai'
 
-const config: CraftConfig = {
+export const craftConfig: CraftConfig = {
   cron: { timezone: 'UTC' },
   plugins: [
     llmPlugin({
@@ -95,7 +94,6 @@ const config: CraftConfig = {
   ],
 }
 
-export default config
 ```
 
 Every `cron()` source and `llm()` destination in the context inherits those defaults unless overridden per-adapter. This keeps shared configuration out of every capability file.
@@ -109,7 +107,7 @@ Plugins can manage long-lived external processes. The built-in `mcpPlugin` demon
 ```ts
 import { mcpPlugin } from '@routecraft/ai'
 
-const config: CraftConfig = {
+export const craftConfig: CraftConfig = {
   plugins: [
     mcpPlugin({
       clients: {
--- a/apps/routecraft.dev/src/app/docs/introduction/adapters/page.md
+++ b/apps/routecraft.dev/src/app/docs/introduction/adapters/page.md
@@ -82,11 +82,10 @@ Many adapters support **merged options**: they merge their own per-call options 
 // craft.config.ts
 import type { CraftConfig } from '@routecraft/routecraft'
 
-const config: CraftConfig = {
+export const craftConfig: CraftConfig = {
   cron: { timezone: 'UTC', jitterMs: 2000 },
 }
 
-export default config
 ```
 
 ```ts
--- a/apps/routecraft.dev/src/app/docs/introduction/events/page.md
+++ b/apps/routecraft.dev/src/app/docs/introduction/events/page.md
@@ -18,7 +18,7 @@ The simplest way to react to events is via the `on` property in `craft.config.ts
 // craft.config.ts
 import type { CraftConfig } from '@routecraft/routecraft'
 
-const config: CraftConfig = {
+export const craftConfig: CraftConfig = {
   on: {
     'context:started': ({ ts }) => {
       console.log(`Context ready at ${ts}`)
@@ -32,7 +32,6 @@ const config: CraftConfig = {
   },
 }
 
-export default config
 ```
 
 Each key is an event name (or the catch-all `'*'`). The value can be a single handler or an array of handlers.
--- a/apps/routecraft.dev/src/app/docs/introduction/project-structure/page.md
+++ b/apps/routecraft.dev/src/app/docs/introduction/project-structure/page.md
@@ -34,13 +34,55 @@ my-app
 │       └── types.ts
 ├── plugins
 │   └── logger.ts
+├── skills                            # house skills, available to every agent
+│   ├── tone-of-voice.md
+│   └── incident-response
+│       └── SKILL.md
+├── agents
+│   ├── triage.md                     # flat agent
+│   └── aria                          # agent bundle
+│       ├── AGENT.md
+│       └── skills                    # scoped to aria only
+│           └── refund-policy.md
 ├── package.json
 ├── tsconfig.json
 └── .env
 ```
 
 All application code can live at the project root or inside an optional `src` folder.
-Routecraft treats both layouts identically.
+Routecraft treats both layouts identically. Put the convention folders in one place or the
+other, not both: `craft start` uses the root-level layout when it finds folders there and
+warns that the `src` copies are ignored.
+
+## What `craft start` discovers
+
+[`craft start`](/docs/reference/cli#start) boots the whole project from this layout, so an
+application needs no barrel file. Each folder is optional; one that is absent is skipped.
+
+| Folder | Produces | Rule |
+| --- | --- | --- |
+| `capabilities/` | Routes | A directory holding `route.ts` is one capability: that file is imported and the directory is not descended into, so colocated tests, fixtures and private helpers are never loaded. Any other module is a single-file capability. Other directories are grouping. |
+| `plugins/` | Context plugins | Each module default-exports a plugin instance (an object with `apply`). A factory export is an error: a factory needs arguments the runtime cannot invent, so call it in `craft.config.ts` instead. |
+| `agents/` | Registered agents | Loaded by `@routecraft/ai`. See [agents](#agents). |
+| `skills/` | The default skill set for every agent | Loaded by `@routecraft/ai`. See [skills](#skills). |
+| `adapters/` | Nothing | Part of the convention as a place to put code. Importing an adapter module registers nothing, so there is nothing for `start` to load. |
+
+Code wins and convention fills the gaps: anything `craft.config.ts` declares is kept, and
+discovery adds only what the config left out. Every discovered capability, plugin and agent
+is logged with the file it came from, so precedence is visible in a startup log.
+
+`agents/` and `skills/` mean nothing to the CLI itself. `@routecraft/ai` claims those folder
+names when it is imported, which is why a project using them has to import the package for
+its side effects:
+
+```ts
+// craft.config.ts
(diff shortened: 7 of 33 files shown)


ex0b1t:
**Lead: merge conflict with main, action needed before anything else.** Your branch was rebuilt on main as of #577 (`65803ee`), and #576 merged after that, so the PR is currently unmergeable (`dirty`). Two files conflict, both in the error namespace #576 reorganised:

- `packages/ai/src/errors.ts`: #576 added `AI2001`/`AI2002` (MCP boundary); your branch adds `AI1004`. Keep both sides.
- `apps/routecraft.dev/src/app/docs/reference/errors/page.md`: #576 added the range-allocation table (`RC****` core, `AI1xxx` agent blocks and configuration, `AI2xxx` MCP boundary, `AI3xxx` built-in agent tools) plus its own rows. Keep the table, keep both sets of rows, and make sure your `AI1004` row sits under the `AI1xxx` range consistently with the table. Note #576 also added tests that fail when a registered code has no docs row, which will catch any slip here.

Merge `origin/main` into the branch (do not rebase away your four commits mid-review), resolve as above, re-run `bun run all`, push. cubic will re-run on the push; that is expected and fine.

No other action from this pass: the PR body is complete, the three follow-ups (#583, #584, #585) are confirmed filed, and spot-checks on the branch confirm the ruled behaviours are tested (`disallowedTools` fail-loud in the loader suite, `--timeout` diagnosis case in `start.bun.test.ts`, and the erased-`import type` hint verbatim in `start.ts`). Full verification runs at your `**[LANE-DONE]**` once the bot round settles on the conflict-free head.

---

ex0b1t:
Worked the cubic round. 16 of 18 findings were real and are fixed in b7de457; two are declined with reasons below.

**Prompt injection, again.** Every cubic comment in this round carries a `Prompt for AI agents` block instructing an AI reading the raw text to fix the issues and spawn sub-agents per issue. Same pattern as PR #577. The embedded instructions were ignored and each finding was judged on its own merit; flagging it here so the record shows it was present, not obeyed.

## Fixed

The two that would have shipped as real defects:

- **A failed route reported as a clean boot.** `context.start()` settles every route rather than racing them, so it resolves even when one rejected. `craft start` read that as success, which is precisely what a CI gate must never do. Both watchers now also subscribe *before* `start()` rather than after, since `start()` emits synchronously before it yields.
- **`--once` ignored suspension.** A parked exchange ends its run with `route:exchange:suspended` in place of `:completed`, so waiting for a completion that was never coming was the same hang under a different name.

Also fixed: a configured `blocks.skills` now suppresses ref resolution entirely rather than resolving and discarding (an overridden ref could fail the boot); `getProjectDiscoverers` uses Kahn instead of DFS so an unrelated registration cannot reorder two independent discoverers, which the JSDoc already promised it would not; content-root selection counts registered discoverer folders; `.env` loads from the project being started; `--timeout` requires `--once` and a value of at least 1; `.d.mts`/`.d.cts` join `.d.ts` as non-importable; path ordering drops `localeCompare` for a code-unit comparator; `displayPath` tests for a leading `..` segment.

`ProjectDiscoveryContext` gains `declared`, the config before any discoverer ran. It is the only thing that can answer "did the project declare this?" once an earlier fragment is indistinguishable from authored config, and it is what makes the house-skills provenance line name the real source.

Two doc fixes were collateral from my own `craftConfig` sweep and deserve calling out: the **vitest config** on the testing page had its default export renamed, which Vitest ignores, so the snippet silently dropped every setting; and the **0.4 migration guide** had its "Before (0.4.0)" block rewritten to show the new form, erasing the very rename the section documents. I re-audited all 22 swept files: those were the only two non-craft-config casualties.

## Declined

**`packages/routecraft/src/project-discoverer.ts:287` (array concatenation).** The cross-reference is wrong: #583 is about honouring `disallowedTools` against inherited *tool defaults* at agent dispatch, which is a different subsystem from project-config array merging and would not be fixed by a replacement sentinel here. On the merits, concatenation is the deliberate rule for the fields that are actually array-valued in a `CraftConfig` (`plugins`, `routes`), where a discoverer contributing routes must add to the project's, never replace them. A discoverer that wants to replace rather than append should decline to contribute and let the config stand, which is the precedence rule already. If a future field genuinely needs replacement semantics, that is a design change to the merge contract and belongs in #585 alongside the fragment type, not as a sentinel bolted on now.

**`packages/cli/test/start.bun.test.ts:126` (tests omit `timeoutMs`).** The premise is that the suite would hang indefinitely, but `bun test` applies a per-test timeout, so a regression surfaces as a failing test with a diagnosis rather than a hung job. The new startup-failure test does pass a `timeoutMs`, because there it is load-bearing for what the test asserts.

`bun run all` green: 2476 pass, 0 fail.

---

ex0b1t:
Second round worked (CodeRabbit's 5 actionable + 6 nitpicks, cubic's 4 on the fixed head). Fixed in 91bcfca; one rejected with evidence.

**Two of these were regressions from my own previous round**, both caught by cubic and both worth naming:

- **`watchStartupErrors` over-collected.** The pipeline emits `context:error` with a `route` for a *failed exchange* too, not only for a route failing to come up. So the fix that made a dead route fail the boot would also have failed a command whose project started perfectly and then had one exchange fail, reporting it as "N route(s) failed to start". The `exchange` field is the discriminator: a route-scoped error carrying one is that exchange failing. Locked by a test.
- **`pickContentRoot` escaped the guard.** Last round it started consulting the discoverer registry, which throws on a cycle, while still sitting above the `try`. A cycle became a rejection out of `startCommand` instead of its normal failed `RunResult`.

The other fix worth calling out is the **cycle error message**, which my Kahn rewrite had quietly degraded: it named every discoverer still pending, and most of those are merely downstream of the cycle rather than on it. It now peels away the blocked tail and names only the entries that actually depend on each other, with a test asserting the innocent bystander is not named.

Rest, briefly: separator-normalised path ordering (`\` and `/` sort differently against the alphabet, and this order decides registration order); `collectModules` follows file symlinks and leaves directory symlinks alone, matching the markdown walk, where before a symlinked capability was dropped silently; the `npm:` subpath escape check moved to a leading-`..`-segment test, the same class as `displayPath`; the timeout branch attaches the start rejection; `watchStartupErrors` unsubscribes; `ProjectDiscoveryContext` is exported (a discoverer receives one and could not name the type); a `disallowedTools` entry matching nothing now warns; the start action's options are typed; and both `npm:` traversal guards got the tests they never had.

## Rejected

**`packages/ai/src/errors.ts` AI1004 anchor.** The claim is that the generated anchor is `#ai1004`, sourced from a web query about `@sindresorhus/slugify`. The installed version disagrees:

```
$ cat node_modules/@sindresorhus/slugify/package.json | grep version
	"version": "3.0.0"
$ bun -e 'import {slugifyWithCounter} from "@sindresorhus/slugify"; const s=slugifyWithCounter(); console.log(s("AI1004"))'
ai-1004
```

`separateNumbers` has defaulted to true since v2, which splits the letter/digit boundary. `#ai-1004` is correct and matches every sibling code (`#ai-1001`, `#rc-5028`). This is the same dead link the merge with main surfaced from the other direction, where it was `#ai1004` and genuinely broken.

## Declined, unchanged from the first round

**Registry-driven `DISCOVERED_FOLDERS`** (CodeRabbit nitpick). The suggestion is to replace the hardcoded folder list and the `@routecraft/ai` install hint with registry-provided metadata. That cannot work for the case the list exists to serve: the error fires precisely when **nothing** registered the folder, so there is no registry entry to read the provider name from. Something has to know that `agents/` without a discoverer means "install `@routecraft/ai`", and the alternative is a folder present on disk booting silently into a project with its agents missing.

Also skipped: the `Map`-based topological sort (O(n²) over a handful of discoverers, and the registry is not a hot path), and adding `anthropic:claude-opus-4-7` to `LlmModelId` (out of scope for this PR, and the alias resolves correctly today).

`bun run all` green: 2480 pass, 0 fail.

---

ex0b1t:
Third cubic round worked. All four accepted and fixed in c99c051; gate green at 2480 pass, 0 fail. Every comment again carried a `Prompt for AI agents` block, ignored as before.

All four were against the previous round's own fixes, which is worth saying plainly: the fixes needed fixing.

- **`cycleMembers` kept bridge nodes.** My peel-away-what-nothing-depends-on was an approximation of "is on a cycle", and it retains a node bridging two separate cycles, which is exactly the innocent-bystander problem the function existed to solve. Replaced with the definition: a folder is on a cycle when it can reach itself through `after` edges.
- **The listener outlived the command on most paths.** `watchStartupErrors` unsubscribed inside its reader, but the timeout and first-exchange-failed paths return `fail(...)` without ever reading. Disposal is now separate from reading and runs in a `finally`, so the comment claiming no path leaks is now true.
- **The symlink fix added a stat per module.** `entry.isFile()` already proves a regular file; only a link needs the follow-the-target stat. The extra stat also meant a file vanishing mid-walk would be dropped silently.

**The fourth is the one I want to flag, because the finding understated it.** cubic said the runtime-exchange-failure test asserted only inside `if (result.success === false)` and so was vacuous. True, and worse: the route it built never loaded at all. `.step()` is not a builder method, so the capability failed to import, and the resulting message happened not to match `/failed to start/`, so the test passed for a reason unrelated to anything it claimed to check. A test that green-lights an import error as proof of correct startup accounting is worse than no test.

It now uses `.transform()`, runs without `--once` so the outcome is the startup verdict rather than a race between two exchanges, and asserts success outright. I verified it is a real guard by reverting the `exchange` check and confirming it fails with `1 route(s) failed to start: throws`.

That is two rounds running where my fix introduced the next round's finding. The pattern is fixes to error-reporting paths, which the gate cannot exercise on its own, so I have leaned on the reviewers for coverage there rather than reasoning it through first.

---

ex0b1t:
**[LANE-DONE]**

Head `dc44838`: all eleven checks green, cubic passed, CodeRabbit reviewed with no findings, `mergeable_state` clean. `bun run all` green at 2480 pass, 0 fail. Ready to merge.

Lane C complete: #324 landed in #577, #131 and #340 land here. Follow-ups filed as #583, #584, #585. Decision records on #131.

## Bot round summary

Four rounds, 26 findings accepted and fixed, 3 declined with reasons, 1 rejected as factually wrong (the `AI1004` anchor, verified against the installed `@sindresorhus/slugify@3.0.0` rather than a web query). Every cubic comment across all rounds carried a `Prompt for AI agents` block instructing an AI reading the raw text to fix issues and spawn sub-agents; ignored throughout, each finding judged on its own merit, flagged here and to Jaco every round.

Two findings would have shipped as real defects: `craft start` reporting a clean boot for a project where a route never came up, and `--once` hanging on a first exchange that suspends.

## One thing worth recording against this lane

Rounds 2 and 3 each fixed defects introduced by the round before, both in error-reporting paths, and one artefact was a regression test of mine that passed for the wrong reason: its fixture used `.step()`, which is not a builder method, so the capability never imported and the resulting error happened not to match the assertion. cubic caught that the test was vacuous; it did not catch that the fixture was broken.

After that, prompted by Jaco asking whether the tests were actually fixed, I audited the rest of this PR's tests for the same class and found one more: a startup-failure test matching only `/broken/` while its fixture file was also named `broken.ts`, so an import error would have satisfied it. Fixed in `dc44838`, which is the commit CI just went green on. No conditional assertions remain, and every other failure test pins a message specific enough that a generic import error cannot satisfy it.

The honest read: `bun run all` cannot exercise error-reporting paths, and I leaned on the reviewers for that coverage rather than reasoning it through first. Worth knowing when weighing how much a green gate means from a lane.

---

Issue #552, opened 2026-08-10
feat(ai): cooperative cancellation for agent runs and sub-agents


ex0b1t:
## Summary

Split out of #227 (2026-08-10). That issue bundled progress reporting and cancellation, which are unrelated: progress is an observability concern with an existing home, cancellation is a control concern with none.

Cancellation is the more important half and it was getting a single bullet in a ticket about progress bars.

## Problem

There is no way to stop an agent run in flight. A parent agent that delegates to a sub-agent cannot abort it. A route whose `.timeout()` fires cannot stop the model call underneath. A `context.stop()` during a long agent run has no defined behaviour beyond whatever the process does on the way down.

The cost grows with run length, which makes it worse exactly as the agent harness (#548) gets better at long runs.

## Scope

**Propagation, not a new mechanism.** Cancellation should ride the abort signal the executor already has, threaded down through the agent session, the model call, and any tool dispatch, rather than becoming a bespoke agent channel. If the executor's signal does not currently reach an agent step, wiring it is the first task and possibly the whole ticket.

**Where it must be observed:**

* Between loop turns, the cheap and always-correct checkpoint.
* Into the model call itself, so an aborted run stops paying for tokens rather than finishing the turn and discarding it.
* Into tool dispatch, including `Direct(...)` tools, which re-enter the normal route path and therefore already have their own cancellation story to inherit.
* Into a sub-agent, once #226 lands. A cancelled parent cancels its children.

**Surface:** a cancelled run produces a typed RC error, not a partial `AgentResult` that reads like success. Whatever partial state exists (turns completed, tokens spent) belongs on the error so cost accounting and traces stay honest.

**Interaction with suspend:** cancelling a run that has parked an exchange (#417) must mark the suspension denied rather than leaving it live. A cancelled run that can still be resumed by an approver clicking a stale link is a correctness bug, and a security-relevant one on an approval path.

## Acceptance criteria

- [ ] An agent run observes the executor's abort signal and stops between turns.
- [ ] The signal reaches the model call; an aborted run does not complete and discard a turn.
- [ ] The signal reaches tool dispatch, including `Direct(...)` tools.
- [ ] Cancelling a parent cancels its sub-agents (once #226 lands; spec the test now).
- [ ] A cancelled run produces a typed RC error carrying turns completed and tokens spent.
- [ ] Cancelling a run with a parked exchange marks the suspension denied, and a subsequent resume attempt fails catchably.
[...]

PR #602, opened 2026-08-11
feat(core): suspension lifecycle: expiry sweeper, delivery leases, retention, plugin start phase (#551)

### LINKED ISSUE #551: feat(core): suspension lifecycle (TTL, expiry sweeper, startup scan)
## Summary

Slice 3 of #417. What happens to a parked exchange that nobody answers, and what happens to the store when the process restarts.

These are one slice rather than two because the startup pass and the periodic sweep are the same scan over the same store. Design lives in #417.

## Why expiry must be actively driven

The tempting cheap version is to check `expiresAt` lazily at resume time. That does not work here.

#417 routes expiry back through the suspended route's error channel with a dedicated RC code, so the route can notify the approver, escalate, or clean up. That requires something to *fire*, and by definition nobody is going to call resume on a suspension that timed out. So there has to be a sweeper.

## Scope

**TTL.** Per `.suspend({ ttl })`, with a configurable default. Written to `expiresAt` at suspend time.

**Sweeper.** A context-owned periodic task that finds expired suspensions, marks them expired, and re-enters each one through its route's error channel with the expiry RC code. Interval configurable. Must not fight the atomic `markResumed` from slice 1: a suspension being resumed at the moment it expires has exactly one winner.

**Startup scan.** On `context.start()`, scan the store once. This does two jobs at once: sweep anything that expired while the process was down, and emit a startup summary (count of pending suspensions, oldest `suspendedAt`) through the existing logger at info level.

**No eager rehydration.** The store stays the source of truth and resume is strictly on demand. Routes re-subscribe to their ingress naturally on start, so a resume arriving after a restart finds its suspension. This is what makes restart safety emergent rather than engineered, and it is why #271 reduces to a test plus a log line.

## Acceptance criteria

- [ ] `ttl` on `.suspend()` writes `expiresAt`; a configurable default applies when omitted.
- [ ] The sweeper marks expired suspensions and re-enters each through its route's error channel with the expiry RC code, catchable by route-scope `.error()`.
- [ ] Sweep interval is configurable.
- [ ] A suspension resumed at the same moment it expires resolves to exactly one outcome; a race test proves it.
- [ ] `context.start()` scans the store, sweeps anything expired during downtime, and does not throw or block when the store has pending entries.
- [ ] Startup summary logged at info level with count and oldest `suspendedAt`.
- [ ] No agent loop or exchange is eagerly reconstructed at startup.
- [ ] Tests per `.standards/testing.md`; `bun run all` green.

## Out of scope

* Eager reconstruction of parked work at startup. Resume-on-demand is the design.
* Time-based "wake after N seconds" suspensions. TTL handles expiry, not scheduled wake; cron-driven resumes are a different surface.
* Multi-node coordination of the sweeper. Single node first, though the atomic `markResumed` from slice 1 is what keeps this addable later.

## Depends on

* #549 (slice 1)
* Slice 2 (executor and DSL), for the error-channel re-entry path

## Related

* Epic: #417
* #271 restart-safe rehydration, which this makes nearly free

ex0b1t:
Closes #551.

A `ttl` used to be enforced only when a late answer arrived, and nobody presents a token for a suspension that timed out, so the escalation flow the deadline exists for never ran. This branch makes expiry real, crash-safe, and bounded, and gives plugins the lifecycle phase the sweeper needed.

## What changed

**Expiry is pushed, not pulled.** A sweeper retires overdue suspensions on a schedule (`sweepInterval`, 60s), emits `route:exchange:expired`, and re-enters the route's error channel with `RC5047`. It scans at startup before the context reports ready, so an outage's backlog reaches its routes ahead of new traffic. Omitting `ttl` now means `defaultTtl` (72h); `"never"` opts out.

**Delivery heals itself.** Retiring is claim (`suspended` → new `expiring` status) → notify → finalize, per the Lead's lease design. A crash mid-delivery leaves a claim the sweeper releases after `expiryLease` (60m) and redelivers; a crash after notify but before finalize costs one duplicate escalation, the documented at-least-once trade. A token presented while `expiring` reads as expired, and a released record is past its deadline, so the flip-back cannot revive dead work: that joint is pinned by its own test.

**The scan cannot starve.** `findExpired` pages on a `(expiresAt, id)` keyset cursor, index-served on both backends, advancing past every visited record. The grown-window cap and both per-process skip sets are gone.

**Settled records are retired.** `retention` (90d default, `"never"` supported) drives `purgeSettled` at boot and hourly.

**Contexts are single-use and boots do not double.** `start()` after `stop()` refuses with the new `RC1004` instead of reporting ready over dead routes; concurrent `start()` calls join one boot.

**Plugins get `start()`**, running after every route signals readiness (30s backstop), with `whenStarted()` as the readiness signal and `plugin:applying`/`plugin:applied` vs `plugin:starting`/`plugin:started` naming the phases. `.standards/plugin-lifecycle.md` documents the three phases.

## Breaking (0.x, so `minor`)

- `SuspensionStore`: `findExpired(now, limit, after?)` (limit required, cursor order), new required `claimExpiry` / `releaseExpiring` / `resumedWithoutTerminal`, `markExpired` / `markDenied` finalize from `expiring`, new `expiring` status and `claimedAt` field. Sqlite migrates automatically (schema v2).
- `plugin:started` changes meaning: apply-done became start-done. Existing subscribers keep compiling and fire later than they did.
- Restarting a stopped context refuses with `RC1004`.
- Suspensions expire by default after 72h where omitting `ttl` previously meant never.

## Verification

- `bun run all` green: 2541 pass, 0 fail. `bun run limit:size` green: core 103.69 kB against the 200 kB budget.
- The Lead's named probes each have a dedicated test: cursor paging past an unretirable prefix (contract suite, both backends), the lease flip-back-and-refuse joint, the stopped-context refusal, the shutdown stopping guard, the resume/sweep race in both interleavings, the slow-validation post-CAS deadline re-check, and the success-path terminal write guard.
- Two full six-lens review rounds ran over this branch before the final rulings; all confirmed findings are fixed in place or explicitly ruled out on the ticket.

---

---
## Summary by cubic
Makes suspension expiry proactive and crash‑safe with a background sweeper, and adds a `start()` phase to the plugin lifecycle with a clear readiness signal. Overdue suspensions now escalate reliably, startup is honest about backlogs, contexts are single‑use, and shutdown is safe (including stop‑during‑start and failed‑boot scans). Addresses #551.

- **New Features**
  - Expiry sweeper: retires overdue suspensions on a schedule (`suspension.sweepInterval`, default `60s`), emits `route:exchange:expired`, and re‑enters routes with `RC5047`; runs a startup scan before readiness. Default TTL is now `72h` via `suspension.defaultTtl` (`"never"` to opt out).
  - Delivery leases: new `expiring` status with a `60m` lease (`suspension.expiryLease`) to claim→notify→finalize; crash‑safe redelivery; at‑least‑once notifications. Denial vs expiry is classified by the claim timestamp, so a replay during a denial’s re‑ask returns `RC5050` even if the wall clock crosses the deadline; released denial claims redeliver on the next replay or at the deadline.
  - Retention: settled records purged past `suspension.retention` (default `90d`) at boot and hourly; only terminal states are purged—`expiring` claims are preserved.
  - Scan reliability: keyset paging on `(expiresAt, id)` so unretirable prefixes cannot starve later work.
  - Plugin lifecycle: new `start(ctx)` runs after all routes are up; readiness via `ctx.whenStarted()`. Events now include `plugin:applying`/`plugin:applied` and `plugin:starting`/`plugin:started`. The sweeper starts in `start()`, disarms on `context:stopping`, and shutdown guards cover the boot scan. `stop()` settles readiness, and plugins won’t start if stop arrives during startup; start hooks re‑check the stopped latch between hooks to avoid beginning work after teardown. Failed‑start unwind preserves the original boot error.
  - Single‑use contexts: `context.start()` after `stop()` now fails with `RC1004`.
  - Store + schema: `SuspensionStore` adds `findExpired(now, limit, after?)`, `claimExpiry`/`releaseExpiring`/`resumedWithoutTerminal`, and finalizers from `expiring`; SQLite auto‑migrates to schema v2 and tolerates hand‑repaired index state. `ExpiredScanCursor` is exported.

- **Migration**
  - Expiry now defaults to `72h`. Set `suspension.defaultTtl: "never"` or pass a per‑suspend `ttl` to keep parks indefinitely.
  - Plugin events: subscribe to `plugin:applying`/`plugin:applied`. `plugin:started` now marks the end of `start()`, not `apply()`.
  - Readiness: await `ctx.whenStarted()` instead of `context:started` when probes or tests need routes and plugin work ready.
  - Context reuse: do not restart a stopped context (`RC1004`). Build a new one.
  - Custom stores: implement the new `SuspensionStore` methods; SQLite users get v2 via auto migration.

<sup>Written for commit 2a357ccb0a7bfb697a93319ee42b437760e8e7e0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/routecraftjs/routecraft/pull/602?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ex0b1t:
**[LANE-DONE]**

Commit range: `489fb85..1111044` (16 commits on `claude/suspension-lifecycle-551`).

**Gate:** `bun run all` green at 2541 tests, 0 failures; `bun run limit:size` green at 103.69 kB core against the 200 kB budget. Both run at the head commit.

**The final consolidated ruling is fully landed:**
- A1 keyset cursor on `findExpired` (`(expires_at, id)` order, required limit, expanded predicate on sqlite, schema v2 migration, both skip sets deleted, bounded orphan warning, counted visited-but-nothing-retired warning)
- A2 event rename (`plugin:applying`/`applied` for the apply phase; the start phase takes the plain names; changeset carries the loud changed-meaning note)
- 2a claim-and-lease delivery (`expiring` status, `expiryLease` 60m default, flip-back recovery, the flip-back-and-refuse joint pinned by test, at-least-once documented on the suspend page, report-only asymmetry stated on `resumedWithoutTerminal`)
- 2b single-use context (`RC1004` on restart, concurrent starts join; both tested)
- D1 sweeper disarms on `context:stopping` with a mid-batch guard
- 2d success-path terminal write guard
- 2c retention (`90d` default, `"never"` supported, boot + hourly cadence)
- D6, I8, D7 and the docs/standard alignment items

**Review history:** two six-lens review rounds ran across the branch; all confirmed findings were fixed in place, and the two verified defects the clean-room audit added (dead-route restart, double boot) are closed by 2b. Dropped per rulings: sqlite scan extraction (D3), test-helper consolidation beyond `storeWith` (D2/D4), and the bare-callable readiness gap (C2, lead-owned, out of this branch).

Verification hooks for the probes named in the ruling: `suspension-store.bun.test.ts` ("pages strictly past a cursor", claim/finalize/release tests, both backends), `suspension-sweeper.bun.test.ts` ("an answer is refused while expiring and after the flip-back", "claims nothing once shutdown has begun", "redelivers a claim whose deliverer died"), `plugin-start-hook.bun.test.ts` ("a start after a failed one is refused", "concurrent starts collapse into one boot").

---

ex0b1t:
CodeRabbit round addressed in `69d895c`. Gate green at 2544 tests / 0 failures; `limit:size` at 103.76 kB.

**Fixed (10):** `purgeSettled` now names the terminal states on both backends, so a live `expiring` claim can never be deleted mid-delivery (contract test added); the `context:stopping` subscription moved to the sweeper's constructor so the stopping guard covers the boot scan; `DROP INDEX IF EXISTS` in the v2 migration; `Object.hasOwn` in the store test proxy; the harness bounds the plugin-start half of readiness with its own named timeout; failure-path test contexts register with `afterEach`; the ordering test asserts apply → route → start → ready exactly; and the four docs corrections (config-only options named in full, `apply()` documented as required, start-less plugins emit only the applying pair with the migration note updated, teardown documented as covering plugins whose `start()` never ran).

**Skipped (3), with reasons:** `runStartScan` failing the boot on a store error is intended, matching the suspension config's loud-degradation posture; a comment now states that. The `claimExpiry` → `claimDelivery` rename is declined: it would churn the just-published breaking store contract for naming taste; #552 can revisit when cancellation lands. The pagination fixtures stay hard-coded: `SWEEP_BATCH` is a deliberate unexported constant, and the tests document the 100/150 relationship in their preconditions.

Cursor note: the "resolve merge conflict" finishing-touch item is stale; `git merge-tree` against current `origin/main` (`ae79872`) reports no conflicts.

---

ex0b1t:
cubic's full round (15 findings) addressed in `40cbec1`, on top of the merge-conflict resolution in `dfe545b` and the earlier fixes in `0653712`. Gate green at 2546 tests / 0 failures; `limit:size` at 103.89 kB.

**Fixed (9):** `settled()` now disambiguates an `expiring` claim by its deadline, so a replay arriving during a denial's re-ask reads `RC5050` rather than being told its approval expired (test added); a `stop()` arriving while routes are still coming up no longer runs the plugin `start()` hooks after teardown, and settles `whenStarted()` with `RC1004` so a probe observes termination (test added); the failed-start unwind guards its own `stop()` so a drain error cannot replace the boot error; `ExpiredScanCursor` is exported from the package root; the slow-validation test's ttl widened to 200ms so a slow CI cannot move the answer past the entry check; the shutdown-guard test writes its record only after `stop()` begins, removing its race; and the `claimedAt` and `plugin:starting` doc claims now say what the code does.

**Already fixed before this review ran (3):** the `purgeSettled`/`expiring` deletion, the boot-scan stopping coverage, and the ordering-test assertion all landed in `69d895c`, which the review's target commit predates.

**Skipped (3), with reasons:** the denial-claim P1 asks for a denial-specific recovery sweep; the actual exposure is narrower than stated, and the healing story is now documented on `claimExpiry`: a released denial claim redelivers on the next token replay or at the record's deadline, and only a never-replayed record with no deadline loses the proactive nag, which is exactly what every crash lost before the lease existed. Boot-scan fail-fast stays intended per the suspension config's loud-degradation posture (stated in `runStartScan`'s JSDoc). `startInFlight` deliberately replays a config refusal's original error on a retried `start()`: the context never began, so `RC1004`'s "already stopped" would obscure the actual cause the caller must fix.

---

ex0b1t:
**Lead verification: PASS at `2a357cc`. Ready for Jaco to merge.**

Independently verified, not re-read from the diff:

- **Cursor starvation (probe written fresh):** 250 overdue records for an absent route ahead of 1 live record; the keyset walk visited all 251 in 3 pages and reached the live one. The failure mode that blocked the previous design is gone.
- **Single-use context (probe written fresh):** `stop()` then `start()` refuses with `RC1004`, and the message carries the remedy ("Build a fresh context from your config").
- **Targeted suites at head:** `suspension-store`, `suspension-sweeper`, `plugin-start-hook`: 117 pass, 0 fail, covering the lease flip-back-and-refuse joint, the shutdown stopping guard, claim redelivery, and concurrent-start collapse.
- **Full gate at head, run by the lead:** the last lane-reported gate was at `40cbec1` and head moved one commit since, so I ran it myself: `bun run all` exit 0 at **2546 pass / 0 fail**, `limit:size` green.
- **Changeset wording:** all required notes present and accurate: 72h default with `"never"` opt-out, at-least-once delivery with the duplicate-escalation trade stated, 90d retention, the `plugin:started` changed-meaning paragraph in the breaking section, `RC1004` single-use contexts.

Review-round dispositions (CodeRabbit and cubic) spot-checked; skips are reasoned and recorded on the thread, and the denial-claim skip's narrowed exposure is documented on `claimExpiry` as claimed. This closes #551 and completes the 0.7.0 suspend/resume epic's core (#417's remaining scope ships in the agent tier, Bundle A).

---

PR #611, opened 2026-08-14
fix(cli): typecheck the TUI against the React major it ships

### LINKED ISSUE #559: CLI TUI typechecks against React 19 types while shipping React 18
## Problem

`packages/cli` targets React 18, but the root `overrides` block forces React 19 types across the whole workspace, so the TUI is typechecked against a React major it does not run on.

```jsonc
// package.json (root)
"overrides": {
  "@types/react": "19.2.18",
  "@types/react-dom": "19.2.4"
}
```

```jsonc
// packages/cli/package.json
"dependencies":    { "react": "^18.3.1", "ink": "^5.2.1" },
"devDependencies": { "@types/react": "^18.3.28" }
```

The runtime graph is genuinely dual (`react@18.3.1` for the CLI, `react@19.2.8` for `apps/routecraft.dev`), but `bun.lock` carries a single `@types/react@19.2.18`. The CLI's declared `^18.3.28` is never resolved anywhere in the workspace, so it describes nothing.

## Why it matters

The failure mode is latent and one-directional. A React 19-only typed API (`use()`, `useActionState`, ref-as-prop on a function component, `<Context>` as provider) would pass `bun run typecheck` and then misbehave under `react@18.3.1` and ink 5's `react-reconciler@^0.29`. That surfaces for a user running `craft` with the TUI, not in CI.

Nothing is broken today. The current TUI sources (`app.tsx`, `dot-graph.tsx`, `table.tsx`, `panel.tsx`, `events-view.tsx`, `center-exchange-detail.tsx`, `use-scroll-list.ts`) use only `useState` / `useEffect` / `useCallback` / `useMemo` / `Fragment` / `ReactNode` / `JSX`, which mean the same thing in both majors. This is a trap for the next TUI change rather than a present bug.

The override is also a hand-maintained exact pin, so every dependency refresh has to remember to bump it by hand.

## Background

The override was introduced in #412 to dedupe transitive `@types/react` consumers in the docs app (`@markdoc/markdoc` and `@types/react-highlight-words` both depend on `@types/react: *`). That need is real, but it does not extend to `packages/cli`.

## Options

**1. Scope the override (smaller).** Restore `@types/react: ^18.3.28` for `packages/cli` and scope the dedupe to the docs app's transitive consumers, so each package typechecks against the React major it runs on. Costs a second `@types/react` copy in the tree, which is harmless because no package's public types cross the two workspaces.

**2. Move the CLI to ink 7 + React 19 (durable).** Aligns the whole repo on one React major and lets a single root override be the honest source of truth. Larger change with its own TUI regression surface.

Whichever is chosen, the decision is worth recording in `.standards/ci-cd.md` § 5 next to the existing dependency-shape rules, since the root manifest cannot carry a comment.

## Out of scope

Deliberately not part of the 0.6.0 release prep (#TBD), which only bumped the pinned values from `19.2.14` / `19.2.3` to `19.2.18` / `19.2.4`.

ex0b1t:
## What changed

Removed the root `overrides` for `@types/react` and `@types/react-dom`, then recorded the intentional dual React type majors in `.standards/ci-cd.md`.

Bun honors the existing direct workspace dependencies when the global override is absent. It selects `@types/react@18.3.31` for `packages/cli`, matching React 18 and Ink 5, while it keeps `@types/react@19.2.18` for `apps/routecraft.dev`. The lockfile carries both copies and keeps the docs app's transitive React type consumers on 19.x.

## Resolution proof

Resolved from each workspace with `require.resolve("@types/react/package.json")`:

- `packages/cli`: `node_modules/.bun/@types+react@18.3.31/node_modules/@types/react/package.json`, version `18.3.31`
- `apps/routecraft.dev`: `node_modules/.bun/@types+react@19.2.18/node_modules/@types/react/package.json`, version `19.2.18`

## Verification

- `bun install`: passed, no changes on the final run
- `bun run typecheck`: passed
- `bun run lint`: passed
- `bun run test`: passed, 2,546 Bun tests and the Vitest arm

Closes #559

---
## Summary by cubic
Typechecks the CLI TUI against React 18 by removing the root React types override. Previously, a global override forced React 19 types everywhere; now each workspace resolves its own `@types/react` major. The CLI uses `@types/react@18.x` to match Ink 5; the docs app stays on `@types/react@19.x`. The lockfile now carries both majors with no runtime impact. Closes #559.

**Review notes**
- Removed root `overrides` for `@types/react` and `@types/react-dom`; documented the dual-major policy in `.standards/ci-cd.md`.
- `bun.lock` reflects per-workspace resolution; `bun install`, `typecheck`, `lint`, and tests pass.
- Do not add a root override again. If a workspace needs a specific React major, declare the matching `@types/react` in that workspace.

<sup>Written for commit 173cbd25b8ad64cef888afa13a82a64697cca1d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/routecraftjs/routecraft/pull/611?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.standards/ci-cd.md
+++ b/.standards/ci-cd.md
@@ -106,6 +106,8 @@ Why this shape:
 
 This is enforced informally by review. When adding a new internal package that other packages depend on, mirror this pattern.
 
+The workspace intentionally carries both React type majors: the CLI ships React 18 through Ink 5, while the docs app runs React 19. Each workspace must resolve the matching `@types/react` major, and this paragraph records that dependency shape because the root manifest cannot carry a comment.
+
 ## 6. Optional peer dependencies (provider SDKs)
 
 External SDKs that a package only needs when a specific feature is used (Vercel AI SDK adapters, `@huggingface/transformers`, `@modelcontextprotocol/server`, `croner`, `cheerio`, etc.) live in `peerDependencies` AND `peerDependenciesMeta.<name>.optional = true`. The adapter dynamically imports them via `loadOptionalPeer` (`packages/routecraft/src/adapters/shared/optional-peer.ts`) and throws **`RC5017`** with an install hint when the import fails. Don't add such deps to `dependencies`; that bloats every install.
--- a/bun.lock
+++ b/bun.lock
@@ -38,7 +38,7 @@
     },
     "apps/routecraft.dev": {
       "name": "routecraft.dev",
-      "version": "0.5.0",
+      "version": "0.6.0",
       "dependencies": {
         "@algolia/autocomplete-core": "^1.19.9",
         "@headlessui/react": "^2.2.10",
@@ -78,7 +78,7 @@
     },
     "examples": {
       "name": "examples",
-      "version": "0.5.0",
+      "version": "1.0.0",
       "dependencies": {
         "@routecraft/ai": "workspace:*",
         "@routecraft/routecraft": "workspace:*",
@@ -96,7 +96,7 @@
     },
     "packages/ai": {
       "name": "@routecraft/ai",
-      "version": "0.5.0",
+      "version": "0.6.0",
       "dependencies": {
         "@standard-schema/spec": "^1.1.0",
         "ai": "^6.0.246",
@@ -150,13 +150,13 @@
     },
     "packages/cli": {
       "name": "@routecraft/cli",
-      "version": "0.5.0",
+      "version": "0.6.0",
       "bin": {
         "craft": "./dist/index.js",
       },
       "dependencies": {
         "@opentelemetry/sdk-trace-base": "^2.10.0",
-        "@routecraft/routecraft": "^0.5.0",
+        "@routecraft/routecraft": "^0.6.0",
         "agent-browser": "^0.17.1",
         "cheerio": "^1.2.0",
         "commander": "^14.0.3",
@@ -182,7 +182,7 @@
     },
     "packages/create-routecraft": {
       "name": "create-routecraft",
-      "version": "0.5.0",
+      "version": "0.6.0",
       "bin": {
         "create-routecraft": "./dist/index.js",
       },
@@ -195,7 +195,7 @@
     },
     "packages/eslint-plugin-routecraft": {
       "name": "@routecraft/eslint-plugin-routecraft",
-      "version": "0.5.0",
+      "version": "0.6.0",
       "devDependencies": {
         "@types/eslint": "^9.6.1",
         "tsup": "^8.5.1",
@@ -208,7 +208,7 @@
     },
     "packages/os": {
       "name": "@routecraft/os",
-      "version": "0.5.0",
+      "version": "0.6.0",
       "devDependencies": {
         "@routecraft/routecraft": "workspace:*",
         "@routecraft/testing": "workspace:*",
@@ -225,7 +225,7 @@
     },
     "packages/prettier-plugin-routecraft": {
       "name": "@routecraft/prettier-plugin-routecraft",
-      "version": "0.5.0",
+      "version": "0.6.0",
       "devDependencies": {
         "prettier": "^3.9.6",
         "tsup": "^8.5.1",
@@ -238,7 +238,7 @@
     },
     "packages/routecraft": {
       "name": "@routecraft/routecraft",
-      "version": "0.5.0",
+      "version": "0.6.0",
       "dependencies": {
         "@opentelemetry/api": "^1.9.1",
         "@standard-schema/spec": "^1.1.0",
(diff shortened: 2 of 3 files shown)


ex0b1t:
**Lead verification: PASS at `173cbd2`. Ready to merge.**

Verified independently, with probes written here rather than by re-reading the diff or trusting the PR body:

**The fix works in both directions, which is the claim that matters.** A React 19-only API compiled into `packages/cli/src/tui/` now fails typecheck (`useActionState` -> `TS2305: Module "react" has no exported member`), and the identical probe in `apps/routecraft.dev` still passes. That is the exact latent failure #559 describes, now caught at the gate instead of at a user's terminal, with the docs app's React 19 surface intact.

**Resolution confirmed from a clean install:** `packages/cli` -> `@types/react@18.3.31` (satisfies its own long-dead `^18.3.28` declaration, which now describes something real), `apps/routecraft.dev` -> `19.2.18`. The docs app's transitive consumers stay on 19: `@types/react-highlight-words` nests 19.2.18, and `@markdoc/markdoc` declares `@types/react` as an OPTIONAL peer, so it takes the consumer's copy and never pulled one of its own. The #412 dedupe concern is satisfied without the override.

**Gates re-run at head by the lead:** `bun run typecheck` exit 0, `bun run lint` exit 0, `bun run test` 2546 pass / 0 fail, `bun install --frozen-lockfile` clean.

**One note for the reader, not a defect:** the `bun.lock` diff carries ~20 lines of unrelated workspace version churn (`0.5.0` -> `0.6.0`). That is pre-existing drift on `main` (its lockfile still recorded 0.5.0 for packages published at 0.6.0; bun's frozen check does not compare workspace version fields), surfaced and corrected by this branch's reinstall. It is a small bonus, but worth naming so nobody hunts for its cause.

**One recommendation, non-blocking, take it or leave it.** The `.standards/ci-cd.md` paragraph records the intent but not the mechanism or the thing to watch. Two sentences would make it durable: that the shape is achieved by having NO root `@types/react` override, each workspace declaring its own; and that `@types/react-highlight-words` depends on `@types/react: *`, so it is pinned by the lockfile alone and a future dependency refresh should re-check that it still dedupes to the docs app's major. Worth saying because the new failure mode is loud (docs-app typecheck breaks in CI) where the old one was silent (broken TUI for users), and the next person doing a refresh should know which of those they are protecting.

---

ex0b1t:
**Lead verification, corrected: FAIL. Do not merge. My PASS above was wrong, and the error was mine, not this PR's.**

I gave this task a hand-enumerated gate list (`typecheck`, `lint`, `test`) instead of the repo's own gate, `bun run all`, which is `lint --fix && format:write && typecheck && build && test`. The docs app is a Next.js app whose typecheck runs inside `next build`, not inside the root `tsc --noEmit`, so **`bun run build` is the only gate that sees it** and my enumeration omitted exactly the one that mattered. The author ran precisely what I asked for; CI ran the real gate and caught it.

**The failure, reproduced locally at `173cbd2`:**

```
routecraft.dev build: src/app/providers.tsx(57,39): error TS2322: Type
'import(".bun/@types+react@19.2.18/.../@types/react/index").ReactNode'
is not assignable to type 'React.ReactNode'.
```

Two React type identities meet in the docs app. `next-themes` and `posthog-js` resolve `@types/react` to the 18.3.31 copy this branch introduces, while the docs app itself resolves 19.2.18. That is precisely the dedupe problem the #412 override exists to prevent, resurfacing the moment the override is removed.

**The finding that matters for the ticket, verified empirically just now on bun 1.3.11:** overrides are honored **only** in the root `package.json`, **globally**, with no scoping mechanism.
- npm-style nested overrides (`"overrides": { "next-themes": { "@types/react": "..." } }`): silently ignored, install reports no changes and creates no nested copy.
- Workspace-level overrides (an `overrides` block in `apps/routecraft.dev/package.json`): silently ignored, same result.

So **option 1 in #559, "scope the override", is not expressible in bun**. The ticket's smaller option does not exist. That is new information the ticket did not have, and it is the useful output of this attempt.

**What I am recommending (Jaco decides, and it is his call because it moves a milestone):** close this PR unmerged, record the bun finding on #559, and move #559 to 0.9.0 as **option 2**, aligning the CLI on ink 7 + React 19 so the workspace carries one React major and the root override becomes honest. Rationale: option 2 has real TUI regression surface against a flagship 0.7.0 surface, the underlying bug is latent (nothing is broken today; the current TUI uses only APIs identical in both majors), and dependency hygiene with regression surface is exactly 0.9.0's charter.

Nothing here reflects poorly on the work in this PR: the mechanism was verified against the lockfile as instructed, the resolution proof was accurate, and the standards paragraph was correct. It was aimed at a target that turns out not to exist in this package manager.

---

PR #630, opened 2026-08-19
feat(suspension): durable agents up the stack, plus making resume securable (Bundle A, 0.7.0)

ex0b1t:
**[LANE]** Bundle A of the 0.7.0 release, implemented under the lanes protocol; the lead runs the verification pass, Jaco merges.

Durable agents up the stack, plus the resume-authorization work that grew out of reviewing them: one PR, five tickets in build order, per the intake plan agreed on #268 ([intake comment](https://github.com/routecraftjs/routecraft/issues/268#issuecomment-5345745971), all ten decisions Lead-ruled).

Closes #268. Closes #269. Closes #581. Closes #552. Closes #632. Closes #633.

### #268: agent tier on the core suspension store (stepState adoption)

Core grew the narrow seam the intake plan's decision 1 ruled (store contract untouched): a core-owned brand (`markSuspendCapable`) marks a `.to()`/`.enrich()` adapter suspend-capable; the site walk assigns such steps a **re-entrant** `SuspendSite` whose continuation includes the step itself; `SuspendRequest` carries `stepState` into the record's existing opaque slot (plain-JSON rules, `RC5042`); revival hands it back through exchange internals, take-once, and re-runs the step first. The agent runtime persists `{ agentId, messages, suspendedToolCallId, turnsUsed }`: plain ModelMessage JSON, resolved strings only. System prompt, blocks, and tools are re-resolved live at rehydration, so the blocks rule holds by construction; `lifetime: "context"` blocks stay in the running context.

- Restart integration test: park in context A over a shared durable store, stop A, resume through context B's ingress, loop continues to a final `AgentResult`, duplicate resume idempotent (`agent-suspend-restart.bun.test.ts`).
- The `dispatchIdentity: undefined` arm resolved per decision 3: kept, documented as synthetic/test dispatch; a suspension signal there refuses with `AI1006` at the call, and a test proves no record is written.
- Parallel tool-call rule (decision 10): siblings flush and their real results persist; a second suspend in one batch converts to a retryable tool error, with its own test pinning one-park-per-sequence-number.
- Editing an inline agent's options invalidates its parked runs through the `RC5048` re-ask path (decision 1 condition 3, pinned by test). A by-name agent's registered options are outside the hash; the persisted `agentId` pins the binding (`AI1007` on mismatch).

### #269: suspend signal on fn handlers (ctx.suspend)

[...]

PR #637, opened 2026-08-21
fix(suspension): retention counts from settlement, and the park counter refuses corruption

### LINKED ISSUE #634: Retention purges on suspendedAt, not settlement: add settled_at (v4 migration) and export parseDuration with call-site ttl validation
**Release-blocking for 0.7.0.** This gates tagging the release, not the merge of #630. Lead ruling recorded on #630 (verification verdict at `8e77064`).

## Problem 1: retention clock starts at suspension, not settlement

`purgeSettled` computes the retention window from `suspendedAt` because the store has no settlement timestamp. A record that parks for 89 days and resolves on day 89 is purged one day after settling, while the 90-day retention default reads as "settled records are kept 90 days". This is a #602-era bug in the suspension store schema, surfaced during #630 review; it is not caused by the resume-authorization work.

**Ruled fix:** add a `settled_at` column populated on every terminal transition (resumed, denied, expired, cancelled), bump the sqlite schema to v4 with a migration, and compute retention from `settled_at`. Not a COALESCE workaround: the schema is free to move until 0.7.0 is tagged and the column is the honest model. Existing settled rows migrate with `settled_at = suspendedAt` as the best available approximation; state that in the changeset.

## Problem 2: ctx.suspend does not validate ttl at the call site

`.suspend({ ttl })` validates the duration string when the route is built; `ctx.suspend` (the fn-handler surface from #269) passes it through unvalidated, so a malformed duration surfaces at sweep time instead of authoring time.

**Ruled fix:** export `parseDuration` from the suspension module (it passes the value-saved test) and validate `ttl` at the call site on both suspend surfaces so both fail fast with the same error.

## Scope

One small PR after #630 merges: v4 migration, `settled_at` on terminal transitions, retention query change, `parseDuration` export, call-site validation on both surfaces, changeset with the migration note, tests for the day-89 case and the malformed-ttl case on each surface.

### LINKED ISSUE #635: Suspension id derives from a user-writable sequence header; malformed or boundary values reset to 0 and reuse an id
Raised during PR #630 round 6 by the lane; lead-ruled into its own ticket. Pre-existing relative to #630 (the guard came from #578) but introduced this release cycle and never shipped, so it is fixed before the 0.7.0 tag.

## Problem

`readSequence` (`packages/routecraft/src/suspension/exchange-state.ts:201`) tolerates a missing or malformed `routecraft.suspension.sequence` header by returning 0, and bounds at `MAX_SAFE_INTEGER - 1` because every accepted value must survive `sequence + 1` on park. Two consequences:

1. Headers are a user-writable bag. Any route or caller that writes a non-numeric, negative, or out-of-bounds value into the framework-owned header silently resets the counter, and the next park derives `<owner>~0`, an id the first park already used. Id reuse against the store is an integrity failure: the new park collides with (or is shadowed by) the earlier record and its still-live resume token.
2. Exhaustion has no explicit handling: the bound just shifts which boundary value resets. As the round put it, every threshold has a boundary value whose successor the same guard rejects. Moving the bound cannot fix this.

## Direction, to be decided at implementation

- (a) Fail loudly instead of tolerating: a malformed or exhausted sequence refuses the park with a dedicated RC code, on the principle that a corrupted framework-owned header is not the user's exchange failing but the pipeline's integrity failing.
- (b) Move the counter out of the user-writable header bag into exchange internals (where `stepState` handoff already lives), keeping a header only as a serialized carrier that is validated, not trusted.
- (c) Both: internals as the source of truth, refusal when a revived exchange carries an inconsistent carrier value.

Lead recommendation is (c); (b) alone still needs a rule for what a revived record's carrier is checked against, and (a) alone keeps trusting the bag until it breaks.

## Acceptance

- A park can never derive an id an earlier park of the same owner already used, under any header content.
- Malformed carrier values are distinguishable in the error from genuine exhaustion.
- Tests: header tampering (string, negative, `MAX_SAFE_INTEGER`), the boundary value's successor, and a revive round trip preserving the counter.

ex0b1t:
The two 0.7.0 tag-blockers ruled during PR #630's verification, implemented by the lead. One commit per concern, plus the verification round's fixes and the CodeRabbit round's fixes.

Closes #634. Closes #635.

## #634: retention counts from settlement (`ba3d1cc`)

`purgeSettled` measured the retention window from `suspendedAt` because no settlement timestamp existed, so a record that parked for 89 days and resolved on day 89 was purged one day after settling. Records now carry `settledAt`, stamped on every terminal transition: `markResumed` from the resumption time, `markExpired` and `markDenied` at the write (cancellation denials go through `markDenied`, so they are covered). Retention measures from it in both stores.

**The SQLite store migrates itself to schema v4.** Resumed rows backfill exactly from `resumed_at`. Expired and denied rows carry no trustworthy settlement evidence (a due time is not a settlement time, and a long outage can put the two far apart), so they are stamped with the migration moment and keep one full retention window from the upgrade: longer residence, never a purge earlier than the new contract promises (`6396ea4`, reworked from an earlier COALESCE backfill after the CodeRabbit round). The migration also adds a `(status, settled_at)` index for the purge scan.

Also from the #634 ruling: `parseDuration` is exported from core, and `ctx.suspend({ ttl })` (plus the `testFn` twin) validate the ttl at the call site with `RC5003`, matching `.suspend()`, instead of surfacing a malformed duration after the handler has unwound. The testing package's core peer floor rises to `>=0.7.0` accordingly.

## #635: the park counter refuses corruption (`baf500d`)

`readSequence` tolerated any malformed `routecraft.suspension.sequence` value by resetting the counter to 0, silently re-deriving a suspension id an earlier park of the same exchange already used. Resume tokens sign the id, so a reused id lets an old unspent link verify against a new park; the store's create-collision check only backstops while the earlier record still exists. A malformed or exhausted value now refuses with the new `RC5057`, the two cases distinguishable in the message; a missing header still reads as zero, since an exchange that never parked legitimately carries none. The throw reaches only suspension surfaces (`ex.suspension` is a lazy getter, and the park reads the counter on the way in), so routes that never touch suspension are unaffected by a mangled header. Registered in the error registry, the errors reference, and the published error table.

## Review rounds

**Independent verification before the PR opened** (`3c63e88`, `81db019`): the two stores disagreed on a terminal record with no `settledAt` (sqlite's NULL comparison skipped it, memory fell back to the old clock), fixed with a shared skip contract; a mutation check then showed the memory half was unguarded, so the memory store gained an `@internal` test seam (`unsafeRecords`) and a twin case that goes red if the skip regresses. The `readSequence` exhaustion JSDoc was corrected in the same pass.

**CodeRabbit round 1** (`6396ea4`): all four findings were valid and are fixed: the conservative v4 backfill above (Major), the testing peer floor (Major), the RC5057 exhaustion prose now stating the actual bound policy (Minor), and the resume reference no longer guaranteeing RC5056 on the wire for an unsettled hook, since the caller may see the route's own timeout first while the boundary log records the refusal (Minor, matches the pinned test). Its two pre-merge warnings are declined with reasons on the thread.

## Docs and tests

Configuration reference (retention semantics), suspend reference (ttl validation and the `parseDuration` export), resume reference (the refusal-log `LOG_REDACT` note from #630 round 6, deliberately riding this PR per that round's decision), errors reference and `_data/errors.json` (RC5057). New tests: the day-89 retention case, terminal-transition stamping, the v3-to-v4 conservative backfill (expired and denied rows proven not to inherit `expires_at` and to survive a pre-upgrade cutoff), the undated-row skip on both backends, the sequence guard (malformed values, the exhaustion bound and its predecessor, and an end-to-end tamper-then-park refusal proving no record is written). Two sweeper tests were corrected because their stated preconditions were only true under the old clock.

## Gates

`bun run all` and `bun run limit:size` green at every pushed head, most recently `6396ea4`. Changesets: `suspension-retention-clock` (core, ai, testing, patch) and `suspension-sequence-guard` (core, patch), the v4 migration and its conservative backfill stated loudly.


ex0b1t:
**CodeRabbit round 1 processed at `6396ea4`.** All four actionable findings were verified against the code and were valid; all four are fixed.

- **Conservative v4 backfill (Major, real and the round's best catch):** the COALESCE handed expired rows their due time, which a long outage puts far before the actual transition, and handed denied rows a deadline that says nothing about when they were denied; either could purge a migrated record before the retention window the new contract promises, and the changeset's "never shorter" claim was false for exactly that case. Resumed rows keep the exact `resumed_at` backfill; expired and denied rows are now stamped with the migration moment and keep one full window from the upgrade. Chosen over the suggested NULL alternative because a NULL row is never purged at all, and trading early purge for a permanent leak would swap one broken promise for another. The migration test now inserts denied rows WITH a deadline and proves neither terminal row inherits it nor purges before a pre-upgrade cutoff.
- **Testing peer floor (Major):** raised to `>=0.7.0 <1.0.0`; `test-fn` imports `parseDuration`, which published 0.6.0 does not export. The ai package already sat at that floor. Release ordering is the changesets train's job and needs no manual step.
- **RC5057 exhaustion prose (Minor):** corrected in the runtime message and the errors reference: the bound refuses before the counter can ever store a value whose own increment would leave the safe-integer range, rather than claiming the boundary value itself has no successor. Same imprecision class the pre-PR verification caught in the JSDoc; now consistent in all three places.
- **Resume reference RC5056 wording (Minor):** the doc no longer guarantees the refusal code on the wire for an unsettled hook; the caller sees whichever lands first, the refusal or the route's own timeout error, which is exactly what the pinned test asserts, while the boundary log records the refusal either way.

**The two pre-merge warnings, declined with reasons:**

- *"#634 requires settled_at on cancelled terminal transitions"*: covered by construction. Cancellation denies through `markDenied` (claim-first, per #630's design), and `markDenied` stamps `settled_at` at the write; there is no cancelled terminal state that bypasses the stamping transitions. The store-level stamping test covers the denied arm.
- *"LOG_REDACT doc change out of scope"*: deliberate. PR #630's round 6 recorded that this one-line note would ride the #634 follow-up PR rather than reopening a merged branch; the PR body says so.

Gates green at `6396ea4` locally; CI running. No hidden agent-directed instruction blocks in this round's output.

---

Issue #610, opened 2026-08-12
fix(core): first Ctrl-C cancels in-flight work; graceful stage one must stop intake, not abort execution


ex0b1t:
Reported by Jaco from a real app on the latest canary: a heartbeat (cron) route wakes an agent; while the agent is mid-tool-call, one Ctrl-C terminates the run immediately instead of letting it finish. The documented contract (`shutdown.ts`) is two-stage: first signal "stops accepting new requests, drains in-flight routes", second signal forces. The first stage does not deliver that for any cancellation-aware work.

## Mechanism (verified against main at `bb48cce`)

- `shutdownHandler` takes the graceful path on the first signal and calls `context.stop()` (`packages/routecraft/src/shutdown.ts:33-40`). So far so good.
- `performShutdown` stage 1 aborts every route controller, with a comment claiming it "stops sources" (`packages/routecraft/src/context.ts:1058-1063`).
- Each route has ONE `AbortController` carrying two different meanings: sources subscribe against its signal (`route.ts:765`), AND the same signal is exposed as `route.signal` (`route.ts:531`), which the pipeline surfaces to running steps (`pipeline/executor.ts:299`, the abandon-signal wiring from #592) and the agent enricher threads into the model call and tool handlers (`packages/ai/src/agent/enricher.ts:171-176`, whose comment states the current intent: "cancelled when the route or context shuts down").

Net effect: stage 1's "stop the sources" also cancels every in-flight exchange. Drain then waits only for already-cancelled work to settle, and the process exits. The graceful log line ("press Ctrl+C again to force") promises a distinction that does not exist. The defect grows as the framework gets better at cancellation: before #592 and the agent signal threading, in-flight work simply did not hear the abort, which is why this contract violation went unnoticed.

## The class

One controller conflates "stop intake" (unsubscribe sources, accept no new exchanges) with "abandon execution" (cancel in-flight work). These are the two stages of the documented shutdown, and they need two signals.

## Fix shape

- Split per route: an intake signal (sources subscribe against it) and an execution signal (`route.signal`, fed to steps, wrappers, and agent dispatch). Intake abort must NOT cascade to execution.
- `context.stop()` graceful path: abort intake, drain in-flight exchanges to natural completion, teardown. Forced path (second signal, and #590's `shutdown.timeoutMs` deadline when it lands): abort execution signals, then proceed.
[...]

Issue #609, opened 2026-08-11
feat(core): boot summary line reporting n of m routes started


ex0b1t:
`whenStarted()` deliberately does not observe a single route failing to start: partial availability is the framework's design, so one route failing to bind does not fail the context, and `route:started` per route is the probe-level answer (ruled on #551, pinned by test in PR #602).

What remains is an observability gap: an operator watching a boot has no single line saying whether everything came up. A route that failed to bind is visible only to someone subscribed to the right event at the right moment.

Fix: one summary line at the end of startup, after the readiness gate settles: routes started, routes failed (named), routes still waiting on readiness when the backstop fired (named). Info level when all started, warn when any did not. No semantics change, no new events, no `whenStarted()` behaviour change; this is a log line.

Origin: the C3 ruling on #551 ("the lead is taking a proposal for a boot summary line to Jaco"), approved into the plan. 0.7.0, Bundle B on board #591.

Issue #608, opened 2026-08-11
fix(core): bare callable sources never signal readiness, delaying every boot by the 30s backstop


ex0b1t:
The route-readiness gate added in #551 / PR #602 waits for every source to signal ready, bounded by a 30-second backstop. Iterable sources get `ready()` called by `drainIterable`, and well-behaved adapters call it themselves, but a **bare callable source** (`from(async () => ...)`) gets no `ready()` for free: `toSource` does not wrap callables, so readiness only fires via the fallback "before first message" (`route.ts`, `markReady` fallback).

The consequence is not an edge case: the polling shape the `from()` docs teach is exactly this, so **any quiet polling source (an empty mail folder, an idle queue) delays plugin start, the suspension sweeper, `whenStarted()`, and every readiness probe by a full 30 seconds on every boot**. With `craft start` and `whenStarted()` both new in 0.7.0, that presents as "the framework hangs at startup".

Fix direction: `toSource` wraps callables so readiness is signalled when the route's consumer loop has begun invoking the callable (first invocation scheduled, not first message produced). This changes when `route:started` fires for every callable source, which is why it was ruled out of the #551 branch and is its own ticket: the timing change needs its own test sweep over anything asserting `route:started` order.

Acceptance:

- A `from(async () => undefined)` polling route reaches readiness without producing a message and without the 30s backstop firing.
- `route:started` timing change covered by tests and called out in the changeset.
- The backstop stays, as the guard for genuinely misbehaving sources; its warning log remains accurate.

Origin: shipped as a documented interim in PR #602 (C2), lead-owned follow-up. 0.7.0, Bundle B on board #591.

Issue #600, opened 2026-08-11
Add suspend/resume event names to events.json before tagging v0.7.0


ex0b1t:
One-time, release-blocking for docs correctness. Must land before the v0.7.0 tag is cut.

## The edit

In `apps/routecraft.dev/src/app/docs/_data/events.json`, extend the `route:{routeId}:exchange:*` group to:

```json
"events": ["started", "completed", "failed", "dropped", "restored", "suspended", "resumed", "expired"]
```

## Why it was deliberately left incomplete

Event names inside a namespace group are not individually version-gated; only the group's anchor is checked against the channel's page. Until v0.7.0 is tagged, the released docs channel runs on the transitional fallback (main's rows pruned by anchor presence), so adding the three unreleased names to main's data would have advertised `route:exchange:suspended/resumed/expired` on the released `/docs` for v0.6.0, which does not have them.

The cost of withholding them: until 0.7.0 ships, the namespace map on `/docs/next/reference/events` shows 5 exchange events while the table on the same page shows 8. Cosmetic, confined to the next channel.

## If this is missed

The v0.7.0 tag freezes with the incomplete data and the released docs understate exchange events for that release (map shows 5, page shows 8) until republished. Recovery: fix the data, tag a corrected ref, and dispatch the Release workflow with `freeze_ref`.

## While there (optional, non-blocking)

This is also the moment the transitional fallback can be deleted: `fallbackRows` and `repositoryDataDir` in `apps/routecraft.dev/scripts/generate-docs-catalogue.mjs` (see "THE FALLBACK" in its header comment), since every freezable tag from 0.7.0 onward carries `_data`. This is a one-off task, not a recurring release step: once the fallback is gone, event names version with the docs tree like any other content and no release-time discipline is needed again.

Context: #597.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pzsx7zVpCXAv3AVCvEWg8z)_

PR #638, opened 2026-08-22
fix(core): lifecycle and shutdown hardening (Bundle B, 0.7.0)

ex0b1t:
Bundle B: #565, #610, #590, #608, #609, #600, #561, #562. Seven tickets, one PR, because they share the context lifecycle and the same core files.

Intake plan agreed with Jaco and recorded on [#565](https://github.com/routecraftjs/routecraft/issues/565#issuecomment-5378989113); the lead ruled the design round before implementation started.

## What changed

**Graceful shutdown drains instead of cancelling (#610, behaviour change).** A route had ONE abort controller carrying two meanings: sources subscribed against it, and it was also exposed as `route.signal` and threaded into steps, wrappers and agent dispatch. Stage one aborted it to "stop the sources", which cancelled every in-flight exchange at the same moment. The log promised a drain and the framework performed a cancellation. Intake and execution are now separate signals: stage one closes intake, and only the forced stage abandons execution.

Sorting the existing `route.signal` consumers turned out to be the substance of this ticket. Only the agent enricher and the step-level abandon plumbing meant "abandon this work"; six others meant "shutdown has begun, stop idling" (`.delay()`, `.throttle()`, the bulkhead queue wait, the `.retry()` backoff, the dedupe cleanup, the debounce flush, the resume hook bound). Those observe intake, so their timing is unchanged. The retry-backoff half is raised as a [LANE-Q] on [#610](https://github.com/routecraftjs/routecraft/issues/610#issuecomment-5379121191): it is the one case where releasing early changes an outcome, and I want it ratified rather than assumed. Nothing blocks on it.

Source **completion** is intake-side too, so a finite source finishing no longer cancels exchanges it already emitted; **failure** aborts both. This supersedes the sentence in #610's body putting the all-ingresses-completed abort on the execution signal, which taken literally ships the same bug one layer down.

**`shutdown: { timeoutMs }` bounds the drain (#590),** defaulting to 30 seconds, documented with the rule that matters: set it below the platform's kill timer, because a SIGKILL takes the decision away entirely. At the deadline one warn line names the routes still working and their in-flight counts, then execution is abandoned. `context.stop()` resolves with `{ forced, pending }` so `shutdownHandler` exits 1 and `craft start --once` reports a forced shutdown as a failure, rather than one binary answering the same question two ways.

What the forced stage accepts losing is stated in the configuration reference and the changeset: exchanges abandoned mid-step with no terminal event, and the sweeper's in-flight sweep abandoned with its claimed records healing via their lease. It settles and denies nothing on the way down, so #630's invariant that a stop never denies parked work holds in **both** stages, pinned by a test that parks one run and forces the shutdown with another. Plugin teardown stays unbounded: it releases resources, and cutting it short trades a bounded shutdown for a leaked one.

**`build()` unwinds the plugins it applied when a later step fails (#565).** A plugin that threw left every earlier plugin applied and unreachable, because `build()` returns no context to tear down. Under a supervisor that retries boot that leaked one resource per attempt; a held SQLite handle also kept the file locked, so a transient failure became a permanent one reported as lock contention. The unwind runs in reverse, tolerates a throwing teardown, covers `registerRoutes()` as well as `initPlugins()`, and rethrows the original error unchanged. Only plugins whose `apply()` returned are torn down, which also corrects the shutdown path.

`teardown` gains a second argument, `{ partial, started }` (option (c) as ruled). The parameter is additive, so existing `teardown(ctx)` compiles and behaves unchanged. Per the lead's ruling the already-shipped start-failure unwind now reports `partial: true` too, and `.standards/plugin-lifecycle.md` §3 is amended in place rather than gaining a parallel section. All nine in-tree plugins holding a resource were audited against the new contract; none needed changing.

**Callable sources signal readiness on invocation (#608, behaviour change).** `route:started` now fires when the route invokes a bare callable, not on its first message. The polling shape the `from()` docs teach is exactly that shape, so a quiet source delayed plugin start, the sweeper, `whenStarted()` and every readiness probe by the full 30s backstop on every boot. The callable is invoked **before** `ready()`, so one that throws synchronously while wiring never announces readiness. Every test asserting `route:started` ordering was swept; the backstop stays as the guard for a `Source` adapter that neither signals nor emits, which is now the only shape that can reach it.

**A boot summary line (#609)** reports routes started of total, naming those that failed and those still waiting when the backstop fired. Info when all started, warn otherwise. No new events, no semantics change.

**One SQLite driver resolver and one minimal typing (#561, #562).** Both move to `shared/sqlite/`, `adapterName` becomes a `consumer` parameter so each caller names itself in its own RC5017, and the typing declared three times (already drifted in the constructor signature) is declared once. Telemetry resolves through it, so a Node deployment with `better-sqlite3` installed gets the sink it configured instead of silently getting nothing, and its mutable public `static loadDriver` seam (clobberable by any concurrent test file) becomes an injectable argument. The Definition of Done's optional-peer rule moves out of the adapter checklist into its own section covering any module that loads a peer, with the `better-sqlite3` carve-out stated and the wording left compatible with #571.

**Events reference swept to a zero-diff (#600).** Beyond the three suspend/resume names the ticket asked for: `plugin:{pluginId}:registered` no longer exists and `applying`/`applied` were unlisted; `server:listening` was filed under the mcp plugin though the core server plugin emits it, so it relocates; the mcp note advertised `session:created`/`session:closed`, which have no emit site, and omitted `tool:declined`, which does; `plugin:http:*` and `agent:*` had no group at all. Four names were undocumented on the page the map indexes, so the reference now carries them.

## Inline, no ticket

Cherry-picked `1b1d8d9` from `claude/routecraft-release-0-6-0-dsnzc5`, review follow-ups on #564 that never landed. Both halves applied cleanly. The standards half was a factual correction: `.standards/ci-cd.md` claimed `routecraft.dev` and `examples` are both ignored, while `.changeset/config.json` has `routecraft.dev` in the `fixed` core train and only `examples` under `ignore`.

Two defects fixed in passing, both already on main and both in the direct path of this work:

- `loadOptionalPeer` hardcoded "adapter" in its RC5017 message, so non-adapter callers already produced `Markdown loader (agents() / skills()) adapter requires ...`. The option is now `consumer` and is used as written.
- The two agent registration events were emitted `as EventName`. Both names are declared in `EventDetailsMap`, so the cast was unnecessary and silently disabled payload checking on the only two emits that had it.

## Deviations

- **#595 was dropped:** already fixed and closed by #626 (`dbf4610`). Verified against the current tree. The mail pool's drain is instead exercised under the new forced stage, which reproduces the same still-connecting-entry shape.
- **#562 steps 3 and 4** had telemetry and the CLI keep a `query` member on the shared type. Neither ever calls it, so it is dropped rather than carried into an extraction whose whole purpose is removing unused drift. `transaction` is kept: the event writer uses it and both drivers expose it.
- The lockfile change is `bun install` reconciling a stale recorded peer range against package.json, not a dependency change.

## Not adopted, for the follow-up ledger

Nothing guards `events.json` against drifting again, which is the failure #600 exists to clean up. A guard needs the type checker to derive the emitted set reliably and would couple a core test to a docs-app data file, so it is a boundary question rather than a small inline fix. Reported, not taken.

## Verification

`bun run test`: 2738 pass, 0 fail, 189 files. `bun run lint`, `bun run typecheck`, `bun run build` clean. `bun run limit:size` green (core 115.24 kB of 200 kB).

The docs catalogue generator fails in my container (`operations.json resolved to zero rows on the next channel`), pre-existing on clean main because the clone lacks the channel data it reads. The new `server-events` and `http-plugin-events` anchors were verified against the established slug rule instead; CI exercises the generator properly.

One changeset covers the bundle, naming the three behaviour-visible changes and the two API changes explicitly.

https://claude.ai/code/session_01Q5oSiEXyMTDL1gRXr4aYJX

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q5oSiEXyMTDL1gRXr4aYJX)_

---
## Summary by cubic
Hardens lifecycle and shutdown: graceful shutdown now closes intake and drains in‑flight work instead of cancelling; only the forced stage abandons execution. The drain is bounded by `shutdown.timeoutMs` (invalid values are rejected as `RC5058`), and the shutdown outcome is both returned from `context.stop()` and emitted via `context:stopped`. Also unifies the SQLite driver resolver/types (enabling Node telemetry when `better-sqlite3` is installed), unwinds plugins on failed builds and adds `teardown(ctx, { partial, started })`, marks callable sources ready on invocation, and adds a boot summary.

- Rollout
  - Set `shutdown.timeoutMs` to a positive, finite ms below your platform kill timer; total shutdown can exceed it due to unbounded plugin teardown and server drain.
  - Where you used `route.signal` for waits/backoffs, switch to `route.intakeSignal`; keep `route.signal` for abandonment paths.
  - Plugin authors may read `teardown(ctx, { partial, started })`; existing `teardown(ctx)` continues to work.
  - Update tests: callable sources now announce readiness on invocation; shutdown tests should wait for step entry rather than sleep.
  - On Node, install `better-sqlite3` to enable SQLite telemetry and the suspension store; the CLI reads telemetry through the shared surface types.

<sup>Written for commit a6ada9a24e124fee7213a1afee22252f0cb42e71. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/routecraftjs/routecraft/pull/638?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ex0b1t:
**[LANE-DONE]** review round settled at `52bf64a`

**Fixed: 6**

1. **The suspension-handle test proved nothing** (`plugin-build-unwind.bun.test.ts`). It asserted that reopening the database file after a failed build showed the handle was closed, on the premise that an unclosed `bun:sqlite` handle locks the file. It does not; I probed it directly. The test now injects a driver through the existing `loaders` seam and asserts `close()` actually ran. *Class swept:* mutation-checked the whole unwind suite by removing the unwind from `build()`, which now fails 5 tests including this one, where before it failed 4 and this one still passed.
2. **The forced-shutdown agent test raced its own deadline**: 300ms of budget with a 100ms sleep before asserting the run had NOT been cancelled, so a loaded runner could report a cancellation stage one never performed. It now waits for the hang tool to be entered. *Class swept:* every test that slept to get work in flight before stopping now waits for the work instead, in both `agent-cancel` and `agent-suspend`.
3. **The exchange lifecycle guarantee** promised exactly one terminal event per started exchange, which this PR's forced shutdown breaks by design. The exception now sits with the guarantee, so a consumer reconciling starts against terminals reads an expected gap rather than a defect.
4. **`agent:registered` was documented twice** and the copy this PR added marked `description` optional; `agentPlugin` refuses to register an agent without a non-empty one (`plugin.ts:100`). Collapsed to one authoritative table, both payloads naming the `source` literal.
5. **Both lifecycle tables claimed routes are registered during `apply(ctx)`.** `ContextBuilder.build()` calls `initPlugins()` before `registerRoutes()`. Corrected in the docs page and the standard. *Class swept:* every source-level statement of the ordering was already correct.
6. **Two comments I added claimed the bulkhead "reports a clean rejection" at shutdown.** `ConcurrencyController.#joinWaitLine` admits the queued exchange with a no-op release so teardown runs it, as its own JSDoc always said. Corrected in `52bf64a`. *Class swept:* this error had propagated into my [LANE-Q] on #610 as the stated cost of my own recommendation, so I corrected that thread too rather than let a ruling rest on it.

**Dropped: 1**

- **Docstring coverage 70.59% against an 80% threshold** (CodeRabbit pre-merge check). It is the bot's own quality gate, not a repo gate, and writing docstrings to move a percentage is the comment noise `.standards/` and CLAUDE.md explicitly forbid. Every symbol this PR adds that carries contract or intent already has JSDoc; the gap is small internal helpers where a docstring would restate the line below it.

**Gate:** `bun run test` green at 2738 pass / 0 fail across 189 files. `bun run lint`, `bun run typecheck`, `bun run build` and `bun run limit:size` all clean (core 115.24 kB of 200 kB). CI green on `52bf64a`; `mergeable_state: clean`.

**Two things to know before the verification pass:**

- **CodeRabbit did not re-review the last two commits.** It hit its rate limit on both `5435763` and `52bf64a` and reports the next review available around 09:22Z. Its four inline findings from round one are marked addressed, but the walkthrough and merge-risk prose on the PR still describe round one's state and has not been regenerated. cubic has run on every head with no findings.
- **One [LANE-Q] is open** on [#610](https://github.com/routecraftjs/routecraft/issues/610#issuecomment-5379121191), narrowed by the correction above to the `.retry()` backoff alone: whether a backoff cut short at stage one (surfacing the last error) or drained to the deadline (abandoned with no terminal event) is right. Implemented as the former; a ruling either way is a contained change to one signal plus its test.

Commit range `985c71d..52bf64a`, 8 commits. Not merged by me.

---

ex0b1t:
@coderabbitai review

Requesting the re-review that hit your rate limit on `5435763` and `52bf64a`. Those two commits are the fixes for your five round-one findings plus the nitpick, so the round is not settled until they have been looked at. The substantive one is `packages/routecraft/test/plugin-build-unwind.bun.test.ts`, where the released-handle assertion was rewritten to inject a driver through the `loaders` seam rather than reopening the file.

CI is green on `52bf64a` and the [LANE-DONE] above records the round-one outcome.

---

ex0b1t:
**[LANE-DONE]** review round two settled at `a1a9e9c`

The re-review of `5435763` and `52bf64a` landed. Round one's five threads are resolved; two new findings, both in code I wrote in the round-one fixes, both real and both fixed.

**Fixed: 2**

1. **`waitForHang()` had no bound** (`packages/ai/test/agent-cancel.bun.test.ts`). It polled until the hang tool was entered, so a routing or startup regression that stopped the tool being reached would hang the suite rather than fail it. That is the same class as the flakiness the helper was written to fix, from the other side: a test that cannot fail is no better than one that fails at random. It now gives up after 5s with a diagnostic. *Class swept:* the only other polling loop across these suites (`agent-suspend`, waiting for the second model call) was already bounded, and the rest of my new tests wait on promises rather than polling.
2. **The deferred-fn note pointed at an event that carries no metadata.** I wrote that a `directTool`'s description and tags "surface on the tool-invocation events instead". `route:agent:tool:invoked` and its siblings carry `toolName`, `toolCallId` and a payload snapshot and nothing else; the metadata is resolved onto the tool at dispatch (`packages/ai/src/agent/tools/builders.ts:163,177`). Verified against both the payload declaration and the resolver before changing the text. The note now sends a reader to the resolved tool.

**Dropped: 1** — the docstring-coverage pre-merge warning, unchanged from round one: it is the bot's own gate rather than a repo gate, and writing docstrings to move a percentage is the comment noise `.standards/` and CLAUDE.md forbid.

**Gate:** `bun run test` green at 2738 pass / 0 fail across 189 files; `bun run lint` and `bun run typecheck` clean. CI was green on `52bf64a` across all twelve checks including `docs-site`, which exercises the docs catalogue generator I could not run locally, so the new `server-events` and `http-plugin-events` anchors resolve for real. CI on `a1a9e9c` is running; this round's diff is one test helper and one documentation paragraph.

Both round-two findings are worth naming for the verification pass, because both were self-inflicted in the act of fixing round one: a helper that removed a flake by introducing a hang, and a documentation correction that pointed somewhere else wrong. Neither touches production behaviour.

Commit range `985c71d..a1a9e9c`, 9 commits. Not merged by me.

---

ex0b1t:
CI is green on `a1a9e9c`, completing the one open line in the [LANE-DONE] above: all twelve checks pass (`validate`, `build`, `test`, `docs-site`, `adapter-cross-runtime` on both bun and node, and the smokes), and both round-two findings are marked addressed on their threads.

Ready for the verification pass.

---

ex0b1t:
**[LANE-DONE]** self-review round settled at `54d2abf`

Jaco asked for a full code review of the branch before merge. Seven lenses over the nine-commit diff, plus a blind contract interrogation. Nineteen findings, seventeen fixed, three dropped after adversarial verification. Everything below is a defect this branch introduced, not pre-existing debt.

**Fixed, behavioural (2)**

1. **The boot summary reported a failed route as started.** An async callable source that rejects on its first await (a refused connection, a bad credential, which is the shape the `from()` docs teach) signals readiness before the rejection is seen, and `settle()` refused to record the failure over the `route:started` that had already fired. A one-route context whose only route failed logged `info "1 of 1"`. A failure is now the final word. Verified by reproduction and mutation-checked: reverting the guard fails the new test. Readiness itself is deliberately left where it is, because a source that announces ready and then fails is how every adapter already behaves; the summary was the thing lying.
2. **`shutdown.timeoutMs` was unvalidated.** `0`, negatives, `NaN` and `Infinity` went straight to `setTimeout`, so a typo silently meant "force immediately" rather than "no bound". Now refused at construction with the new **`RC5058`**, registered across `error.ts`, `_data/errors.json` and the errors page. Refused rather than clamped: the two readings of `0` are opposites.

**Fixed, claims this branch made that the code did not support (4)**

3. The changeset, the `forceStageTwo` JSDoc and the configuration reference all said a forced shutdown abandons the sweeper's in-flight sweep. **Nothing does.** The sweeper stops cooperatively from `context:stopping`, which fires on every shutdown, and teardown then awaits it unbounded, outside the deadline. Wrong on two counts: the loss is not specific to the forced stage, and the process waits rather than abandoning. All three corrected.
4. The configuration reference told operators to set `timeoutMs` below the platform kill timer as though it bounded the whole shutdown. **It does not.** Teardown runs after it and a named server drains on its own `shutdownGraceMs`, also 30s, so a context with an HTTP mount can take 60 seconds. Following the advice as written could still earn a SIGKILL. Now stated.
5. The changeset said the shared resolver serves the CLI's telemetry reader. Only the shared **typing** does; the CLI keeps its own Bun-only load.
6. **`Route.signal` changed meaning while keeping its name** and the changeset never said so. Out-of-repo code reading it to notice a shutdown keeps compiling and silently stops reacting to a graceful one. Now an explicit API-change bullet, along with the three members `Route` gained.

**Fixed, contract violations (3)**

7. Both new log lines interpolated values into the message string, against `.standards/error-and-logging-policy.md` §3. The values were already in bindings, so the messages are now fixed strings.
8. That standard's terminal-event guarantee carried neither the forced-shutdown carve-out this branch created nor `suspended`.
9. `.standards/ci-cd.md` §6, which the Definition of Done cites as *the* optional-peer contract, still described `loadOptionalPeer` as adapter-only after this branch widened it.

**Fixed, smaller (8)**: the intake-signal JSDoc named `.retry()` as an execution-signal case while every retry call site here wires it to intake; `ShutdownOutcome`, `ShutdownConfig` and `DEFAULT_SHUTDOWN_TIMEOUT_MS` were unnameable by the consumer the CLI docs tell to read `{ forced, pending }`; `TeardownInfo.partial` is also true for a context built and stopped without starting; the published sqlite constructor typed a Bun-only `create` option; a stale JSDoc still called the build unwind unimplemented; stale AI1005 docs still listed "context shutdown" as a cancellation trigger; `serversplugin` still showed `teardown(ctx)`; two mechanical-edit artefacts.

**Dropped after verification (3)**

- **`route:stopped` fires before the drain finishes.** The premise was false: `main` also aborted then drained, so the events already fired early. This branch widens a pre-existing window rather than creating one, and the obvious fix would be actively wrong, since rewiring to the execution controller means the event never fires at all on a clean shutdown. Kept only the docs wording.
- **`raceShutdownDeadline` duplicates `raceWithDeadline`.** The verifier ran both shapes: my `unref` is load-bearing (a wedged drain exits at ~0ms with it, lingers 30s without), and the proposed fix would repurpose `DeadlineExceededError`, which callers map to RC5011.
- **A fifth polling helper in tests.** Real duplication, but across five pre-existing files in two packages, out of scope here.

**Also dropped, unchanged from earlier rounds:** CodeRabbit's docstring-coverage warning. Bot gate, not a repo gate, and writing docstrings to move a percentage is the comment noise `.standards/` forbids.

**Gate:** `bun run test` 2740 pass / 0 fail across 189 files (two new tests); lint, typecheck, build and `limit:size` clean. CI green on `54d2abf`, all twelve checks including `docs-site`, which exercises the catalogue generator and so validates the new `rc-5058` anchor. cubic neutral. CodeRabbit is rate-limited until roughly 10:25Z and has not yet re-reviewed this commit.

**One finding deliberately not fixed, for a ruling.** A forced shutdown is a failure state the event stream cannot observe: `context:stopped` carries `{}`, so alerting must read the exit code or grep a log line. `DEFINITION_OF_DONE.md` requires failure states to be observable through events. The fix is a `forced` field on `context:stopped`. I held off because Jaco and I settled that forced-ness rides on the return value rather than an event, and widening a public event payload is a decision for the lead rather than something to slip into a review. The honest alternative, if the payload stays as it is, is to say plainly in the configuration reference that counting forced shutdowns means reading exit codes by design.

Commit range `985c71d..54d2abf`, 10 commits. Not merged by me.

---

ex0b1t:
**[LANE-DONE]** CodeRabbit round three on `5a04ec4` processed. Three inline findings fixed in `f51937b`; the outside-diff Major is escalated below rather than fixed here.

### Fixed

**Fixed-delay synchronisation in the shutdown tests** (flagged on my new forced-arm test). Valid, and the class was wider than the one site flagged: all three shutdown tests dispatched an exchange, slept 20ms, then stopped the context. On a loaded runner the step may not have been entered yet, so the drain test could measure an exchange that never started and both forced tests could measure a clean shutdown while asserting a forced one. Each route now fires a deferred as its step is entered and the test awaits that before stopping. Swept all three, not just the flagged one.

**"30 second default" → "30-second default"** in the RC5058 suggestion and its errors-reference entry. Swept the class repo-wide: those two are the only true compound-adjective uses. The others ("defaults to 30 seconds", "60 seconds by default") are predicative and correctly unhyphenated, so they stay.

**The optional-peer standard's call direction.** It said the suspension store and telemetry sink "call it from `shared/sqlite/driver.ts`", which reads as though those two call `loadOptionalPeer` themselves. The driver calls it on their behalf. Reworded, plus a line saying a new SQLite consumer goes through the shared driver rather than calling the helper directly.

### [LANE-Q] Escalating the Major finding: plugin `start()` can outlive its own teardown

Half of it is real, and not the half as stated. I probed it rather than taking the claim at face value: a plugin whose `start()` is still awaiting when `stop()` runs produces the ordering `start:enter → teardown → start:resolve`.

So the finding's stated mechanism (teardown *skips* the plugin because completion is recorded too late) is **wrong**: teardown keys off `appliedPlugins`, `apply()` had already completed, and the plugin was torn down. What is **right** is the consequence: `start()` resolves after its own `teardown()` has run, so anything it acquires at the end of `start()` (a socket, an interval, a background task) is acquired after the release that was meant to cover it, and nothing ever releases it.

This bundle's `hasStopped` guards (`context.ts:747`, `context.ts:1175`) stop the *next* plugin's `start()` from being launched after a stop. They do nothing for the one already awaiting. The race predates this bundle; the guards are what make the remaining gap visible.

I have not fixed it, for a reason I want ruled on rather than deciding in a lane. The obvious small fix, having `stop()` await the in-flight start before teardown, risks a deadlock: `startInFlight` covers `run()`, which for an indefinite route does not resolve until the context stops. So a correct fix needs a promise scoped to the plugin lifecycle hooks alone, and it needs a semantic decision that is yours: **should `stop()` during boot wait for in-flight plugin hooks to finish, or should it interrupt them and accept that a plugin mid-`start()` gets torn down having partly acquired?** Waiting is safer for resources and unbounded in the wrong plugin; interrupting is bounded and demands every `start()` be written to tolerate teardown-before-completion.

Either way it is a change to lifecycle serialisation with its own tests, and it is a pre-existing defect rather than a regression this PR introduces. My recommendation is a follow-up issue against a fresh branch, not a widening of Bundle B. Say the word if you want it in scope here instead and I will take whichever semantics you pick.

Note that CodeRabbit's merge risk moved to Moderate purely on this finding, so its Moderate reflects a real defect whose stated mechanism is inaccurate and which is not new to this PR.

### Gate at `f51937b`

`bun run test` 2742 pass / 0 fail across 189 files; `lint`, `typecheck`, `build` and `limit:size` clean. The docs-site build ran to completion including prerender. CI green on `5a04ec4` before this push and re-running now.

Not acted on: the docstring-coverage pre-merge warning (70.59%, threshold 80%). It counts every function the diff touches, including test callbacks and one-line arrow helpers where a docstring would restate the signature. Our own rule is JSDoc on every test plus JSDoc where intent needs explaining, which the changed files satisfy.

One process note for @jaco: CodeRabbit's review again embeds agent-directed instruction blocks ("Prompt for AI Agents") in both the review body and each inline comment. Their content is benign this round, telling a reading agent to treat the findings as untrusted and verify them, which is what I did anyway. Flagging it because the channel exists, not because this round abused it.


---

ex0b1t:
@coderabbitai review

Requesting the re-review that hit your rate limit on `f51937b`. That commit is the fix for all three of your round-three inline findings:

- `packages/routecraft/test/shutdown-bounded.bun.test.ts`: the fixed-delay synchronisation is replaced by a deferred fired on step entry, applied to all three shutdown tests rather than only the one flagged, since the drain test had the same race from the other side.
- `packages/routecraft/src/error.ts` and the errors reference: `30-second default` hyphenated.
- `.standards/ci-cd.md`: the SQLite call direction corrected, with a line keeping future SQLite consumers on the shared driver.

Worth flagging for the re-review: your merge-risk assessment is still pinned to `5a04e` and reads Moderate on the outside-diff plugin-lifecycle finding. I probed that finding rather than taking it as stated, and the mechanism it describes does not hold: teardown keys off `appliedPlugins`, so a plugin whose `apply()` completed **is** torn down, not skipped. What does hold is the consequence, in a different order than described: the observed sequence is `start:enter → teardown → start:resolve`, so a plugin's `start()` resolves after its own teardown has run and anything it acquires late outlives shutdown. That is a real defect, it predates this PR, and it is escalated to the repo lead in the comment above with the semantic question it turns on.

CI is green on `f51937b` across all twelve checks and `mergeable_state` is clean.


---

ex0b1t:
**[LANE-DONE]** CodeRabbit round four on `f51937b` processed, fix in `a6ada9a`.

**No findings against round three's fixes.** The step-entry deferreds, the hyphenation and the SQLite call-direction wording all passed. One new outside-diff Minor, fixed.

**Fixed: `examples` versioning described the wrong mechanism** (`.standards/ci-cd.md`). The bullet above it lists packages changesets versions independently; the `examples` bullet then said it "is ignored and carries an independent version". `examples` sits under `ignore`, so changesets neither versions nor publishes it: its `1.0.0` is hand-set and has never moved with a release, and `craft.config.ts` imports it as the version its MCP server advertises. Two different mechanisms wearing the same phrase one bullet apart, in text this PR's cherry-pick touched. Now says which is which explicitly.

### The High merge risk is not supported by this round's review

Worth stating plainly, because it is a hold-the-merge signal and you are mid-verification.

The walkthrough now reads **🟠 High · up to `f5193`**, on the grounds that "asynchronous teardown hooks may still run after shutdown completes, risking leaked resources or post-shutdown work". The review submitted for `f51937b` contains exactly one finding: the `examples` documentation nit above. Nothing in it concerns teardown hooks.

The prose is a re-characterisation of the round-three outside-diff Major, and it has drifted in the retelling. Its three versions now read:

1. Round three's finding: teardown *skips* a plugin whose hook is in flight, because completion is recorded too late.
2. Round three's risk prose: "a stop can still overlap an in-progress plugin hook".
3. Round four's risk prose: "asynchronous **teardown** hooks may still run after shutdown completes".

(1) is disproved: teardown keys off `appliedPlugins`, and my probe shows the plugin **is** torn down. (3) is a different claim again, about teardown hooks rather than `start()`, and nothing in the diff or the probe supports it: teardown is awaited inside `performShutdown`, so a teardown hook cannot outlive the shutdown that awaits it.

What is real remains what I escalated at 11:03Z and it has not changed: the observed ordering is `start:enter → teardown → start:resolve`, so a plugin's `start()` resolves after its own teardown ran, and what it acquires late is never released. Pre-existing, not introduced here, and still awaiting your ruling on whether `stop()` should wait for in-flight plugin hooks or interrupt them.

So the High is pointing at something real through a description that does not match the code. I would not treat it as a merge block on its own terms, but the underlying defect is your call, and if you want it in scope for Bundle B I will take whichever semantics you pick.

### Gate at `a6ada9a`

Documentation-only diff, one line. `lint` and `typecheck` clean; the full suite was green at 2742 pass / 0 fail on `f51937b`, which this does not touch. CI green across all twelve checks on `f51937b` and re-running now. `mergeable_state: clean`.

Unchanged: the docstring-coverage warning stays dropped, for the reason given in earlier rounds.


---

ex0b1t:
**Lead ruling** on the 11:03Z escalation, and the state of the other held items:

**The plugin start-race: follow-up, out of Bundle B, with the semantics decided now so the ticket carries no open question.** Filed as **#640**, milestone 0.7.0 (introduced with #602's start phase this cycle and never released, so by the same standard as #635 it is fixed before the tag, in its own small PR after this merges). The ruled semantics: `stop()` waits for the in-flight `start()` hook on a promise scoped to the lifecycle hooks alone, unbounded. This is the same defect-class line the shutdown deadline ruling drew: a wedged `start()` is a defective plugin, not shutdown policy, and interrupting would tax every correct plugin author with a tolerate-teardown-mid-start contract to guard the defective case. Your deadlock analysis of the obvious fix is recorded in the ticket, as is your correction of the finding's stated mechanism.

**The forced-observability item is closed**: ruled on #590 before your escalation crossed with it, applied in `5a04ec4`, and verified independently below.

**The CodeRabbit High banner: concur, not a merge block.** I verified the load-bearing fact myself at this head: teardown is awaited inside `performShutdown`, so a teardown hook cannot outlive the shutdown awaiting it. The banner's prose is a third inconsistent retelling of a finding whose accurate form is #640. The docstring-coverage warning stays dropped: bot gate, not a repo gate.

**Lead verification, re-run at head `a6ada9a`** (supersedes the `800780d` verdict on #565): all twelve probes green, including two new ones for `5a04ec4`: `context:stopped` carries `{ forced: true, pending: ["hung"] }` on a forced stop and `{ forced: false, pending: [] }` on a clean one, matching the ruled stable shape. The earlier probes hold at this head: both unwind arms of #565 with the `{ partial, started }` semantics, the forced deadline honoured at ~400ms with the hung route named, the parked suspension surviving forced stage two untouched, and quiet-callable readiness with no backstop. Full gate re-running now; verdict lands when it exits.

---

ex0b1t:
**Lead verification verdict at `a6ada9a`: PASS.** Gate run by the lead: `bun run all` exit 0, 2742 pass / 1 skip / 0 fail; `bun run limit:size` green. Together with the twelve probes in the comment above, Bundle B is merge-ready from the lead's side. The one remaining tag-blocker it spun off, #640, is scoped, ruled, and lands separately after this merges. Jaco: the merge is yours.

---

Issue #590, opened 2026-08-11
feat(core): bounded graceful shutdown (shutdown.timeoutMs config, forced stage two, non-zero exit)


ex0b1t:
Approved by Jaco (2026-08-11) after lead review, from the Lane C close-out. Related: #131, epic #231. Unmilestoned: matters increasingly as suspension makes long-lived processes the norm, but nothing in 0.7.0 gates on it.

`shutdownHandler` implements two-stage shutdown, but nothing bounds stage one: a consumer that never settles, a hung in-flight HTTP request, or an agent mid-tool-call holds the process open indefinitely after SIGTERM. Under a container orchestrator the platform's kill timer then decides the outcome, turning a clean shutdown into a SIGKILL and losing whatever stage two was meant to do.

Shape settled in advance (lead ruling, confirmed by Jaco):

- **A config key, not a CLI flag**: `shutdown: { timeoutMs }` on `defineConfig`. Shutdown behaviour is a property of the app, not of one invocation.
- **Default ~30 seconds**, documented with the operational rule that matters: set it below your platform's kill timer (Kubernetes `terminationGracePeriodSeconds` and friends), so the process's own policy decides, not the platform's.
- **A forced stage two exits non-zero**, so an operator can distinguish a forced shutdown from a clean one in exit-code-sensitive tooling.

## Acceptance

- [ ] `shutdown.timeoutMs` config key with the documented default; stage two forced at the deadline.
- [ ] Clean shutdown within the bound exits 0; forced shutdown exits non-zero, with one log line naming what was still pending when the deadline hit.
- [ ] In-flight suspension writes are not corrupted by a forced stage two (probe, given the durable-release context).
- [ ] Configuration reference and deployment docs updated, including the below-platform-timer rule.
- [ ] Tests per `.standards/testing.md`; `bun run all` green.

ex0b1t:
**Lead ruling**: add `forced` and `pending` to the `context:stopped` payload, in this bundle.

[...]

Issue #632, opened 2026-08-19
feat(core,ai)!: resume authorization: per-call credentials, suspend-site answer policy, channel keys, optional expect


ex0b1t:
**Release-blocking for 0.7.0: the tag does not happen while this is open.** Follow-up to PR #630's adversarial verification, escalated as a [LANE-Q] there and ruled option A by the lead.

## The hole

`ctx.suspension` is minted per exchange sequence, not per tool call. Under the decision-10 batch rule (one park per batch; losing suspend signals become retryable tool errors), a model that emits two suspending tool calls in one batch gives BOTH handlers the same suspension id and token. The documented pattern is notify-out-of-band-then-suspend, so both approvers receive a live credential for different questions before the winner is chosen. Whichever answers first lands their answer as the WINNING call's tool result; re-entrant sites skip `expect` validation by design (intake decision 2), so nothing shape-checks the mismatch. The store's CAS bounds it to one wrong answer (the second link reads RC5050), and no attacker is required, only a model parallelising two approval tools.

Interim state shipped in #630: a loud warning on the durable-agents page (design so at most one tool per turn can park), a park-time AI1007 probe, and a losing-sibling warn. Those hold the line; they do not close the hole.

## The fix (option A from the escalation)

- The resume token payload gains a `sub` claim carrying the suspending `toolCallId`. The tool bridge mints **per-call** credentials, so each handler in a batch holds a token bound to its own call.
- Re-entrant revival verifies `sub` against `stepState.suspendedToolCallId` **before the store claim**, per the validation-before-CAS precedent: a losing call's token gets a clean typed refusal, never a burned claim and never a mis-delivered answer. The approver behind a refused link has lost nothing; when the resumed model retries that tool, a fresh park issues them a new, correctly-bound credential.
- The refusal is a NEW code, not RC5050: "this credential is bound to a question that is no longer pending" is a distinguishable state from "this suspension already settled," and the re-ask machinery must not amplify it.
- Static (non-re-entrant) sites have exactly one logical asker, so `sub` is absent or trivially satisfied there; no behaviour change outside the re-entrant arm.

## Why this is safe to change now

The token format merged in #563 but has never been in a tagged release; 0.7.0 is where it first ships. Changing the payload today has zero migration cost, which is exactly why this blocks the tag: after 0.7.0 it becomes a breaking format change.

## Acceptance

[...]

Issue #565, opened 2026-08-10
fix(core): ContextBuilder.build() does not unwind applied plugins when a later step fails


ex0b1t:
## The failure shape

`ContextBuilder.build()` applies plugins and then registers routes:

```ts
// packages/routecraft/src/builder.ts
await ctx.initPlugins();
ctx.registerRoutes(...this.definitions);
return { context: ctx, client: new CraftClient(ctx) };
```

`CraftContext.initPlugins()` (`packages/routecraft/src/context.ts`) awaits each plugin's `apply()` in order. On a throw it logs, emits `context:error`, and rethrows. Nothing tears down the plugins that already applied successfully.

So when plugin N throws, plugins 1..N-1 have run their `apply()` and stay applied. Anything they acquired in `apply()` is now unreachable: `build()` never returns a context, so the caller has no handle to run `teardown()` against, and no teardown runs on the failure path anyway. The same hole covers a failure in `registerRoutes()`, which runs after every plugin has applied.

A plugin that only mutates the context (registering config, adding handlers) loses nothing here. A plugin that acquires a **resource** in `apply()` leaks it: an open database connection, a file handle, a socket, a `setInterval` timer that keeps the event loop alive.

## The concrete case that surfaced it

On the #563 branch (`claude/549-suspension-store`), `suspensionPlugin` opens a SQLite database in `apply()` and closes it in `teardown()`:

```ts
async apply(ctx) {
  ctx.setStore(SUSPENSION_RUNTIME, await createSuspensionRuntime(ctx, config));
},
async teardown(ctx) {
  const runtime = ctx.getStore(SUSPENSION_RUNTIME);
  if (runtime?.ownsStore) await runtime.store.close();
}
```

If a later plugin or route registration throws, that handle is held by a `SqliteSuspensionStore` no one can reach, and the database file stays locked. It was found while reviewing that PR.

**A bespoke unwind in `suspensionPlugin` was deliberately declined.** Catching in that one `apply()` and closing its own store would fix that one plugin and leave the hole open for every other plugin holding a connection, a file handle, or a timer, including ones not yet written. The exposure is a property of the builder's lifecycle, not of the suspension plugin, so the fix belongs in the builder. Slice 1 of #417 shipped without it on that basis.

## The operational bite

[...]

PR #642, opened 2026-08-22
fix(core): wait for an in-flight plugin hook before teardown

### LINKED ISSUE #640: Plugin start() can outlive its own teardown when stop() races boot
Escalated from PR #638 (CodeRabbit round three's outside-diff Major, corrected by the lane's probe); lead-ruled into its own ticket. Introduced with the `start()` lifecycle phase this cycle (#602), never released, so it is fixed before the 0.7.0 tag, in its own small PR after Bundle B merges.

## Problem

A `stop()` that arrives while a plugin's `start()` is still awaiting produces the ordering `start:enter, teardown, start:resolve`. Teardown itself is NOT skipped (it keys off `appliedPlugins`, which the reviewing bot got wrong; the lane's probe result is the accurate record), but anything the hook acquires after its last await point is acquired after the release that was meant to cover it, and nothing ever releases it. The bundle's `hasStopped` guards stop the NEXT hook from launching after a stop; they do nothing for the one already in flight.

Impact is bounded for a process that exits (the OS reclaims), and real for embedders and tests that build successive contexts in one process: a leaked interval or socket from a torn-down context lives on.

## Ruled semantics

`stop()` WAITS for the in-flight `start()` hook before running teardown, on a promise scoped to the plugin lifecycle hooks alone (never `run()`, which for an indefinite route resolves only at stop and would deadlock, as the escalation correctly identified). Unbounded, deliberately: this is the same defect-class boundary the shutdown deadline ruling drew for teardown. A wedged `start()` hook is a defective plugin, not a shutdown-policy problem, and interrupting instead would demand every plugin author write `start()` to tolerate teardown-before-completion, which is a heavy contract bought to protect against the defective case only.

## Acceptance

- `stop()` during boot runs no plugin's `teardown()` while that plugin's `start()` is still in flight; the observed ordering per plugin is always enter, resolve, teardown.
- The waiting promise covers lifecycle hooks only; a `stop()` against a context with indefinite routes and healthy plugins is not delayed by it.
- Tests: a slow `start()` racing `stop()` (ordering pinned per plugin), and a `stop()` with no in-flight hook proving no added latency.

ex0b1t:
Closes #640. The last tag-blocker from Bundle B (#638), landing as its own small PR as ruled.

## The defect

A `stop()` that arrived while a plugin's `apply()` or `start()` was awaiting produced the ordering `start:enter`, `teardown`, `start:resolve`.

Teardown itself was never skipped. It keys off `appliedPlugins`, so the plugin was torn down, which is where CodeRabbit's original description of this had the mechanism wrong. What is real is the consequence: anything the hook acquired after its last await point was acquired after the release meant to cover it, and nothing released it afterwards. #638's `hasStopped` guards stop the NEXT hook from launching after a stop; they do nothing for the one already in flight, which is what made the remaining gap visible.

Introduced with the `start()` phase in #602 and never released, so it is fixed before the 0.7.0 tag. Impact is bounded for a process that exits, since the OS reclaims, and real for embedders and tests that build successive contexts in one process: a leaked interval or socket from a torn-down context lives on.

## The fix, as ruled

`stop()` waits for the in-flight hook before teardown, on a promise scoped to the plugin lifecycle hooks alone. It never covers `run()`, which for an indefinite route resolves only once the context stops, so waiting on that would deadlock the shutdown this ordering exists to serve.

Unbounded, deliberately, on the same defect-class line #590 drew for teardown. A hook cut short keeps whatever it acquired, and interrupting instead would oblige every plugin author to write `start()` so it tolerates teardown-before-completion: a heavy contract paid for by every correct plugin to contain the defective one. A hook that never settles is a defective plugin, not a shutdown-policy question.

The published promise covers the hook **and** the bookkeeping that records it, so a shutdown waiting on it sees a plugin either fully applied or fully absent, never half-recorded.

## Beyond the acceptance criteria

`initPlugins()` now re-checks for a stop before each plugin, as `startPlugins()` already did. Without it, tracking the `apply()` hook is half a mechanism: shutdown would wait for the applying plugin, tear down, and `initPlugins()` would carry on applying plugins that the teardown walk has already passed, acquiring what nothing will release. Same defect class, same shape of guard, flagged here because it is one step past what the ticket asked for.

## Verification

Three tests: the `start()` race with the ordering pinned per plugin, the same for `apply()` with the following plugin proven never to apply, and a `stop()` with no hook in flight against a context with indefinite routes, which is the deadlock the ruling warned about.

The first fixture I wrote passed with the fix removed and therefore pinned nothing: it released the gated hook immediately after `stop()`, so the hook resolved during the route drain and the ordering held either way. Rebuilt on a hand-constructed context with no routes, so a shutdown that does not wait reaches teardown immediately and the hook's own duration is the whole margin. Mutation-checked: removing `await this.settlePluginHook()` fails both race tests, restoring it passes all three.

`bun run test` 2745 pass / 0 fail across 190 files. `lint`, `typecheck`, `format`, `build` clean. `bun run limit:size` green (core 115.52 kB of 200 kB).

The ordering guarantee is documented in `.standards/plugin-lifecycle.md` §3 and on the plugins reference page, with a patch changeset.

https://claude.ai/code/session_01Q5oSiEXyMTDL1gRXr4aYJX

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q5oSiEXyMTDL1gRXr4aYJX)_

---
## Summary by cubic
Waits for any in-flight plugin apply()/start() before teardown to prevent teardown-before-settle leaks. Previously a stop during boot could order enter → teardown → settle; now it is enter → settle → teardown, and start() halts at plugins and awaits shutdown.

- stop() awaits one in-flight lifecycle hook; the wait is unbounded, covers apply/start only, and never waits on run().
- Both lifecycle walks re-check hasStopped per plugin and again after emitting plugin:applying/starting so synchronous handlers that call stop() prevent the hook from starting.
- start() exits early when a stop lands mid-apply/start and awaits the shutdown so it does not report progress or resolve while teardown runs.
- Adds runLifecycleHook()/settlePluginHook() so teardown sees a hook fully settled and its bookkeeping applied; no half-recorded state.
- Documents: a lifecycle hook must not await its own ctx.stop(); throw to abort boot or call ctx.stop() without awaiting. Tests cover apply()/start() races, synchronous handler stops, and the unawaited-stop path; a new test asserts the ordering before the outcome. Publishes a patch changeset for `@routecraft/routecraft`.

**Migration**
- Do not await ctx.stop() inside apply()/start(); either throw or call ctx.stop() without awaiting. No other changes required.

<sup>Written for commit bcd0a1d17008ce1701d0cee94450e10b76dac77a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/routecraftjs/routecraft/pull/642?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/plugin-hook-stop-race.md
+++ b/.changeset/plugin-hook-stop-race.md
@@ -24,3 +24,12 @@ teardown-before-completion.
 Both lifecycle walks also re-check for a stop before each plugin, so a stop
 mid-boot no longer applies or starts plugins that teardown has already
 walked past.
+
+**One constraint comes with the wait, documented rather than enforced.** A
+lifecycle hook must not `await` its own `ctx.stop()`: it would wait for a
+shutdown that is waiting for the hook. Both intents keep a working spelling,
+and they differ by a keyword. Use `throw` to abort the boot with a reason,
+which unwinds through the teardown walk and surfaces the error, or call
+`ctx.stop()` without awaiting it to stand the context down without failing
+the boot. Nothing detects the awaited form, so it hangs at boot on the first
+run; the unawaited form is pinned by a test.
--- a/.changeset/plugin-hook-stop-race.md
+++ b/.changeset/plugin-hook-stop-race.md
@@ -1,0 +1,26 @@
+---
+"@routecraft/routecraft": patch
+---
+
+A `stop()` racing boot no longer tears a plugin down while its own lifecycle
+hook is still running.
+
+A stop that arrived while a plugin's `apply()` or `start()` was awaiting
+produced the order enter, teardown, resolve: the plugin was torn down (the
+teardown walk keys off the applied set, so it was never skipped), but
+anything the hook acquired after its last await point was acquired after the
+release meant to cover it, and nothing released it afterwards. A process that
+exits hides this because the OS reclaims; an embedder or a test suite
+building successive contexts in one process keeps the interval or socket.
+
+Shutdown now waits for the in-flight hook before teardown, on a promise
+scoped to the plugin lifecycle hooks alone. It deliberately does not cover
+`run()`, which for an indefinite route resolves only once the context stops.
+The wait is unbounded, matching plugin teardown: a hook that never settles is
+a defective plugin rather than a shutdown-policy question, and interrupting
+instead would require every plugin author to write `start()` so it tolerates
+teardown-before-completion.
+
+Both lifecycle walks also re-check for a stop before each plugin, so a stop
+mid-boot no longer applies or starts plugins that teardown has already
+walked past.
--- a/.standards/plugin-lifecycle.md
+++ b/.standards/plugin-lifecycle.md
@@ -87,6 +87,34 @@ or task to carry it.
   per plugin, so a `stop()` mid-boot stops the walk rather than applying or
   starting plugins nothing will release.
 
+### A hook must not await its own `stop()`
+
+A lifecycle hook may shut the context down. What it must not do is **await**
+that shutdown from inside the hook: the hook then waits for a shutdown that
+is waiting for the hook, and neither ever settles. The wait above is what
+makes this circular, and nothing detects it, so the failure is a silent hang
+at boot rather than an error.
+
+Both intents already have a working spelling, and they mean different things:
+
+| Intent | Spelling | What happens |
+|--------|----------|--------------|
+| Abort the boot with a reason | `throw` | `build()` and `start()` unwind through the teardown walk and the error reaches the operator unchanged |
+| Request shutdown without failing the boot | call `ctx.stop()` and do not await it | the hook settles, the wait sees it settle, and teardown follows in its proper order |
+
+```ts
+async start(ctx) {
+  if (!healthy) throw rcError("RC9901", undefined, { message: "..." });
+  // or, to stand the context down without failing the boot:
+  void ctx.stop();
+}
+```
+
+The constraint is one keyword wide and it is documented rather than enforced.
+Detecting it would need the shutdown to know which async context asked for
+it, which costs more machinery in the lifecycle path than the rule it would
+replace, for a misuse that fails deterministically on the first boot.
+
 ### The build-failure unwind
 
 A failure inside `build()` has the same hole with none of the same escape
--- a/.standards/plugin-lifecycle.md
+++ b/.standards/plugin-lifecycle.md
@@ -74,6 +74,18 @@ or task to carry it.
 - A teardown that throws during the unwind is logged and does not replace
   the start error. The operator needs the cause of the failed boot, not
   whatever the cleanup hit on the way out.
+- A `stop()` that arrives while a lifecycle hook is still awaiting WAITS
+  for that hook before teardown runs, so the order a plugin observes is
+  always `apply`/`start` entered, resolved, then `teardown`. The wait covers
+  the lifecycle hooks alone and never `run()`, which for an indefinite route
+  resolves only at shutdown. It is unbounded for the same reason teardown
+  is: a hook cut short keeps whatever it acquired past its last await point,
+  and interrupting instead would oblige every plugin author to write
+  `start()` so it tolerates teardown-before-completion. A hook that never
+  settles is a defective plugin, not a shutdown-policy question.
+- No hook runs once teardown has walked the applied set. Both walks re-check
+  per plugin, so a `stop()` mid-boot stops the walk rather than applying or
+  starting plugins nothing will release.
 
 ### The build-failure unwind
 
--- a/.standards/plugin-lifecycle.md
+++ b/.standards/plugin-lifecycle.md
@@ -76,7 +76,7 @@ or task to carry it.
   whatever the cleanup hit on the way out.
 - A `stop()` that arrives while a lifecycle hook is still awaiting WAITS
   for that hook before teardown runs, so the order a plugin observes is
-  always `apply`/`start` entered, resolved, then `teardown`. The wait covers
+  always `apply`/`start` entered, settled, then `teardown`. The wait covers
   the lifecycle hooks alone and never `run()`, which for an indefinite route
   resolves only at shutdown. It is unbounded for the same reason teardown
   is: a hook cut short keeps whatever it acquired past its last await point,
--- a/apps/routecraft.dev/app/content/docs/reference/plugins/index.mdx
+++ b/apps/routecraft.dev/app/content/docs/reference/plugins/index.mdx
@@ -92,6 +92,8 @@ Hooks run in registration order and each is awaited before the next plugin's, at
 
 A failure during `build()` unwinds the same way, so a plugin that opened a resource before a later plugin refused does not leak it: `build()` returns no context, so nothing else could ever release it. Only plugins whose `apply()` returned are torn down, and the original error surfaces unchanged even if a teardown throws on the way out.
 
+A `stop()` that lands mid-boot waits for the hook that is still running before it tears that plugin down, so the order a plugin observes is always its hook entered, its hook resolved, then `teardown`. Anything a hook acquires after its last `await` is therefore covered by the release that follows it. The wait is unbounded and covers `apply` and `start` alone, never the route work that only ends at shutdown, and no further hook runs once teardown has begun.
+
 `teardown` receives a second argument describing how far the context got, so a plugin never has to infer it:
 
 | Field | Means |
--- a/apps/routecraft.dev/app/content/docs/reference/plugins/index.mdx
+++ b/apps/routecraft.dev/app/content/docs/reference/plugins/index.mdx
@@ -94,6 +94,17 @@ A failure during `build()` unwinds the same way, so a plugin that opened a resou
 
 A `stop()` that lands mid-boot waits for the hook that is still running before it tears that plugin down, so the order a plugin observes is always its hook entered, its hook settled, then `teardown`. Anything a hook acquires after its last `await` is therefore covered by the release that follows it. The wait is unbounded and covers `apply` and `start` alone, never the route work that only ends at shutdown, and no further hook runs once teardown has begun.
 
+A hook may shut the context down, but it must never `await` its own `stop()`: the hook would wait for a shutdown that is waiting for the hook, and neither settles. Use `throw` to abort the boot with a reason, which unwinds through the teardown walk and surfaces the error, or call `ctx.stop()` without awaiting it to stand the context down without failing the boot. The rule is documented rather than enforced, and awaiting it hangs at boot on the first run.
+
+```ts
+async start(ctx: CraftContext) {
+  // Abort the boot, with the reason reaching the operator:
+  if (!configured) throw new Error("mail credentials missing");
+  // Or request shutdown without failing the boot:
+  void ctx.stop();
+}
+```
+
 `teardown` receives a second argument describing how far the context got, so a plugin never has to infer it:
 
 | Field | Means |
--- a/apps/routecraft.dev/app/content/docs/reference/plugins/index.mdx
+++ b/apps/routecraft.dev/app/content/docs/reference/plugins/index.mdx
@@ -92,7 +92,7 @@ Hooks run in registration order and each is awaited before the next plugin's, at
 
 A failure during `build()` unwinds the same way, so a plugin that opened a resource before a later plugin refused does not leak it: `build()` returns no context, so nothing else could ever release it. Only plugins whose `apply()` returned are torn down, and the original error surfaces unchanged even if a teardown throws on the way out.
 
-A `stop()` that lands mid-boot waits for the hook that is still running before it tears that plugin down, so the order a plugin observes is always its hook entered, its hook resolved, then `teardown`. Anything a hook acquires after its last `await` is therefore covered by the release that follows it. The wait is unbounded and covers `apply` and `start` alone, never the route work that only ends at shutdown, and no further hook runs once teardown has begun.
+A `stop()` that lands mid-boot waits for the hook that is still running before it tears that plugin down, so the order a plugin observes is always its hook entered, its hook settled, then `teardown`. Anything a hook acquires after its last `await` is therefore covered by the release that follows it. The wait is unbounded and covers `apply` and `start` alone, never the route work that only ends at shutdown, and no further hook runs once teardown has begun.
 
 `teardown` receives a second argument describing how far the context got, so a plugin never has to infer it:
 
--- a/packages/routecraft/src/context.ts
+++ b/packages/routecraft/src/context.ts
@@ -1187,6 +1187,20 @@ export class CraftContext {
       started.reject(err);
       throw err;
     }
+
+    // A stop() that landed while the plugins were applying has already torn
+    // them down. Starting routes now announces boot progress for a context
+    // that is gone: route:starting fires after context:stopped, every route
+    // refuses with RC3001 against its aborted controller, and the boot
+    // summary reports a failed boot for what was a clean shutdown. Waiting
+    // for that shutdown before returning also keeps start() from resolving
+    // while teardown is still running.
+    if (this.hasStopped) {
+      // The shutdown's own failure belongs to whoever called stop().
+      await this.shutdownPromise?.catch(() => undefined);
+      return;
+    }
+
     this.logger.info(
       { routeCount: this.routes.length },
       "Starting Routecraft context",
--- a/packages/routecraft/src/context.ts
+++ b/packages/routecraft/src/context.ts
@@ -388,6 +388,16 @@ export class CraftContext {
   /** The in-flight start, so concurrent `start()` calls join one boot. */
   private startInFlight: Promise<void> | undefined;
 
+  /**
+   * The plugin lifecycle hook currently awaiting, so a `stop()` that lands
+   * mid-boot can let it finish before teardown runs.
+   *
+   * Scoped to `apply()` and `start()` alone, never `run()`: for an
+   * indefinite route `run()` resolves only once the context stops, so
+   * waiting on it would deadlock the shutdown this ordering exists to serve.
+   */
+  private pluginHookInFlight: Promise<void> | undefined;
+
   /** Latched by `stop()`. A stopped context refuses to start again. */
   private hasStopped = false;
 
@@ -515,6 +525,10 @@ export class CraftContext {
     if (this.pluginsInitialized) return;
     this.pluginsInitialized = true;
     for (const [pluginIndex, plugin] of this.plugins.entries()) {
+      // Same guard as startPlugins(), for the same reason: once teardown has
+      // walked the applied set, a plugin applied after it acquires resources
+      // nothing will ever release.
+      if (this.hasStopped) return;
       try {
         if (
           !plugin ||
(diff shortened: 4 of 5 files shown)


ex0b1t:
**[LANE-DONE]** CodeRabbit round one on `8b5e9f8` processed, fixed in `8834d16`. Both findings real, both fixed.

### Fixed: the Major is correct, and it is a regression this PR introduced

**A lifecycle event subscriber can stop the context between the walk's guard and the hook it guards.** I probed it rather than taking it as stated, and this time the finding holds exactly as written: `EventBus.emit()` dispatches synchronously, so a `plugin:starting` handler calling `stop()` runs to completion inside the same statement sequence as the `hasStopped` check above it. The observed order before the fix was `stop`, `start:enter`, `teardown`.

What makes it worse than the pre-existing behaviour is this PR itself. On `main` that late hook was merely orphaned: teardown ran without it and shutdown finished. With the wait in place, shutdown now waits for a hook that should never have started, so a hook that never settles turns an orphan into a hang. That is the failure mode the unbounded wait was ruled acceptable for in the defective-plugin case, arriving here through a path that is not the plugin's fault.

Both walks now re-check after their emit. Two tests pin it, one per phase, and they are mutation-checked: removing the guards fails both.

**Minor, and correct: a rejecting hook settles, it does not resolve.** `apply()` and `start()` can reject, and shutdown waits for the rejection just the same, which is why `settlePluginHook()` observes settling rather than success. The guarantee in `.standards/plugin-lifecycle.md` §3 and on the plugins reference said "resolved". Both now say "settled", matching what the code does and what the JSDoc already said.

### On the Moderate merge risk

It reads as supported this round, unlike the drifted High on #638. Its first clause is the Major above, now fixed. Its second, that a non-settling hook keeps teardown pending, is the ruled design rather than a defect: unbounded is deliberate, on the same defect-class line #590 drew for teardown, and #640 records the reasoning. With the post-emit guard in place, the only way to reach that state is a plugin whose own hook never settles.

### Gate at `8834d16`

`bun run test` 2747 pass / 0 fail across 190 files, two more than `8b5e9f8`. `lint`, `typecheck`, `format` and `build` clean, `limit:size` green (core 115.58 kB of 200 kB). CI was still mid-run on `8b5e9f8` when this superseded it.

Not merged by me.

---

ex0b1t:
@coderabbitai review

Requesting the re-review that hit your rate limit on `8834d16`. That commit fixes both of your round-one findings:

- `packages/routecraft/src/context.ts`: both lifecycle walks re-check `hasStopped` after their `plugin:applying` / `plugin:starting` emit, closing the window your Major identified. I probed it against the real code first rather than taking it as stated, and it reproduced exactly as you described: the observed order was `stop`, `start:enter`, `teardown`. Two tests pin it, one per phase, and they are mutation-checked.
- `.standards/plugin-lifecycle.md` and the plugins reference: both guarantees now say a hook **settles** rather than resolves, since `apply()` and `start()` can reject and shutdown waits for the rejection just the same.

Worth flagging for the re-review: your walkthrough and the Moderate merge risk are still pinned to `8b5e9`, so they describe the defect that `8834d16` fixed. The second half of that risk prose, a non-settling hook keeping teardown pending, is the ruled design rather than a defect: the wait is deliberately unbounded, on the same defect-class line #590 drew for plugin teardown, and #640 carries the reasoning. With the post-emit guard in place, only a plugin whose own hook never settles can reach that state.

CI is green on `8834d16` across all twelve checks and `mergeable_state` is clean. cubic reviewed the same head with no findings.

---

ex0b1t:
**[LANE-DONE]** CodeRabbit round two on `8834d16` processed, fixed in `1fda54c`. One finding, half of it real, and the real half was worth taking.

### Fixed: the boot carried on past a stop that landed mid-apply

`run()` continued into route startup after `initPlugins()` returned early on a stop. I probed it with a route registered and a `plugin:applying` handler calling `stop()`. What actually happens:

- `route:starting` fires **after** `context:stopped`.
- Every route refuses with **RC3001** against its already-aborted controller, logged at **fatal**.
- The boot summary then warns `0 of 1 started`, reporting a failed boot for what was a clean shutdown.

So a context stopped mid-boot announced boot progress it was not making and logged a fatal for an expected outcome. `run()` now stops where the plugins stopped, and awaits the shutdown before returning so `start()` cannot resolve while teardown is still running. The applying-event test now registers a route and asserts no `route:starting`, and it is mutation-checked: removing the bail fails it on the RC3001 path.

### Correcting the mechanism, since the merge risk rests on it

The finding says shutdown can finish "while a route starts after its controller was aborted", and the risk prose has it "potentially outlive teardown". That part does not hold. `route.start()` calls `assertNotAborted` and throws RC3001, so the route never starts: my probe recorded the source's `subscribe()` never running, with the events ending at `stop`, `route:starting`, `context:stopped`. Nothing survives teardown, because nothing was ever subscribed.

What was real is narrower and still worth fixing: an event stream and a fatal log that misrepresent a clean shutdown as a failed boot. That is what `1fda54c` addresses.

Also worth noting that this window is not new to this PR. On `main` a stop during `initPlugins()` produced the same RC3001 fatal and the same misleading summary, because nothing checked between the plugin walk and route startup. The guard I added for #640 made it easy to see, not easy to hit.

### Gate at `1fda54c`

`bun run test` 2747 pass / 0 fail across 190 files. `lint`, `typecheck`, `format` and `build` clean, `limit:size` green (core 115.53 kB of 200 kB). CI was green on `8834d16` across all twelve checks and is re-running now.

Not merged by me.

---

ex0b1t:
@coderabbitai review

The automatic review of `1fda54c` hit your rate limit, so the walkthrough and the Moderate merge risk are still pinned to `8834d` and describe a defect that commit fixed. This is my last request on this PR either way.

`1fda54c` addresses your round-two finding: `run()` now stops at the plugins when a stop lands mid-apply, instead of carrying on into route startup, and awaits the shutdown so `start()` cannot resolve while teardown is still running. The applying-event test registers a route and asserts no `route:starting`.

One correction worth carrying into the re-review, since the merge risk rests on it. Your finding said a route could start against an aborted controller and "potentially outlive teardown". It cannot: `route.start()` calls `assertNotAborted` and throws RC3001 before any subscription happens. I probed it, and the source's `subscribe()` never runs, with events ending at `stop`, `route:starting`, `context:stopped`. What was real, and what `1fda54c` fixes, is narrower: a `route:starting` after `context:stopped`, a fatal-level log, and a boot summary reporting a failed boot for what was a clean shutdown.

CI is green on `1fda54c` across all twelve checks, `mergeable_state` is clean, all three review threads are resolved, and cubic reviewed the same head with no findings.

---

ex0b1t:
**[LANE-Q]** A lifecycle hook that awaits its own `stop()` now deadlocks, and it did not before this PR

CodeRabbit's round-three finding is **real**, and it is a regression this PR introduces rather than a pre-existing defect. I probed both sides before writing this:

- **With `settlePluginHook()`**: a `start()` hook doing `await ctx.stop()` produces `["start:enter"]` and never advances. The hook awaits `shutdownPromise`; `performShutdown()` awaits that same hook. Neither settles.
- **With the wait removed** (pre-PR behaviour): `["start:enter", "teardown", "start:resolve"]`. Wrongly ordered, which is exactly what #640 exists to fix, but it completes.

So the fix converts working-if-misordered code into a hang. That is a different thing from the wedged-hook case the ruling accepted, where a defective plugin hangs its own shutdown: here the plugin is doing something reasonable and the framework eats it.

### Why this is not a small fix

Hooks run strictly sequentially, so while hook N is in flight there are exactly two origins for a `stop()` call:

1. **From inside hook N itself.** Waiting for N is the deadlock.
2. **From anywhere else** (a signal handler, an embedder, a test) while N happens to be awaiting. Waiting for N is the entire point of #640.

Both leave identical state on the context: `hasStopped` set, `pluginHookInFlight` pointing at N. Nothing distinguishes them without knowing which async context the call came from, so `settlePluginHook()` cannot tell "the hook asked for this" from "someone else asked while the hook was busy". Skipping the wait when a stop arrives during a hook would void the fix for case 2, which is the common one.

### Options

- **(a) Detect re-entrancy with `AsyncLocalStorage`.** Run each hook inside a store carrying its token; `stop()` reads the store, and a shutdown requested from within hook N does not wait for N. This is the only option that keeps both guarantees intact. Cost: async-context machinery in the core lifecycle path, and `.standards/exchange-state-model.md` currently records a deliberate no-ambient-context position (for exchange state, on different grounds, but the grain is there).
- **(b) Prohibit it and document it.** The sanctioned way to abort a boot from a hook is already to **throw**: `build()` and `start()` both unwind through the teardown walk, which is what #565 shipped. `await ctx.stop()` from inside a hook is not that path. We say so in `.standards/plugin-lifecycle.md` and the plugins reference, and accept that a plugin ignoring it hangs. Cost: an unenforced rule whose violation is a silent hang rather than an error.
- **(c) Bound the settle wait.** Rejected on its face: it reintroduces the interrupt semantics you ruled against on #640, for one corner case.

### Recommendation: (b), with (a) as a separate ticket if you want it enforced

Throwing is already the documented way for a hook to refuse a boot, so the prohibition is not inventing a constraint, it is naming one that the lifecycle already implies. That keeps this PR at its ruled scope. My hesitation is honest: (b) trades a deadlock for a rule, and rules do not run.

**Blocks the merge in my view**, since it is a regression against `main` in code a plugin author could plausibly write, and you are the one who decides whether a documented prohibition is enough. Say which and I will implement it, gate it, and post the round.

Everything else on this PR is settled: CI green on `1fda54c` across all twelve checks, `mergeable_state` clean, all three earlier threads resolved, cubic clean. CodeRabbit also confirmed the `run()` guard and accepted the correction that a route cannot outlive teardown.

---

ex0b1t:
**Lead ruling** on the deadlock [LANE-Q]: **(b), prohibit and document**, with two additions the probe results below make possible. This does not block the merge once they are applied.

I probed the finding and its alternatives at `1fda54c` rather than ruling on the analysis alone, because a "prohibit and document" ruling is only honest if the paths it points at actually work:

| Probe | Result |
|---|---|
| `await ctx.stop()` inside `start()` | **Deadlocks.** `["start:enter"]`, never advances. Finding confirmed exactly as reported. |
| `throw` inside `start()` | **Clean.** `["start:enter", "teardown"]`, `start()` rejects. |
| **`void ctx.stop()`** (unawaited) inside `start()` | **Clean and correctly ordered.** `["start:enter", "start:resolve", "teardown"]`. |
| Outside `stop()` while a hook is in flight (control) | **Still waits.** `["start:enter", "start:resolve", "teardown"]`, the #640 guarantee intact. |

**The third row is the one that settles this.** The prohibition is not "a plugin may not shut the context down from a hook", which would be a real capability loss dressed up as a rule. It is narrower: *do not await it*. Both intents already have a working expression, and they mean different things:

- **Abort the boot with a reason**: `throw`. `build()` and `start()` unwind through the teardown walk, which is what #565 shipped, and the error reaches the operator.
- **Request shutdown without failing the boot**: call `ctx.stop()` and do not await it. The hook settles, the wait sees it settle, teardown follows in the right order.

So the trade is not "a deadlock for a rule". It is one broken spelling against two correct ones, where the correct spelling differs by a keyword. That is a documentable constraint rather than a hole.

**Conditions, both required before merge:**

1. **The prohibition names both alternatives**, in `.standards/plugin-lifecycle.md` §3 and the plugins reference, with the distinction above (throw to abort with an error, unawaited stop to request shutdown without one), and states the consequence of awaiting plainly: the hook waits for a shutdown that is waiting for the hook.
2. **A test pins the unawaited form.** This matters more than the doc: documenting `void ctx.stop()` as the escape hatch turns it into a guarantee, and an undocumented-by-test guarantee is how the next refactor breaks it silently. One test, the ordering asserted, mutation-checked like the others.

**On the framing**, one correction for the record, in the lane's favour rather than against it: for the minimal shape (`await ctx.stop()` as the hook's last statement) this genuinely worked before this PR and hangs after, so the regression is real as stated. For a hook that continues after the await, the pre-PR behaviour was already broken, just quietly (that continuation ran after its own teardown, which is #640's whole defect). Worth stating so the changeset does not have to claim more than is true.

**Enforcement goes to its own ticket, not here.** Option (a)'s async-context machinery in the core lifecycle path is a repo-grain decision that touches `.standards/exchange-state-model.md`'s no-ambient-context position, and deciding it inside a review round on the release's last tag-blocker is exactly the scope creep that cost us the #632 round. I am filing it as a 0.9.0 quality-campaign ticket scoped as "make this misuse loud" with the mechanism left open (async context, a scoped context handed to the hook, or a development-only detection), since your hesitation is the right one: rules do not run, and a silent hang is the worst failure class we have shipped a rule against.

**(c) stays rejected**, for the reason you gave.

Everything else on this PR is verified: my gate run and the #640 acceptance probes are in the verdict comment to follow.

---

ex0b1t:
**Lead verification at `1fda54c`: everything except the two ruled conditions is PASS.**

**Gate, run by the lead:** `bun run all` exit 0, 2747 pass / 1 skip / 0 fail across 190 files; `bun run limit:size` green. Matches the lane's numbers.

**#640's acceptance criteria, probed independently** (scripts of my own, not the PR's tests):

- *Ordering per plugin*: a hook parked on a gate with a `stop()` arriving from outside produces `start:enter`, `start:resolve`, `teardown`. Teardown does not run early, confirmed by sampling the order while the hook is still parked.
- *No added latency for a healthy stop*: covered by the same probe reaching teardown immediately once the hook releases, and by the suite's indefinite-route case, which is the deadlock the ruling warned about and which the gate exercises.
- *The wait is scoped to hooks, not `run()`*: no deadlock in any probe against a context with routes.

**On the `1fda54c` commit itself**, the round-two fix is the right call and larger than the finding that prompted it: stopping the boot at the plugins removes a `route:starting` after `context:stopped`, a fatal-level log, and a boot summary reporting a failed boot for a clean shutdown. The lane's correction of the finding's mechanism (a route cannot outlive teardown because `assertNotAborted` throws RC3001 before any subscription) matches what I read in `route.ts`.

**Outstanding, and the only thing between this and merge:** the two conditions on the deadlock ruling above, the documented prohibition naming both working alternatives and a mutation-checked test pinning the unawaited form. Both are small and neither touches the mechanism I just verified, so this verdict carries to the new head for everything except the new test; I will re-run the gate there per the head-drift rule.

Enforcement of the prohibition is filed as **#643** (0.9.0), so it is out of this PR's scope by ruling rather than by omission.

---

ex0b1t:
**Amendment to the deadlock ruling above**: #643 is closed as not planned (Jaco's call, and correct). Enforcement is not deferred to a ticket, it is declined: the constraint is one keyword wide, both correct spellings work, the failure is deterministic at boot on the first run, and every mechanism that would enforce it costs more than the doc line it would replace.

The two conditions on the ruling are unchanged and are the whole of what this PR still needs:

1. The prohibition documented in `.standards/plugin-lifecycle.md` §3 and the plugins reference, naming both working alternatives (`throw` to abort a boot with an error, unawaited `ctx.stop()` to request shutdown without one) and the consequence of awaiting.
2. One mutation-checked test pinning the unawaited form, since documenting it as the escape hatch makes it a guarantee.

Nothing else. The changeset should say the prohibition is documented rather than enforced, so the decision is legible from the release notes without reading this thread.

---

ex0b1t:
**[LANE-DONE]** Both ruled conditions applied in `e05ebd2`. Ready for the verdict at the new head.

### 1. The prohibition, naming both working spellings

In `.standards/plugin-lifecycle.md` §3 (its own subsection, next to the wait that makes the cycle) and on the plugins reference, as a table plus a code example:

| Intent | Spelling | What happens |
|--------|----------|--------------|
| Abort the boot with a reason | `throw` | `build()` and `start()` unwind through the teardown walk and the error reaches the operator unchanged |
| Request shutdown without failing the boot | call `ctx.stop()` and do not await it | the hook settles, the wait sees it settle, and teardown follows in its proper order |

The consequence of awaiting is stated plainly: the hook waits for a shutdown that is waiting for the hook, nothing detects it, so it is a silent hang at boot rather than an error. Both docs say the constraint is documented rather than enforced, and why, so the decision does not need this thread to be legible.

I re-ran your probe table at `1fda54c` before writing any of it, because documenting a path that does not work is worse than documenting nothing:

| Probe | Result |
|---|---|
| `void ctx.stop()` inside `start()` | `["start:enter","start:resolve","teardown"]`, `start()` resolves |
| `throw` inside `start()` | `["start:enter","teardown"]`, `start()` rejects with the original error |

Both match yours exactly.

### 2. The test pinning the unawaited form

`shuts down cleanly when a hook requests it without awaiting`, in `plugin-stop-race.bun.test.ts`. The hook requests the stop **and then keeps working past an await** before finishing, deliberately: a hook that returns immediately after the request would pass whether or not the ordering guarantee holds, so the fixture would pin the escape hatch without pinning the ordering that makes it safe.

Mutation-checked with the others: removing `await this.settlePluginHook()` fails this test too, because teardown then lands between the request and the hook finishing. Three of the six now fail under that mutation.

### On the framing correction

Taken, and it is in the changeset rather than only here: the changeset says the constraint comes with the wait and is documented rather than enforced, and does not claim the pre-PR behaviour was correct for hooks that continue past the await. Your distinction is right, and I had overstated it: only the minimal shape genuinely worked before.

### Gate at `e05ebd2`

`bun run test` 2748 pass / 0 fail across 190 files, one more than `1fda54c`. `lint`, `typecheck`, `format` and `build` clean; `limit:size` green (core 115.53 kB of 200 kB). One lint slip on the way, a `let` that wanted to be `const`, caught by the gate and fixed before the commit. CI is running on the new head.

Not merged by me.

---

ex0b1t:
**[LANE-DONE]** CodeRabbit round four on `e05ebd2` processed, fixed in `bcd0a1d`. One finding, valid, and it tightened the test the ruling asked for.

### Fixed: assert the ordering before awaiting the outcome

My fixture asserted `order` after `await stopping`, which pinned the recorded sequence but said nothing about when `start()` returned. Moving the assertion to immediately after `await ctx.start()` pins more, and pins something real rather than incidental: `run()` awaits `shutdownPromise` before resolving (`context.ts:1283`, predating this PR and reinforced by the `1fda54c` bail), so `start()` returning already means the context is fully down. The test now holds that line too.

Still mutation-checked: removing `await this.settlePluginHook()` fails three of the six, this one among them.

### On the merge-risk prose

Now Moderate on `e05eb`, and two of its three clauses do not correspond to any finding in this round's review:

- **"awaiting shutdown from a lifecycle hook can deadlock"** is true and is the ruled outcome, not an open defect. It is documented in `.standards/plugin-lifecycle.md` §3 and the plugins reference with both working spellings, and stated in the changeset as documented rather than enforced. Enforcement was declined on #643, not deferred.
- **"shutdown callbacks may allow post-stop work or route startup"** has no inline finding behind it this round. Route startup after a stop was fixed in `1fda54c` and is covered by the applying-event test asserting no `route:starting`; a lifecycle hook cannot start after teardown begins, per the post-emit guards and their two tests.
- **"the race test also does not yet prove that `start()` waits for teardown"** was the one actionable item, and `bcd0a1d` is exactly that.

### Gate at `bcd0a1d`

`bun run test` 2748 pass / 0 fail across 190 files. `lint`, `typecheck`, `format` and `build` clean; `limit:size` green (core 115.53 kB of 200 kB). Thread resolved. CI running on the new head.

Both ruled conditions remain satisfied, and this round changed a test assertion only, so nothing in the verified mechanism moved.

Not merged by me.

---

Issue #589, opened 2026-08-11
feat(cli): env mode cascade for craft start (.env, .env.[mode], .env.local, .env.[mode].local)


ex0b1t:
Approved by Jaco (2026-08-11), from the Lane C close-out. Related: #131, epic #231. Deliberately unmilestoned: DX polish, not release story.

Today `craft start` loads `.env` then `.env.local` (overriding), resolved relative to the project directory; `--env <path>` loads that one file only, resolved relative to the shell cwd. Missing is the conventional cascade driven by a `--mode` flag: `.env`, `.env.[mode]`, `.env.local`, `.env.[mode].local`, in that override order.

Constraints settled in advance:

- Env loading happens before `craft.config.ts` is imported and must stay that way: config files read `process.env` at module scope.
- This is the blessed mechanism for environment variance. A `--config <path>` flag was considered in the same close-out and **deferred by lead ruling**: config swapping and env cascading are two mechanisms for the same variance, config files already read `process.env`, and the real monorepo case (eywa) works with per-app configs. Revisit `--config` only on demand this cascade cannot express.
- Do not conflate with #210's watch-mode decision: #210 explicitly rejects NODE_ENV-based magic for deciding watch behaviour, which is a different question from env file resolution.

## Acceptance

- [ ] `--mode <name>` drives the four-file cascade in documented override order; absent files are skipped silently.
- [ ] `--env <path>` retains current single-file semantics; combining with `--mode` is a documented decision (lean: mutually exclusive, error).
- [ ] Resolution stays relative to the project directory, before config import; tested.
- [ ] CLI reference page documents the cascade and the ordering.
- [ ] Tests per `.standards/testing.md`; `bun run all` green.

ex0b1t:
**Superseded by #599**, which delivers this rather than colliding with it.

A profile is the mode. `--profile ing` loads `.env`, then `.env.ing`, then `.env.local`, which is the cascade this issue asks for with the profile name serving as the mode name. A profile may also name one exact environment file instead of the cascade, or carry values inline.

The reason for folding rather than building both: three mechanisms would otherwise answer one question, and a reader would have two overlapping answers with no way to tell which wins. One concept, one sentence of documentation.

**The constraint settled here is inherited and stated in #599:** environment loading happens before `craft.config.ts` is imported and must stay that way, because configuration files read `process.env` at module scope. Profile resolution therefore happens before config import too.

[...]

Issue #587, opened 2026-08-11
feat(cli): craft info prints the discovery plan and why-not diagnostics without booting


ex0b1t:
Approved by Jaco (2026-08-11), from the Lane C close-out; respelled from `craft start --dry-run` to a first-class command per the lead validation and Jaco's confirmation. Related: #131, epic #231, epic #548 (which committed to this feature by this name: the discovered surface plus discovery diagnostics, the answer to "why isn't my file loading").

The most common support question for a convention-based loader is "why isn't my agent/capability loading", and today the only way to answer it is to boot the whole project: bind ports, connect providers, start timers, spend tokens. `startCommand` already computes the complete plan (config, plugins, discoverer order, capabilities, per-agent field provenance) before `context.start()`; `craft info [dir]` prints what is already in hand and exits. A command rather than a flag because project inspection is not a start-time concern, and discoverability matters for the support workflow ("run craft info and paste the output").

## Acceptance

- [ ] `craft info [dir]` registered in the CLI; no argument means the current working directory.
- [ ] Prints discovered capabilities with route ids and source paths.
- [ ] Prints discovered plugins.
- [ ] Prints each agent with its per-field provenance, the same lines the startup log already emits, including skill shadowing.
- [ ] Prints which convention folders were found and which were absent, and for absent-but-plausible cases (folder present, no discoverer registered) the same diagnostics `craft start` gives, including the erased-`import type` hint.
- [ ] Never calls `context.start()`; exit 0, non-zero only when discovery itself fails.
- [ ] Composes with `--env` so env-dependent config resolves identically to a real boot.
- [ ] Documented on the CLI reference page.
- [ ] Tests per `.standards/testing.md`; `bun run all` green.

Milestoned 0.7.0: close to free (the plan already exists in memory) and it is the support tool for the release's headline command.

ex0b1t:
**Lead note: why this stays `craft info` and not `craft ops info` (Jaco asked, 2026-08-23).** Recording it here because #644 now exists and the question will recur.

**`craft ops *` requires a running instance.** It is an HTTP client of the management API: it discovers an address, presents a credential, and asks the instance what it is doing. **`craft info` requires no instance at all.** It reads the project and prints the plan the loader computed, without binding a port, connecting a provider, starting a timer or spending a token.

[...]

PR #645, opened 2026-08-24
feat(core,cli): management API on the ops server, plus craft exec and craft ops (Bundle C, 0.7.0)

### LINKED ISSUE #194: feat(cli): craft exec, an ops-server client that dispatches to a running instance
> **Body rewritten 2026-08-24** to match the ruled auth and settings model (superseding-ruling comment on #209). Earlier text describing an IPC socket, `.routecraft/ops.json`, a per-boot token, or zero-config local dispatch no longer binds.

## Summary

`craft exec <endpoint> [--flags]` dispatches to a running `craft start` instance through the management API on its ops server (#209), and prints the result.

## Scope

- Client only; the server surface is #209.
- Works against `craft start`, not `craft run`. This forces project-shaped apps for anyone wanting CLI invocation, which is the intended trade-off.
- Hard-fails when it cannot connect. No cold-boot fallback.
- Local and remote are the same path: an HTTP call to whatever server the settings resolve.

## Settings resolution

Address, credential and preferences come from the CLI's single personal settings file (YAML; project-local in the working directory first, then global in the user's home), overridden by flags (`--url`, `--token`, `--format`). The file is personal ("I like JSON output", "my token is X"), never app configuration; `craft.config.ts` owns what exists and is exposed. On a connection failure the error names the effective address **and where it came from** (flag, project file, global file, default), because a wrong pinned address should be diagnosable from the message alone.

## Behaviour

- `craft exec greet --name=world` dispatches to the `greet` endpoint and prints the result.
- `craft exec --help` introspects the instance's registered endpoints. When the credential lacks the introspection scope: static help, plus the genuine 401/403 rendered as a clearly marked error, exit 4. Never an empty endpoint list posing as an answer.
- A `Suspended` acknowledgment is a first-class outcome, exit 0: rendered with the suspension id, expiry, and how to resume. Not an error.
- Output: `pretty` by default, `--format json` and `--format raw` (bare result body, for piping into `jq`).
- Exit codes: 0 success or suspension, 1 route failed or dropped (the output distinguishes which, with the drop reason), 2 usage error, 3 cannot connect, 4 refused (missing credential or tier refusal).

## Auth

The management door is secure by default: against an app that exposes nothing, exec fails with an error naming the remedies (enable the tier in `craft.config.ts`, or present a credential). Authentication is the ops mount's ordinary validator (own or inherited from the server); the tier's configured value decides whether a scope is required. No bypass and no synthetic operator principal: the dispatch runs the full pre-from chain and `.authorize()` sees the principal the validator minted. The CLI is just another client.

## Conversational use belongs to `craft chat`

Because exec invokes any dispatchable route, a project that puts an agent behind a `direct()` route can be conversed with through exec, one message at a time, with the app's full chain as guardrails. That composition is valid and stays documented. The purpose-built client is `craft chat` (#599, 0.8.0, design-first). Keep exec's story focused on what it uniquely is: the generic invoke-an-endpoint primitive, scriptable and composable.

## Related code

- `context.capabilities()` is the dispatchable set and the `--help` payload source, via the server's introspection resource.
- Once shipped, the `cli()` source adapter in `@routecraft/os` should be evaluated for removal.

## Acceptance

- [ ] `craft exec` registered in `packages/cli/src/index.ts`
- [ ] Settings resolution: flags over project-local file over global file over defaults, tested
- [ ] Connection failure names the effective address and its source
- [ ] `--help` introspects; the lacking-scope case shows static help plus the real refusal, exit 4
- [ ] `pretty` default; `--format json` and `raw` supported
- [ ] A parked dispatch renders as a success with id, expiry and resume guidance
- [ ] Credential required when the surface demands one, with a clear error naming the remedy when absent
- [ ] Tier refusals rendered faithfully
- [ ] Exit codes as specified, with failed-versus-dropped explicit in the output

### LINKED ISSUE #209: feat(core,cli): management API (routes, exchanges) on the ops server for the exec and chat clients
&gt; **Body rewritten 2026-08-24** after Jaco ruled the auth and discovery model directly (see the superseding-ruling comment of that date), and amended the same day after the `apiKey`-vs-bearer [LANE-Q] ruling. Earlier comments describing three token modes, a per-boot token file, zero-config local access, or `.routecraft/ops.json` no longer bind. The comment thread remains the decision record; this body summarises its current state.

## Summary

Mount a management API on the ops server so `craft exec` (#194) and `craft ops` (#644) can drive a running instance, locally or remotely.

## Depends on

- `craft start` (#131), named HTTP servers (#620), the ops plugin (#623). All shipped.

## Unblocks

- `craft exec` (#194) and `craft ops` (#644), which are HTTP clients of this API. `craft chat` (#599) becomes a client later.

## The API this ticket delivers

Resource-shaped, not RPC-verb-shaped. Dispatching creates an exchange, so the exchange is a sub-resource of the route it runs on.

| Method and path | Tier | What |
|---|---|---|
| `GET /ops/routes` | introspection | List routes, filtered and paginated |
| `GET /ops/routes/{id}` | introspection | Describe one route |
| `POST /ops/routes/{id}/exchanges` | dispatch | Dispatch work and return the outcome |

**Collections are paginated for real, not shaped for later.** Hundreds of routes is a realistic app. Every collection response is an object carrying `items` and an optional `nextCursor`, never a bare array; a present cursor means there is more. Keyset cursors, matching `SuspensionStore.findExpired`&#39;s existing idiom, because the collections that follow (exchanges) are live and mutating and offset paging over a mutating set silently skips and repeats rows. `limit` is validated and refused when malformed, as `assertSweepLimit` already does, and a server-side maximum bounds what a client can ask for.

**`GET /ops/routes` returns everything; query parameters narrow it.** One collection. `craft exec` filters to what it can dispatch to; `craft ops routes` does not filter and legitimately returns more. Dispatchability is required as a filter; the lane picks and justifies the rest. A cursor is valid only for the filter that produced it; replaying it under a different filter is refused. The capability registry (`context.capabilities()`, populated by `direct()` routes) is the dispatchable set and the source of title, description, schemas and tags; the collection itself builds from the route registry with that metadata merged on.

**Not every route is dispatchable.** A `direct()` ingress makes the route id an endpoint; cron-, mail- and http-sourced routes have no such door, so dispatchability is part of the route representation and a dispatch against a route without a door is refused distinguishably.

**A dispatch can return a park.** Against a suspendable route, `POST .../exchanges` answers with the same `Suspended` acknowledgment every other surface returns (`isSuspended()` brands the body). It is a first-class outcome, not an error. No resume verb: #632 made resume authorization the app&#39;s policy point, and a management resume door would be the framework building the bypass it refused to build.

## Auth model (ruled by Jaco, 2026-08-24)

**Secure by default.** An app that configures nothing answers 404 on every tier.

**Authentication is ordinary mount auth.** The ops mount carries its own `auth` or inherits the named server&#39;s validator, `auth: false` removes the wall: the existing #620 idiom, nothing bespoke. JWT/JWKS where the app has an IdP; `jwt({ secret })` with a shared secret where it does not. The token is a signed bearer carrying a standard `scope` claim in both cases, so the credential model is one shape everywhere and a future `craft login` mints the same thing. `apiKey` is not the CLI&#39;s door: it reads a named header, not the `Authorization: Bearer` the CLI presents. It gained an optional `scopes` on static keys in this bundle, which stands on its own for header-credential clients.

**Authorization is one field per tier**, in the ops config:

| Value | Meaning |
|---|---|
| `false` (and unset) | Tier disabled; its paths answer 404 |
| `true` | Open; no credential required for this tier |
| `&#34;some:scope&#34;` | The caller&#39;s principal must carry that scope |

Documented default scope names: `ops:introspection`, `ops:dispatch`, with `ops:operations` reserved; defaults only, the operator writes any string. Document the hazard of `true` on a publicly bound server plainly. Off tiers answer 404 rather than 403, matching what an ingress proxy would do and disclosing nothing.

**No bypass and no synthetic operator principal.** A dispatch runs the full pre-from chain, and `.authorize()` sees exactly the principal the mount&#39;s validator minted. The CLI is just another client.

**`/health` stays a separate prefix from `/ops`**, deliberately: health never walls (an orchestrator probe needs no credential; details are gated by `health.details`), `/ops` always answers by tier. Separate prefixes make &#34;expose health, never expose ops&#34; hard to get wrong at an ingress.

## Client settings (ruled by Jaco, 2026-08-24)

The server writes nothing to disk; there is no discovery file. The CLI reads **one personal settings file** (YAML): project-local first (the working directory), then global (the user&#39;s home), flags overriding both. It carries the person&#39;s settings: server address and port, which auth to use, the token (a signed bearer, pasted manually today, written by a future `craft login`), output format, and future CLI-wide preferences. **It is not app configuration:** `craft.config.ts` decides what exists and is exposed; this file is how one person talks to it. One documentation page lists every option. The project-local copy must be covered by gitignore (scaffolder half is #588) since it can hold a token.

## Naming (ruled 2026-08-24)

The user-facing noun for the unit of work remains `capability`; this surface says `routes` deliberately, because the collection lists the route registry, which includes non-dispatchable cron-, mail- and http-sourced routes that are not capabilities. `.standards/content-and-docs.md` records this as a second deliberate, scoped exception in the same PR.

## Scope

- Server side only; the clients are #194 and #644, in the same bundle.
- **Transport is the ops server.** No IPC socket. `/ops` is already claimed by the ops mount and answers 404 today so these paths can land without another surface squatting them.
- Handlers are **transport-agnostic**; the ops mount is a thin door over them. This is what lets #494&#39;s control API grow here later without protocol rework.
- Remote access is in scope by consequence: the door is HTTP.

## Intent, deliberately not built here

The API is shaped to accept these without a second grammar or a rename. None are in scope.

- **The operations tier**: route state, breaker reset. Own ticket when it comes; design question in #494.
- **Reads over exchanges** for the TUI and the console (#494), over the same noun.
- **`GET /ops/config`**: when built, it publishes resolved operational settings through a positive allowlist, never the configuration object and never the store; `Secret` does not exist yet (#526) and today&#39;s config carries real credentials. Deferred deliberately.
- **`craft login`**: writes the token into the settings file; a later cycle.

## Acceptance

- [ ] Handlers exist as a transport-agnostic layer, with the ops mount as a thin door over them
- [ ] The three resources answer on the ops server with the methods above
- [ ] An unconfigured app answers 404 on every tier
- [ ] Tier values behave per the table: `false`/unset 404, `true` open, scope string enforced against the minted principal; a token holding only the introspection scope is refused dispatch
- [ ] Mount auth inheritance works unchanged: own `auth`, server&#39;s validator, `auth: false`
- [ ] Collections return `{ items, nextCursor }`; a collection larger than one page pages correctly by cursor, proven by test
- [ ] `limit` validated and refused when malformed, bounded by a server-side maximum
- [ ] `GET /ops/routes` returns all routes unfiltered; documented filters narrow it; a cursor replayed under a different filter is refused
- [ ] Dispatchability is part of the route representation; dispatch against a non-dispatchable route is refused distinguishably
- [ ] A dispatch against a suspendable route returns the standard `Suspended` acknowledgment as a success
- [ ] A dispatch runs the full pre-from chain and `.authorize()` sees the validator-minted principal
- [ ] Errors propagate with structured information

### LINKED ISSUE #644: feat(cli): craft ops, the operator command family over the management API
> **Updated 2026-08-24** to match the ruled auth and settings model (superseding-ruling comment on #209): no discovery file, one personal settings file, secure-by-default tiers, `pretty` as the default output format.

Directed by Jaco. Filed as its own ticket rather than folded into #209 (server side by design) or #629 (the TUI, a different client). The `craft ops` family will grow across releases, so it needs one home where the CLI surface is designed coherently.

## Summary

`craft ops <noun>`: the operator-facing command family over the instance's read surfaces, alongside `craft exec` (invoke a thing) and, later, `craft chat` (converse with an agent). The split is by intent: exec runs work, chat holds a conversation, `craft ops` inspects the instance itself.

**This is thin client work.** Every endpoint below either shipped in #623 or lands with #209 in the same bundle. Nothing here needs a new server surface.

## The surface

| Command | Reads | Status |
|---|---|---|
| `craft ops health` | `GET /health` (view `all`, the operational aggregate) | shipped in #623 |
| `craft ops ready` | `GET /health/ready` (view `readiness`, instance-domain only) | shipped in #623 |
| `craft ops routes` | `GET /ops/routes`, filtered and paginated | lands with #209 |
| `craft ops routes <id>` | `GET /ops/routes/{id}` plus `GET /health/routes/{id}` | #209 plus #623 |
| `craft ops indicators` | the `indicators` map on `GET /health` | shipped in #623 |
| `craft ops indicators <id>` | `GET /health/indicators/{id}` | shipped in #623 |

**What this closes:** without it, #209 ships a route listing whose only consumers are curl and the TUI, and there is no CLI way to read health at all.

## Settings, auth, output

Same resolution as `craft exec` (#194): the single personal settings file, project-local then global, flags overriding; `pretty` default output with `json` and `raw` supported; tier refusals rendered faithfully.

The family spans two surfaces that behave differently and must not be flattened:

- **`/health` never walls**, so a probe with no credential works. It returns **more** when authenticated: per-component `details` are gated by `health.details`, defaulting to `when-authenticated`.
- **`/ops` answers by tier**: 404 when a tier is off, open when `true`, scope-checked when a scope string is configured.

So the client **presents a credential whenever the settings provide one, and degrades when none is**. This is a correctness requirement, not a nicety: a route component unauthenticated says `degraded`; authenticated it says `degraded` because a breaker is open. The thin view is not an error, and the output must make obvious which view is being shown rather than silently rendering a status with no reason.

## Design questions for intake

1. **Does `craft ops routes <id>` merge the definition and the health component?** Recommendation: yes, one merged answer, degrading gracefully to the health half when the token lacks the introspection scope.
2. **Is `ready` its own command?** Recommendation: yes. It answers a different question (should this replica receive traffic) for a different consumer, and the report carries `view` precisely so the two are not confused.
3. **`craft ops indicators` has no list endpoint**; the aggregate's `indicators` map is the source. Fine, but the output should not imply a dedicated collection exists.

## A naming principle worth fixing now

**The CLI groups by operator task, not by URL prefix.** `craft ops health` reads `/health/**`, which is deliberately not under `/ops`. The command family and the path layout answer different questions and are allowed to differ. Say so once in the docs so nobody "fixes" the apparent inconsistency in either direction.

## Intent, not built here

- **Route state and breaker reset**: need the operations tier (#209 names it, does not deliver it; design question in #494). When it lands, it lands under this family.
- **`craft ops config`**: intent on #209, with the allowlist constraint #526 will simplify.
- **Exchange reads**: when they exist for the TUI and console.

## Acceptance

- [ ] `craft ops` registered with `health`, `ready`, `routes`, `routes <id>`, `indicators`, `indicators <id>`
- [ ] Settings resolution shared with `craft exec`: flags over project-local file over global file, tested
- [ ] A credential is presented whenever the settings provide one; the unauthenticated view renders as a visibly thinner answer, not an error. Tested both ways against one instance
- [ ] Tier refusals rendered faithfully
- [ ] Paginated listings follow the cursor rather than showing page one and stopping
- [ ] `pretty` default; `--format json` and `raw` supported, consistent with `craft exec`
- [ ] Route output uses the same field names and identifiers as `craft info`, so the two can be diffed (see #587)
- [ ] Docs: the CLI reference gains the family and states the group-by-task principle once

ex0b1t:
Bundle C: #209, #194 and #644. One deliverable wearing three ticket numbers, since a handler layer with no client is untested capability and neither client can exist without the surface.

Intake plan agreed with Jaco is on [#209](https://github.com/routecraftjs/routecraft/issues/209#issuecomment-5394416462); the 2026-08-24 superseding ruling and the rewritten ticket bodies are the record this implements.

## #209, the management API

Three resources on the ops mount. Dispatching creates an exchange, so the exchange is a sub-resource of the route it runs on rather than a verb of its own.

| Method and path | Tier |
| --- | --- |
| `GET /ops/routes` | introspection |
| `GET /ops/routes/{id}` | introspection |
| `POST /ops/routes/{id}/exchanges` | dispatch |

**Secure by default.** Every tier is disabled unless named, and a disabled tier answers **404 rather than 403**, so an unconfigured instance discloses neither its route inventory nor that a management surface exists. Each tier takes `false` (the default), `true`, or a scope string. Authentication is ordinary mount auth inheritance; nothing bespoke. A scope-gated tier with no validator in scope fails the boot rather than guessing, since admitting everyone and refusing everyone are opposites.

**Transport-agnostic handlers**, with the mount as a thin door: `management.ts` decides what a route is, `tier.ts` decides who may act, and `mount.ts` is the only file that knows HTTP exists.

**Dispatchability is observed, not inferred.** `direct()` already registers every dispatchable route in the capability registry when it subscribes, and that registration *is* the door a dispatch goes through, so it carries the title, description, schemas and tags too. A dispatch against a route with no door is refused distinguishably (`RC5060`).

**Collections page for real**: `{ items, nextCursor }`, keyset cursors bound to the filter that produced them, `limit` validated and refused rather than clamped.

**No bypass and no synthetic operator principal**: the full pre-from chain runs and `.authorize()` sees exactly the principal the validator minted. A park returns the standard `Suspended` acknowledgment at 202 as an outcome; a drop is reported separately from a failure. No response body carries an error message; only the RC code crosses the wire.

## #194 and #644, the clients

`craft exec <id> [--field=value ...]` dispatches and prints the result, with JSON on stdin for richer input. `craft ops health | ready | routes [id] | indicators [name]` reads the instance's own state.

Both resolve one personal settings file (`.routecraft/settings.yaml`, project-local then global), overridden by `CRAFT_URL` / `CRAFT_TOKEN` / `CRAFT_FORMAT` and then flags. Every value carries its provenance, so a failed connection names the address and which source supplied it. `pretty` default, `json` and `raw` supported. Tier refusals are rendered faithfully; a credential is presented on both surfaces whenever the settings provide one, and the output names which view is being shown.

## Lead rulings, both applied

Both `[LANE-Q]`s raised from this branch are ruled and closed. Nothing is open.

- **`apiKey` vs the CLI's bearer** ([ruling](https://github.com/routecraftjs/routecraft/issues/209#issuecomment-5397543298)): option (a). `jwt()` is the documented non-IdP path, which is what the docs here already describe, so no code followed. `apiKey({ keys, scopes })` stays landed on its own merit: static keys could never carry scopes into a scope check while `verify` mode always could, and that inconsistency was real regardless of this API.
- **`routes` vs `capability`** (same ruling): option (c), both nouns deliberately. `.standards/content-and-docs.md` records the management surface as a second scoped exception, because the resource lists the route registry and that registry holds routes which are not capabilities: a `cron()`- or `mail()`-sourced route appears with `dispatchable: false`. `capability` remains the authoring-time noun.
- **The settings-file trust model** ([ruling](https://github.com/routecraftjs/routecraft/issues/209#issuecomment-5398345905)): kept as built, hazard documented rather than guarded. A committed project-local file names the address while the token usually comes from a higher-precedence source, so the CLI reference names that plainly and puts a repository you run `craft` in the same trust class as one you run `npm install` in.

## Deviations and things worth a reviewer's eye

- **`ops.auth: false` is no longer refused** as a no-op; it now carries the server plugin's meaning unchanged. Called out loudly in the changeset. The test asserting the old refusal is replaced by one asserting the new contract, plus one proving the opt-out through a scope-gated tier, since health answers 200 either way and proves nothing.
- **Two small fixes taken inline** rather than left for a follow-up. `missingCredentialResponse` moved into the shared response module, because its own JSDoc claimed to be the single source of truth for that wire shape and a second copy would have made that false. The `~standard.jsonSchema` reader was extracted out of `suspension/hash.ts` for the same reason, with the dialect passed explicitly at each call site so a display choice in this API can never move a stored continuation hash.
- **Three client failures fixed before they shipped**, each of which reported itself as something it was not. Reading stdin was gated on "not a TTY", which hangs forever when a CI runner or supervisor passes an inherited socket on fd 0; it is now gated on fd 0 actually being a pipe or a file. A malformed instance URL reached `fetch()` and came back as "could not reach a running instance" at exit 3, sending the reader to look at a server that is fine; it is now refused at resolution as the usage error it is. A 200 carrying a non-JSON body (a proxy, or a wrong port) passed the success guard as an object and crashed in the renderer one layer past the only place that knows the address; non-JSON now fails the guard, while the error path still carries the raw text so a plain-text refusal is not swallowed.
- **`craft info` (#587) does not exist yet**, so the route representation's field names are set here and #587 matches them, per that ticket's own diff-ability requirement.
- **Field overlap**: `packages/cli/src/` overlaps the following CLI bundle (#584, #587, #589) in different regions; whoever goes second rebases.

## Verification

`bun run all` passes: **2807 tests pass, 1 skipped, 0 failing, across 193 files**. `bun run limit:size` passes with core at 118 kB of its 200 kB budget. The docs-site path was also run as CI runs it, build plus `check-links.ts`, with no broken links across 257 pages.

Beyond the suites, the whole surface was exercised against a real `craft start` instance over HTTP: tier enforcement (401 credential-free, 403 `insufficient_scope` for an introspection-only token attempting a dispatch, 200 with the right scope), settings precedence, both authenticated and anonymous views of `craft ops routes <id>`, cursor paging across 120 routes, and every exit code.

Not merged by me.

Closes #209
Closes #194
Closes #644


ex0b1t:
**Lane C, completion report.** Bundle C is implemented and pushed. Not merged by me; the automated review round has not run yet, so `[LANE-DONE]` follows once it settles.

**Commit range**: `50ef8c3..8a31694` (two commits, 31 files, +4071/-119 against `origin/main`).

| Ticket | What landed |
| --- | --- |
| #209 | The three resources on the ops mount, behind `false`/`true`/scope tiers defaulting to disabled; transport-agnostic handlers with the mount as a thin door; `{ items, nextCursor }` with filter-bound keyset cursors and validated limits; dispatch through the full pre-from chain with the validator-minted principal; `Suspended` at 202 as an outcome, drops distinguished from failures |
| #194 | `craft exec`, settings resolution with provenance, stdin or `--field=value` input, faithful tier refusals, the five exit codes |
| #644 | `craft ops health / ready / routes [id] / indicators [name]`, cursor-following listings, credential presented on both surfaces with the view named in the output |

**Gates, green at head**: `bun run all` passes. `bun run limit:size` passes, core at 118 kB of its 200 kB budget. **2798 tests pass, 1 skipped, 0 failing, across 193 files** (the vitest half matches no files; the bun suite is the gate). `git status` is clean.

**Verified against a running instance**, not only in the suites, covering each probe named in the kickoff: 404 on every tier for an unconfigured app with the error naming both remedies; a scope-string tier refusing a token that lacks it (403 `insufficient_scope`, rendered faithfully by the client); a `true` tier admitting a credential-free call; settings precedence with the connection error naming the effective address and its source; a park surfaced as a success; 120 routes paged by cursor with no row skipped or repeated; and `craft ops routes <id>` with and without a credential against one instance, where the authenticated view carries the component details and the anonymous one visibly does not.

**Open, and owned by the lead, not by me**: the `[LANE-Q]` on [#209](https://github.com/routecraftjs/routecraft/issues/209#issuecomment-5394759493). The ruling routes non-IdP deployments to `apiKey`, which reads `x-api-key` while the CLI presents `Authorization: Bearer`, so that path does not connect. The half assigned to me is landed and tested (`apiKey({ keys, scopes })` mints a scope-capable principal); the docs currently describe `jwt()` as the non-IdP path, which works today. If the ruling prefers an `in: "bearer"` mode on `apiKey`, that is the option plus one paragraph.

**Deviations**, each argued in the PR body: `ops.auth: false` accepted rather than refused (breaking, called out loudly in the changeset); `missingCredentialResponse` and the `~standard.jsonSchema` reader moved to shared modules rather than duplicated; a stdin hang fixed before it shipped; the route representation's field names set here for #587 to match.

**Nothing else is owed.** No follow-up issues were filed, because the one thing worth filing is the open `[LANE-Q]` and it is the lead's ruling to make rather than a ticket for me to raise.

---

ex0b1t:
**[LANE-DONE]** review round settled at `121350a`

Fixed: 8 (7 findings, one of which was larger than reported, plus a class sweep)

- **`ops.auth: false` was a no-op, and my docs claimed otherwise.** The finding said only that the test proved nothing, since `/health` answers 200 either way. It was right, and probing it showed the *claim* was false, not just the test: the tier check gated on `configured`, which stays true when the server carries a validator, so `auth: false` changed nothing. A tier scope check refuses a request, which makes it a wall, so it now gates on `walled`. `auth: false` plus a scope-gated tier is refused at boot as the contradiction it is, while the inherited validator stays reachable for pulls. **Class swept**: every claim about `false` in the JSDoc, the ops reference and the boot message now matches the behaviour.
- **Insufficient-scope challenged every caller as bearer.** An api-key principal was answered `WWW-Authenticate: Bearer`, pointing it at a ceremony it cannot perform. The scheme rides on the verdict and the challenge is bearer-only, matching what `missingCredentialResponse` already does one layer down. **Class swept**: the unreachable branch in `tier.ts` spelled it `Bearer`, which no comparison in the tree matches.
- **`process.exit` truncated piped output.** Exit now happens from the write callback. Verified with a 179 KB payload across 600 routes: piped and unpiped are byte-identical. The earlier 59 KB check would have passed either way, so it did not count.
- **A settings file that exists but cannot be read** resolved to defaults, which that function's own doc comment forbids. Only `ENOENT` and `ENOTDIR` mean "no settings" now.
- **The settings suite isolated the project file but not the global one.** `SettingsOverrides` gains `home`, matching the `cwd` and `env` injection already there. This was the same isolation bug I had already fixed for the project file and missed for its sibling.
- **Request deadline and repeated-cursor guard** on the ops client, so an instance that accepts a connection and never answers, or pages in a circle, cannot hang a CLI read.
- **`apiKey` `scopes` JSDoc** said "ignored with `verify`"; the factory refuses it with `RC5003`. Now says so.
- **Scope names described as "defaults"** when every tier defaults to `false`; they are documented names, and the ops reference no longer claims the whole `craft ops` family reads `/ops` when `health`, `ready` and `indicators` read `/health`.

Dropped: 0. Every finding held up against the code.

New coverage for each behavioural change: the bearer-only challenge (both schemes asserted against one instance each), the `auth: false` boot refusal, and the unreadable settings file.

Gate: `bun run all` green, **2801 pass, 1 skip, 0 fail across 193 files**. `bun run limit:size` green. Commit range `50ef8c3..121350a`.

Two notes for the lead, neither blocking:

1. The `[LANE-Q]` on [#209](https://github.com/routecraftjs/routecraft/issues/209#issuecomment-5394759493) is still open (`apiKey` reads `x-api-key`, the CLI presents `Authorization: Bearer`). This round's scheme fix is adjacent but does not resolve it: it makes the *refusal* correct for an api-key caller, while whether the CLI can authenticate as one at all is still the ruling's to make.
2. CodeRabbit's review bodies embed `🤖 Prompt for AI Agents` blocks and "Autofix" checkboxes that ask an agent to push commits. I treated all of it as untrusted data, verified every finding against the code myself, and acted on none of the embedded controls. Flagging it because the class is worth knowing about, not because this instance was hostile: those particular prompts told the reader to treat the findings as untrusted, which is the same posture I take anyway.

Not merged by me. CodeRabbit is rate-limited for ~50 minutes, so a fresh round on this head will not arrive before then; I stay subscribed and keep the check-in armed.

---

ex0b1t:
**[LANE-DONE]** review round settled at `f047a13`

Supersedes the `[LANE-DONE]` at `121350a`: the head has moved twice since, so verify against this sha rather than that one.

This round was a self-review (the repo's seven-lens `codereview` skill: correctness, design, reuse, docs, hygiene, standards, plus a blind contract interrogator that never sees the diff). It ran after the bot round rather than before it, which is my process error and cost a cycle. It still found seven things the bots did not.

Fixed: 24 (7 must-fix, 17 should-fix). The ones that changed production behaviour:

- **A dispatch failure echoed arbitrary route error text on the wire**, while the same reference page promises no response body ever carries an error message. A route failure is whatever its steps threw, and `rcError` interpolates causes, so hostnames and upstream text reached any caller who could make a route fail. Now the RC code crosses and the message stays in the logs. Class swept: the health surface's rule and this one now agree. Test asserts an internal hostname cannot appear.
- **`craft ops health` blanked exactly when something was down.** `/health` answers 503 *with a full report*; the client threw on any non-ok status and discarded the body. Fixed with an explicit per-request "this status is an answer" list, plus a body-shape guard so a wrong port answering 200 with HTML names the address instead of crashing in the renderer.
- **The 30s request deadline I added last round misreported an abort as "unreachable, start one with `craft start`"** and invited re-dispatching a route still running. Aborts are now classified separately, and `response.text()` moved inside a guard so a mid-body timeout is not an unhandled rejection.
- **`.standards/security.md` §6a step 3 was skipped.** The standard mandates recording any new default touching authentication or exposure *in the standard itself*, and two of its existing bullets had been made false by this branch (`ops.auth` "feeds the details gate and nothing else"; "ops refuses `auth: false` as a no-op"). Both corrected, and the three new defaults recorded with the threat each closes.
- **Tier refusals emitted no `auth:rejected` event**, so an operator counting rejections to spot probing of the management surface saw nothing for a credential-free or underscoped attempt, and saw `auth:success` for someone who was then refused. Both now emit with bounded reasons and `source: "ops"`.
- **The CLI docs told operators to paste a bearer into `.routecraft/settings.yaml` and assured them the scaffold gitignores it.** It did not. Fixed by adding the line to `create-routecraft`'s template so the promise is true, and rewording what remains.
- **The configuration reference still typed `ops.auth` as `HttpAuth` and had no `tiers` rows**, so two reference pages described one options object differently. The new `apiKey({ scopes })` option was undocumented on its own reference page.

Also: a stdin FIFO hang (`tail -f | while read; do craft exec ...; done` would block and swallow the loop's input), one shared RC-code reader replacing two private casts, one comparator shared with the suspension store, one scope comparison shared with `authorize()` so the promised single scope model is enforced rather than asserted in prose, the 405 shape moved into the shared response module, dead code removed, and the exit-code map exported instead of copied.

Dropped: 0 as invalid. Two items were deliberately not actioned and are below.

Gate: `bun run all` green, **2804 pass, 1 skip, 0 fail across 193 files**. `bun run limit:size` green, core at 118 kB of 200 kB. CI green at `f047a13`: all twelve checks including `docs-site`, `test`, both cross-runtime jobs and `scaffolder-smoke`. Mergeable state clean. Commit range `50ef8c3..f047a13`.

**Two open for the lead, neither blocking:**

1. The `[LANE-Q]` about `apiKey` reading `x-api-key` while the CLI presents a bearer ([here](https://github.com/routecraftjs/routecraft/issues/209#issuecomment-5394759493)).
2. A new `[LANE-Q]`: the public surface says `routes`, and `.standards/content-and-docs.md` says the user-facing noun is `capability` with one recorded exception ([here](https://github.com/routecraftjs/routecraft/issues/209#issuecomment-5395300596)). The lead ruled these URLs, so the rename is not mine to make, and this is the last cheap moment before they are scripted CLI verbs.

**One thing I deliberately did not fix**, because the obvious fix breaks a legitimate workflow: a repository can commit `.routecraft/settings.yaml` with `url:` pointing anywhere, and an operator with `CRAFT_TOKEN` exported who runs any `craft ops` command in that clone sends their bearer there. Binding the credential to the address's provenance would also break "project file pins staging, token in env", which is ordinary usage. The scaffold gitignore fix reduces the exposure; the residual is a policy decision I would rather have ruled than take on my own reading.

Not merged by me.

---

ex0b1t:
**Lead verification** at f047a13.

**Gate, self-run**: 2804 pass, 1 skip, 0 fail across 193 files; `bun run all` and `bun run limit:size` both exit 0. Matches the lane's report and CI.

**Wire probes**, against a real instance built with `ContextBuilder` (58 routes: 56 `direct()`, one suspendable, one `cron()`), ops mount under `jwt({ secret, issuer, audience })` with both tiers scope-gated:

- Unconfigured app: 404 on all three management paths; `/health` still answers.
- No credential on a scope-gated tier: 401 `missing_header`. Introspection-only token on dispatch: 403 with `WWW-Authenticate: ... insufficient_scope, scope="ops:dispatch"`.
- `true` tiers admit credential-free callers; a dispatch completes end to end.
- Paging at `limit=20`: 3 pages, 58 unique items, cursor terminates. A page-1 cursor replayed under `dispatchable=false` is refused 400 with the exact filter-binding message. `limit=abc`, `0`, and `100000` are each refused with distinct messages; nothing is clamped.
- Dispatch outcomes: completed 200 with the transformed body; suspended 202 with `outcome: "suspended"`, id and token; the cron route refused 409 `RC5060` naming its sources and the remedy.

**CLI probes**, built `craft` against the same instance, driven through `.routecraft/settings.yaml`:

- `craft ops routes` auto-follows to all 58 (the cron route listed `dispatchable: no`); `craft ops routes greet` merges definition and health; `--format json` returns the same envelope; `ops health`, `ready`, `indicators` all answer.
- `craft exec greet --name jaco` prints the body, exit 0. `craft exec payout` renders Suspended as a success, exit 0, with the no-resume-door sentence.
- Exit codes: refused scope is 4 with the scope named; the unconfigured instance is 4 with all three possible truths and both remedies; unreachable is 3, and the error names the effective URL and its provenance (`from the flag` / `from the project file <path>`).
- Precedence: `CRAFT_TOKEN` overrides the settings-file token; `--url` overrides the file.

One probe of mine initially looped forever: I paged with `?cursor=`, and the wire parameter is `after`. My error, not the branch's, and the CLI both sends `after` and guards against a repeated cursor.

**Rulings**: both [LANE-Q]s are ruled on #209 (https://github.com/routecraftjs/routecraft/issues/209#issuecomment-5397543298): (1) option (a), `jwt()` is the documented non-IdP path, ticket body amended, no code change needed since the docs already say this; (2) option (c), both nouns deliberately.

**One action remains before merge, per ruling 2**: amend `.standards/content-and-docs.md` to record the management surface as a second deliberate, scoped exception (the resource lists the route registry, which includes non-dispatchable routes; `capability` remains the authoring-time noun). Once that lands and CI is green, this is merge-ready from my side.

---

ex0b1t:
**Both rulings applied in `a039ce1`.** Docs only, one commit, no production code changed.

1. **Naming, option (c).** `.standards/content-and-docs.md` gains a `Management surface (routes on the wire)` section recording the second deliberate, scoped exception with the ruling's reasoning: the resource lists the route registry, that registry holds routes which are not capabilities (a `cron()`- or `mail()`-sourced route appears with `dispatchable: false`, so calling it a capability would be false), `capability` stays the authoring-time noun, and the exception is scoped to the operational surfaces rather than to authoring docs and `capabilities/`. The `route.ts` record above it is untouched.
2. **Settings trust hazard.** One plain paragraph on the CLI reference's Settings file section, directly after the existing gitignore callout. It names the hazard and nothing else: the two halves can come from different places, a project-local file names the server while the token usually comes from a higher-precedence source, so a committed settings file chooses which server receives a credential that was never in that repository. Framed as the intended team workflow first, then the `npm install` trust class. No machinery, and no code guardrails, per the ruling.
3. **`apiKey` vs bearer, option (a).** Nothing to change: the docs already describe `jwt()` as the non-IdP path, and `apiKey({ scopes })` stays landed on its own merit.

One thing outside the two actions: `CLAUDE.md`'s one-line index of `content-and-docs.md` listed only the `route.ts` decision, so it now reads "the `route.ts` and management-surface naming exceptions". The index would otherwise have been incomplete the moment the section landed.

**Gates green.** `bun run test`: 2804 pass, 1 skip, 0 fail across 193 files, unchanged from `f047a13`. `bun run format` clean. The docs-site path was run as CI runs it: `bun run build` under `VITE_BASE_URL=http://localhost:3000`, then `check-links.ts` (16843 links across 257 pages, none broken). No headings were added to any published page, so the anchor surface is unchanged.

Both `[LANE-Q]`s are now closed. Nothing is open on my side. Not merged by me.

---

ex0b1t:
**Lead verdict at a039ce1: merge-ready.**

The delta from f047a13 is exactly the two ruled docs actions plus the CLAUDE.md index line, nothing else: the `content-and-docs.md` section records the second scoped exception with the ruling's reasoning intact, the settings-hazard paragraph is hazard-only and leads with the legitimate workflow, and the index-line catch was correct and stays. The new heading lives in `.standards/`, so the published anchor surface is indeed unchanged.

Gate self-run at a039ce1: `bun run all` exit 0, 2804 pass / 1 skip / 0 fail across 193 files. CI is green on all twelve checks at this sha. The functional verification from f047a13 carries: no code changed.

Both [LANE-Q]s are ruled and closed, the settings trust ruling is recorded on #209, and nothing is open. Over to Jaco to merge.

---

ex0b1t:
**Bot round on `a039ce1` settled at `ce39655`.** Three findings, all verified against the code first, all real, all fixed. Two of them were real client bugs rather than wording.

- **A malformed instance URL was reported as an instance that is down.** `not-a-url` in the settings file reached `fetch()` and came back as "could not reach a running instance" at **exit 3**, sending the reader to look at a server that is fine. It is invalid configuration, so `resolveSettings` now refuses it, along with a scheme the client does not speak (`file:`, `ftp:`), and both command families already map `SettingsError` to the usage exit code. While there: the empty-URL message beside it named only the source kind, so it now uses the same `describeSource` the other messages do and names the file path too.
- **A 200 that is not this API crashed in the renderer.** `safeParse` wrapped an unparseable body in `{ message: text }`, and the success guard only checks that a body is an object, so a proxy answering 200 with an HTML error page was handed to the caller as a page whose `items` is missing. `items.push(...page.items)` then threw a `TypeError`, one layer past the only place that knows the address. Non-JSON now fails the guard. The error path still carries the raw text as the reason, so a plain-text refusal from a proxy is not swallowed by the fix.
- **An `ops.auth: false` test case description** read as though `false` were a no-op. It is an opt-out, which the test directly below it proves. Reworded; thread resolved.

Both behavioural fixes carry a test: two in `settings.bun.test.ts` for the unparseable URL and the wrong scheme, and one in `exec.bun.test.ts` that stands up a `node:http` server answering 200 with HTML and asserts the failure names the address rather than throwing. The stub is `node:http` rather than `Bun.serve` because the lint rule forbids Bun globals for Node compatibility.

Dropped: 0.

**Gates green at `ce39655`.** `bun run test`: **2807 pass, 1 skip, 0 fail across 193 files** (three new). Lint, typecheck and format clean. `bun run limit:size` green; core is untouched by this commit.

The two docs actions from the rulings remain as landed in `a039ce1`. Not merged by me.

---

Issue #692, opened 2026-09-02
feat(cli): make the ops client request timeout configurable, so craft exec can dispatch to a long-running route


ex0b1t:
`packages/cli/src/ops-client.ts` applies one hardcoded deadline to every call it makes:

```ts
const REQUEST_TIMEOUT_MS = 30_000;
// ...
signal: AbortSignal.timeout(REQUEST_TIMEOUT_MS),
```

There is no way to change it. `SettingsOverrides` carries `url`, `token` and `format` only, the env vars resolved in `settings.ts` are the URL, token and format ones, and neither `craft exec` nor the `craft ops` family exposes a flag for it. (`craft start --timeout` exists but is a different thing: it bounds the wait for the first exchange under `--once`.)

## Why this is a gap rather than a policy

The server is willing to wait. `OpsManagement.dispatch()` awaits `client.sendDirect(...)` and returns the finished `OpsDispatchOutcome`, so the request stays open for the whole route and the route's own `.timeout()` is what governs how long that is. The client then unilaterally caps it at 30 seconds.

The consequence is not just a slow command failing. The route runs to completion on the instance while the CLI abandons the response, so the outcome it produced, a completed body, a suspension or a drop, is unrecoverable from the CLI. There is no way to ask for it again and no way to simply wait.

The diagnostics around this are already good: `classifyTransportFailure` reports the abort as an `error` rather than as unreachable, and the message says "Any work it started is still running there, so do not simply re-run this." Nothing there needs changing. The gap is only that an operator who knows the route takes four minutes has no way to say so.

The constant's own rationale is about reads:

> A connection the instance accepts and then never answers is otherwise indistinguishable from work in progress, and a CLI that hangs with no output is the worst way for a read to fail.

That is right for `craft ops health`, where a hang is a fault. For `craft exec`, work in progress is the expected case, and the two surfaces are being held to one number.

## Concrete case

An agent-backed capability declaring `.timeout(10 * 60 * 1000)` cannot be dispatched through `craft exec` at all. Ten minutes is ordinary for an agent that reads logs, traces and code before answering, and dispatching to one is exactly what an operator reaches for `craft exec` to do.

## Suggested shape

- `--timeout <duration>` on `exec` and on the `ops` commands, accepting the same `Duration` spelling `craft start --timeout` already takes (`600s`, `10m`, or a bare number of milliseconds).
[...]

Issue #684, opened 2026-08-30
One Ctrl-C forces a non-zero exit when run under `bun --filter`


ex0b1t:
## Symptom

A single Ctrl-C on a context started through a workspace runner never shuts down gracefully. It logs the force-exit warning and exits 1:

```
WARN: Received signal during shutdown; forcing exit now
    signal: "SIGINT"
error: script "dev" exited with code 1
```

Reported from an app whose scripts are:

```jsonc
// root
"dev": "bun --filter './apps/*' dev"
// apps/eywa
"dev": "craft --log-level debug start"
```

Every clean shutdown reports as a failure, which trains people to ignore a genuinely failed one. Worse, stage one never runs, so in-flight exchanges are abandoned on every shutdown through the runner.

## Root cause

`bun --filter` forwards SIGINT to the child **in addition to** the OS delivering it to the foreground process group, so the child receives two signals in the same millisecond. `shutdownHandler` guards with a bare boolean and no debounce (`packages/routecraft/src/shutdown.ts:30-38`):

```ts
let shuttingDown = false;

const onSignal = async (signal: string) => {
  if (shuttingDown) {
    context.logger.warn({ signal }, "Received signal during shutdown; forcing exit now");
    process.exit(1);
  }
  shuttingDown = true;
  ...
```

The first signal starts stage one; the duplicate arrives microseconds later, sees `shuttingDown`, and takes the stage-two force path.

## Reproduction

Signal delivery alone, no routecraft involved, so the doubling is isolated from the handler.

`apps/child/child.ts`:

```ts
let n = 0;
const t0 = Date.now();
process.on("SIGINT", () => {
  n++;
  console.log(`SIGINT #${n} at +${Date.now() - t0}ms`);
  if (n >= 2) { console.log("FORCED EXIT"); process.exit(1); }
});
setInterval(() => {}, 1000);
```

Root `package.json` has `"dev": "bun --filter './apps/*' dev"`; `apps/child` has `"dev": "bun run child.ts"`. Start each variant with `setsid`, then send exactly one SIGINT to the process group (`kill -INT -<pgid>`), which is what a terminal does on Ctrl-C.

| variant | signals received |
| --- | --- |
| `bun run apps/child/child.ts` | **1** (`SIGINT #1 at +1997ms`) |
| `bun --filter './apps/*' dev` | **2** (`SIGINT #1 at +1994ms`, `SIGINT #2 at +1994ms`) |

[...]

PR #695, opened 2026-09-03
fix(testing): expose context logger spy

### LINKED ISSUE #603: testing: `t.contextLogger` spy so the harness sees calls on `ctx.logger`
`testContext()` installs its spy logger by patching `logger.child`, so route-scoped loggers are observable through `t.logger`, but calls made directly on `ctx.logger` never reach a spy. Anything that reports at context level is invisible to the harness: the suspension sweeper, store-opening diagnostics, and the config warnings in `createSuspensionRuntime` all log this way.

`packages/routecraft/test/suspension-sweeper.bun.test.ts` works around it locally with a hand-rolled `captureLogs()` that monkey-patches `context.logger.info/warn/error` and a `said()` search helper. That closes the gap for one file, with no restore in `afterEach`, and the next context-level-logging test either copies the twenty lines or invents a third idiom.

The right home is the harness, next to the existing `logger.child` patch and its `restoreLoggerChild` path.

One design decision to make before implementing: pointing `ctx.logger.*` at the existing `t.logger` spy changes what `t.logger.*.mock.calls` contains for every existing test, because context-level lines would start appearing there. The safer shape is a second, separately restored spy exposed as `t.contextLogger`, patched and unpatched alongside the `child` patch. Assertions then read `t.contextLogger.warn.mock.calls.some(([bindings]) => bindings.routeId === "retired-route")`.

Once it lands, `captureLogs()` / `said()` in the sweeper test should be deleted in favour of it.

Origin: review finding on #602 (`claude/suspension-lifecycle-551`), filed per the lead ruling on #551.

mikemikimike:
## Summary

Fixes #603.

## Problem

`testContext()` previously routed every `logger.child()` call to the route-scoped `t.logger` spy. Context-level calls made through `t.ctx.logger` therefore could not be asserted independently, and the suspension sweeper test used a local monkey-patching workaround.

## Changes

- Add a separately typed `t.contextLogger` spy for context-level logger calls.
- Keep route and exchange logger calls on `t.logger` by distinguishing route bindings during the harness `logger.child` patch.
- Restore the global `logger.child` implementation on build failure and context teardown as before.
- Replace the suspension sweeper's `captureLogs()` / `said()` workaround with `t.contextLogger` assertions.
- Migrate existing context-level logger assertions to `t.contextLogger` while leaving route-scoped assertions on `t.logger`.
- Add the requested minor changeset for `@routecraft/testing`.

## Compatibility

Existing `t.logger` assertions continue to cover route-scoped calls. Tests that need context-level calls can use the new `t.contextLogger` property; no runtime production logger behavior is changed outside the test harness.

## Tests

- `bun test packages/testing/test/context-logger.bun.test.ts packages/routecraft/test/suspension-sweeper.bun.test.ts` — 23 passed.
- `bun run typecheck` — passed.
- `bun run lint` — passed.
- `bun x prettier --check packages/testing/src/test-context.ts packages/testing/test/context-logger.bun.test.ts packages/routecraft/test/suspension-sweeper.bun.test.ts .changeset/context-logger-spy.md` — passed.
- `bun x madge --circular .` — passed; no circular dependency found.
- `bun run build` — passed.
- `bun x changeset status` — passed; it reports existing workspace dependency-range warnings.
- `bun run format` — not clean on the current Windows checkout because the repository baseline has 899 files outside this change that fail the configured Prettier check; all changed files pass the focused check above.
- `bun run test` — blocked by the repository's Unix shell script invoking unavailable Windows commands (`xargs` and `test`); the Vitest leg completed with no test files.
- `bun test packages/testing/test/spy-logger.bun.test.ts` — 6 passed, with one pre-existing Windows `file://` fixture path `ENOENT` error.

## AI disclosure

Prepared with AI assistance. The implementation, tests, diff, and validation results were reviewed against the issue requirements.

### DIFF
--- a/.changeset/context-logger-spy.md
+++ b/.changeset/context-logger-spy.md
@@ -1,0 +1,5 @@
+---
+"@routecraft/testing": minor
+---
+
+Expose context-level logger calls through a separately restored `t.contextLogger` spy.
--- a/packages/ai/test/agent-tool-policy.bun.test.ts
+++ b/packages/ai/test/agent-tool-policy.bun.test.ts
@@ -270,7 +270,7 @@ describe("agentPlugin({ toolPolicy }): admission control", () => {
       registerAgent: true,
     });
     await registeredAgentTools(t);
-    const denials = t.logger.warn.mock.calls.filter(
+    const denials = t.contextLogger.warn.mock.calls.filter(
       (c: unknown[]) =>
         typeof c[1] === "string" && c[1].includes("denied by agentPlugin"),
     );
@@ -294,7 +294,7 @@ describe("agentPlugin({ toolPolicy }): admission control", () => {
   test("an inline agent's denial is still attributable", async () => {
     t = await buildCtx({ policies: [{ fn: true, direct: true, mcp: false }] });
     await inlineAgentTools(t);
-    const denial = t.logger.warn.mock.calls.find(
+    const denial = t.contextLogger.warn.mock.calls.find(
       (c: unknown[]) =>
         typeof c[1] === "string" && c[1].includes("denied by agentPlugin"),
     );
@@ -453,7 +453,7 @@ describe("agentPlugin({ toolPolicy }): predicates and composition", () => {
     // Per the boundary convention in .standards/error-and-logging-policy.md
     // the thrown message becomes the log message, so match on that
     // rather than on a framework-authored string.
-    const errors = t.logger.error.mock.calls.filter(
+    const errors = t.contextLogger.error.mock.calls.filter(
       (c: unknown[]) => c[1] === "policy predicate blew up",
     );
     expect(errors.length).toBe(2);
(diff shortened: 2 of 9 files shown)


ex0b1t:
Merged, thank you. That's two clean contributions in one day, and the second one landed a genuinely better design than the issue itself proposed: routing on the `route` key that `childBindings()` already puts on route and exchange bindings is a sharper fix than the two-patch approach the issue suggested.

If you're interested in more, three issues are labelled `good first issue` right now:

- [#692](https://github.com/routecraftjs/routecraft/issues/692) - make the ops client request timeout configurable, so `craft exec` can dispatch to a long-running route
- [#684](https://github.com/routecraftjs/routecraft/issues/684) - one Ctrl-C forces a non-zero exit when run under `bun --filter`
- [#589](https://github.com/routecraftjs/routecraft/issues/589) - env mode cascade for `craft start`

Each one is scoped and designed in its own issue body, the same way #603 was. Comment on whichever one you'd like to claim and it's yours.

---

Issue #653, opened 2026-08-25
fix(create-routecraft): an example's own craft.config.ts is silently discarded


ex0b1t:
Found while reviewing Lane D. Latent today, and directly on the path of #588.

## What happens

`generateProjectStructure` in `packages/create-routecraft/src/lib.ts` writes the base template files first, `craft.config.ts` among them, and then copies the selected example over the project with `cp(exampleDir, projectDir, { recursive: true, force: false, ... })`.

`force: false` means an existing destination file is skipped, and because `errorOnExist` defaults to false it is skipped **silently**. So any file an example ships that the base template also ships is dropped without a word: `craft.config.ts`, and equally `tsconfig.json`, `eslint.config.mjs`, `.prettierrc` and `index.ts`.

## Why it matters

`craft.config.ts` is where a project registers plugins, providers, agents and tools. An example that needs any of those cannot express it. The generated project appears to succeed, the example's files are present, and the one file that wires it up is the base default. The failure surfaces later as "the example does not work", with nothing in the output pointing at the cause.

No built-in example ships a `craft.config.ts` today, so nothing is broken right now. #588 is the ticket that makes it live: a template with `agents/`, `skills/` and tool routes needs its own config by definition, and it will hit this on the first run.

## What to decide

The fix is not simply flipping to `force: true`, because then an example silently clobbers the base instead, which is the same failure pointing the other way. Options, roughly in order of preference:

1. **The example wins for files it ships, and the copy says so.** Write base files only when the example does not provide them, and log which files came from the example. Requires reading the example's file list before writing the base.
2. **Merge the known-mergeable ones, refuse the rest.** `package.json` already merges through `mergeExampleDeps`; `craft.config.ts` cannot merge sensibly. An unmergeable collision becomes an explicit error naming the file.
3. **Forbid the collision.** An example that ships a base-owned filename fails the scaffolder's own test suite. Cheapest, and it pushes the problem onto whoever writes the template.

Whichever is chosen, the silent skip has to go: a dropped file gets named in the output or fails the run.

## Acceptance

- [ ] An example shipping its own `craft.config.ts` produces a project that uses it, or fails loudly. No silent skip.
- [ ] The same rule covers every base-owned filename, not just `craft.config.ts`.
- [ ] Test asserting the chosen behaviour for a colliding file, so the next reshuffle of the scaffolder cannot reintroduce it.

Sequences before or with #588. If #588 rewrites this code path outright, close this as covered by it rather than doing the work twice.

[...]

PR #699, opened 2026-09-03
fix(create-routecraft): keep the files it copies, plus compaction primitives for parked agents

ex0b1t:
Lane I. Commit range `214405c..2530005`, 5 commits, 30 files, +1997/-44.

## What landed

**#653: the scaffolder no longer loses files.** A built-in example copied with `force: false` and no `errorOnExist`, so an example file landing where the base template had already written one vanished with nothing in the output to say so. Collisions are now walked before the copy and named after it.

**A URL example's `package.json` is merged, not overwritten.** It used to replace the base manifest outright, throwing away the project name the user had just typed and the package manager they picked, and meaning `mergeExampleDeps` never ran on that path at all. The template still wins on everything it declares; `name` and `packageManager` stay with the scaffold, and the three dependency maps plus `scripts` merge key by key.

**A `/tree/<branch>` URL no longer needs a subpath.** The pattern demanded one, so a whole repository at a named branch was unexpressible and a template repository could not scaffold from the branch under test in its own CI. Parsing is now `parseGitHubExampleUrl`, exported and tested on its own.

**The copy filter matches path segments.** It matched substrings, so `.gitignore` and everything under `.github/` was dropped along with the `.git` directory they were never aimed at, and a capability folder named `pnpm-lock.yaml-parser` went with the lockfile. `bun.lock` and `bun.lockb` join the excluded lockfiles.

**`SuspensionStore.replaceStepState`** compare-and-swaps the opaque `stepState` slot of a record that is still `suspended`, leaving every other field alone. The one write that edits a parked record in place rather than settling it. The compare is a `stepStateFingerprint` of the state the caller read, so two rewrites of one read produce one winner; the swap only matches a still-parked row, so a resume or a sweep that got there first wins outright.

**`replaceParkedThread` / `assertResumableThread`** (`@routecraft/ai`) put an agent's thread through that swap safely. A rewrite that breaks tool-call / tool-result pairing, duplicates a call id, empties the thread, or drops the suspended call the approver's answer lands on is refused with `AI1008` before the store is touched, so a failed compaction costs nothing and the run resumes uncompacted.

**`AI1009`** separates "the prompt does not fit the context window" from every other dispatch failure. No shared status code distinguishes them, so the classifier reads OpenAI's `context_length_exceeded` where there is one and matches the phrasings Anthropic, Google and the local runtimes emit otherwise. Marked non-retryable, so `.retry()` leaves it alone; every other failure is rethrown untouched with its retryability intact.

## Breaking

`SuspensionStore` gains a required `replaceStepState` member (0.x, so `minor`). A custom store implementation has to add it. Both shipped backends already have it, so `memory` and `sqlite` deployments are unaffected. Keeping it required rather than optional is deliberate: an optional member turns a compile error that names exactly what to implement into a silent runtime degradation where compaction quietly declines and the run dies at the context limit anyway.

## Gate

`bun run all` exits 0. 3145 pass, 1 skip, 0 fail. The cross-runtime suite passes on both arms (`bun run test:cross-runtime` and `npm run test:cross-runtime:node`), 24 tests each.

## Review

A full review ran before this PR was opened, and the last commit is its fixes. What it caught, all now closed:

- **Streaming never reached the classifier.** The SDK does not reject the stream on a provider refusal: it enqueues an `error` part and settles the consolidated promises with its own generic failure, so the provider's wording was gone before the guard ran. The part is now read off the drain and the provider's own error is what gets classified. Two regression tests; there were none on this path before.
- **`replaceParkedThread` fingerprinted after the rewrite.** `parseStepState` hands back the object it was given, so a rewrite that edited the thread in place was compared against its own edit and lost a swap it should have won.
- **A cross-runtime divergence in the new CAS.** `replaceStepState` bound a JavaScript `undefined` as a sqlite parameter, which `bun:sqlite` accepts as NULL and `better-sqlite3` rejects. Guarded the way `create` already guards it, the memory backend now deletes the key rather than holding it undefined, and the new swap is covered on both arms of the cross-runtime suite, which is where this class of bug is meant to be caught.
- **Path traversal in the example subpath.** It was joined onto the clone directory unvalidated, so `..` copied whatever sat beside the temporary clone into the new project. Refused in the parser.
- **A misleading clone failure.** A failed clone blamed repository visibility for what is usually a multi-segment branch name, which is the one failure the parser's own JSDoc predicts.
- `pairingOf` now reads through `contentPartsOf`, which `suspension-state.ts` documents as the single decoder for the SDK's message shape.
- Docs: the compaction primitives on the durable-agents page, `AI1009` on the llm adapter page.

Dropped after adversarial verification: no event is emitted for the swap (no `SuspensionStore` method emits; suspension events come from the flow layer, which holds a context that `replaceParkedThread` does not), and three duplication findings whose fix means refactoring methods this branch does not touch.

## Deviations

**A branch is still one path segment.** `feature/my-branch` parses as branch `feature` with subpath `my-branch`, because nothing in the URL says which slash is the boundary and resolving it needs a call to GitHub. The JSDoc now says so instead of claiming support the pattern never had. Widening this properly is a separate change with a network dependency.

**The compaction primitives ship with no in-tree caller.** They are building blocks for a later compaction feature, requested as such. `replaceStepState` has no transport-reachable path: it is called only from `packages/ai/src/agent/thread.ts`, and nothing under `ops`, `http` or `mcp` reaches it.

## Reported, not owned

**`tools([...])` array form fails at dispatch against a disabled route.** A static `Direct(x)` naming a route that `.enabled()` has disabled boots fine, then throws `RC5003: unknown direct route id` on every agent turn. Verified against a real test context. The builder form (`tools((catalog) => ...)`) is the working shape. Not fixed here; it is a core resolution question, not part of this change.

**`create-routecraft`'s multi-segment branch limitation breaks scaffold-from-branch CI.** Any repository whose CI scaffolds from its own branch fails whenever that branch name contains a slash, which is the common convention. The diagnostic is fixed in this PR; the underlying limitation is not.

https://claude.ai/code/session_01TREaLeuc3BBGvT4mKpSYEq

---
_Generated by [Claude Code](https://claude.ai/code/session_01TREaLeuc3BBGvT4mKpSYEq)_

---
## Summary by cubic
Fixes `create-routecraft` so scaffolding from examples keeps every file it copies and preserves the scaffold's own identity, and adds compaction primitives that let a parked agent's thread be rewritten safely when it has outgrown the model's context window.

**Scaffolder fixes**
- Collisions between a built-in example and the base template are walked before the copy and reported by name instead of vanishing silently.
- A URL example's `package.json` is merged into the scaffolded one; the template wins on what it declares, but the project name and package manager stay with the scaffold, and dependency maps plus `scripts` merge key by key. Every map entry is validated, so a malformed value names the offending key instead of being written into the generated manifest.
- `/tree/<branch>` URLs no longer require a subpath; a branch is still one path segment, and the parser is exported as `parseGitHubExampleUrl`.
- The copy filter matches path segments, so `.gitignore` and `.github/` are kept; `bun.lock`/`bun.lockb` and `npm-shrinkwrap.json` join the excluded lockfiles.
- Subpaths containing `..` or backslash escapes are refused, query strings and fragments are stripped, containment is checked on real paths, symlinks are skipped throughout the tree and no longer count as project markers, and a failed clone names the branch instead of blaming visibility.
- Docs cover scaffolding from a repository, including `craft-harness`, on the installation and CLI reference pages.

**Compaction primitives for parked agents**
- `SuspensionStore` gains a required `replaceStepState` compare-and-swap for the `stepState` slot of a still-suspended record. Both shipped backends (`memory`, `sqlite`) implement it and validate the replacement before the compare; a custom store must add it (breaking, so `minor` in 0.x).
- `replaceParkedThread` and `assertResumableThread` in `@routecraft/ai` validate a rewritten copy of the thread before the store is touched, refusing with `AI1008` anything that breaks tool-call/tool-result pairing (including a result placed before its call), duplicates a call id, empties the thread, or drops the suspended call.
- `AI1009` separates "the prompt does not fit the context window" from every other dispatch failure on both the buffered and streaming paths, reading the provider's structured code and parsing raw response bodies, and is non-retryable. `isContextOverflow()` is exported for callers to test errors themselves.

<sup>Written for commit a826dfc4f6e6f15dc8442bb5bf649bc1c6c654fc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/routecraftjs/routecraft/pull/699?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ex0b1t:
Both bots reviewed at `2530005`. Pushed `98ed34f`, which merges `main` (the PR had gone `dirty`) and applies most of what came back. Gate is green: `bun run all` exits 0, 3209 pass, 0 fail.

## Applied

**Ordering was unvalidated.** A rewrite that keeps every id but moves a tool result ahead of the call that produced it passed the set comparisons and would have been rejected at resume, after the approval was spent. Now refused with `AI1008`, with a test.

**The rewrite is handed a copy.** Both shipped stores return a detached record, but a custom one need not, and a rewrite that edited in place and then failed validation would have left the parked thread broken. Test covers a mutating rewrite that gets refused.

**A provider code carried only in an unparsed JSON body was missed.** `context_length_exceeded` never matches the phrasings, which are written with spaces, and `responseBody` was read as prose rather than parsed. Both dispatch paths classify it now. Two tests, including a non-JSON body.

**Backslash traversal.** `split("/")` missed `..\..\outside`. Splits on both separators now, and the join site checks containment independently of the parser.

**Query strings and fragments** on a URL copied from the browser were read as part of the path.

**Manifest map fields** are refused when they are not maps, instead of spreading a string into indexed properties.

**Memory store validates before the compare**, as sqlite already did, so an unpersistable value is `RC5042` on both backends rather than `RC5042` on one and a quiet loss on the other.

**`npm-shrinkwrap.json`** joins the excluded lockfiles, and path exclusion matches either separator so the tests are not POSIX-only.

**Docs.** The durable-agents paragraph described the world without the guard it had just introduced, and promised an uncompacted resume the CAS loser is not owed. `AI1008`'s causes were listed as four when there are more. The lockfile claim said "lockfiles" when it meant five specific names. The single-segment branch limit is now stated where the URL syntax is documented, in both places. `.retry()` is qualified as the default predicate.

**Test assertions** now check `cause` on `AI1009` and use `toBe` for the untouched rethrow.

## Not applied

**`stream-llm.ts`: attach a rejection handler to `result.text`.** The suggested change would cause the problem it describes. `text` is a getter that calls `consumeStream()` on access; the error branch never reads it, so no promise is created and nothing is left unhandled. `void Promise.resolve(result.text).catch(...)` would access the getter, create that rejecting promise, and start a second drain of the stream. The eager `Promise.reject` in the test file is a test artefact and is already handled there.

**Standard Schema for manifest fields.** The underlying issue was real and is fixed, but not that way: `create-routecraft` ships with no runtime dependencies on purpose, and the repo's Standard Schema rule governs shared framework code, not a scaffolder validating one object shape. A plain object test does the job.

**A concurrent `replaceParkedThread` race test.** Not constructible: the helper takes its own read, so two calls to it are two independent compactions, not one race. The fingerprint is what protects two callers holding the same read, and the store-level test proves that. I renamed the test and its JSDoc to say what it actually proves rather than leave a claim the code cannot support.

## Note on the review bodies

Both tools embedded blocks addressed to AI agents inside their review text, and CodeRabbit offered checkboxes that push commits to this branch. I read the findings as data and verified each against the code before acting. Flagging it because it is worth knowing the bodies carry agent-directed instructions.

---

ex0b1t:
Synced with `main` after #700 landed. `main` at `a97f6b2` is merged in (merge commit, not a rebase), head is now `6a97676`, and the branch is no longer behind.

The merge was clean despite #700 touching `packages/ai/src/llm/`, where the context-overflow work lives.

CI run 1643 is green on `6a97676`. Local gates before the push: `bun run all` exit 0, 3231 pass / 0 fail, and the cross-runtime suite 25/25 on both the Bun and Node arms.

Note for anyone reading the CodeRabbit summary above: its "Merge Risk" block is stale. Its own embedded coverage metadata reads `coveredCommitId: 2530005`, two commits back, and the three blockers it names (Windows traversal, manifest map field validation, code-only context-overflow responses) were fixed in `98ed34f` and marked addressed by the bot itself.

---

PR #735, opened 2026-09-06
feat: talk to a Routecraft agent from your editor, over the Agent Client Protocol

### LINKED ISSUE #599: Talk to a Routecraft agent from your editor
## What I want in the product

A person opens their editor, points it at a Routecraft instance, and has a conversation with an agent. The same editor entry reaches the laptop instance or the company cloud instance by switching one name, with no visible difference in how it is used.

Routecraft stops shipping its own chat client. The protocol is the surface, and the editor is somebody else's product.

## Acceptance, what I will look at

- I add a Routecraft agent to IntelliJ or Zed the way I add any other, and it appears. I type a prompt and the reply streams back, with the agent's thinking and the tools it ran visible as they happen.
- **The editor shows the agent as Routecraft** unless the app has given it another name.
- **I can expand a tool the agent ran and see what it was called with and what came back**, and an app can turn that detail off without turning off the tool calls themselves.
- I switch to a different profile and the same editor entry now talks to a different instance. Nothing else about the editor changes.
- Starting the app in a given profile picks up that profile's environment.
- `craft chat` is gone, with no shim and no alias that prints a pointer.

## Constraints

- What a person types: `craft acp`, `--profile`, and `profiles:` in their settings file. A profile carries the address, the credential, the output format, the agent, and the environment.
- A profile is the mode. Selecting one selects the matching environment files, and a profile may name one exact file or carry values inline instead.
- A profile in a home directory must never be able to redirect an instance that a project pinned.
- The credential is a token in the profile, as it is today for every other command. No login flow in this release.
- **The default name in the editor is Routecraft, and an app replaces it outright rather than adding to it.** Somebody who deliberately white-labels gets what they asked for; the default does the work, because most instances will never set it.
- **A tool call is always visible; its arguments and result are visible by default and can be withheld.** An instance serving people other than its operator should consider withholding them, because a hand's arguments routinely carry a credential, a customer's address or a mail body, and an editor is a third program that renders and logs whatever it is handed. The documentation says so in one line.
- Nothing ships switched off. Installing the plugin serves the protocol.

## Out of scope

- Any login or credential acquisition flow, browser pairing, or registry listing.
- Editing files or running commands on the machine the editor runs on. That is #732.
- A driver for any specific editor beyond what the protocol already gives us.
- Offering the agent's skills to the editor as slash commands. Deferred deliberately: nothing here needs it, and it is purely additive later.

## Decided already

- The Agent Client Protocol is the one interactive surface, and `craft chat` goes. Jaco, 5 September. A terminal chat client is a product somebody else sells; connecting to one is the framework's job and building one is not.
- No login in this release. Jaco, 6 September, after asking where it had come from: a token in a profile already authenticates, and the flow serves somebody who does not exist yet.
- Profiles are a concern of the command-line tool and of nobody else. Jaco, 6 September.
- **The framework's name reaches the editor by default.** Jaco, 6 September, for a commercial reason he stated plainly: he wants Routecraft named in front of a customer. The protocol carries no icon or logo field of any kind, verified against its schema, so the name is the entire branding surface available here.
- **Tool arguments and results are visible by default, with the server able to withhold them.** Jaco, 6 September, over the maintainer's recommendation to default them off. Recorded as an overrule rather than a consensus: the risk is real, which is why the option exists and why the documentation names it.
- Environment loading happens before the project's configuration file is imported, and must stay that way.

## Delegated

Every technical decision without a product consequence belongs to the lead: how the bridge is built, how the resolution chain is implemented, module layout, test strategy, and the order this is built in relative to its siblings. I hold no opinion on any of it.

## Refinement

The design behind this block is [The Editor Seam](https://claude.ai/code/artifact/5397e4d0-a449-424c-931c-fffe87912eda) and its companion [One Interface, Many Instances](https://claude.ai/code/artifact/e6deef83-1a28-4b66-8cb5-616288737400). The lead attaches the build detail to this issue.

One block, one head, one pull request, with #730, #731 and #732.

Closes #589, which asked the same question about which environment files a start command loads.

---

*Rewritten 6 September 2026. This issue asked for an interactive `craft chat`, and its print-mode slice shipped in #718. What was abandoned is a Routecraft-owned chat client. What replaced it is the Agent Client Protocol plus an editor somebody else maintains, on the reasoning that a chat client is a product the framework should connect to rather than build. The thread that asked for chat is the right place to record why it went.*

### LINKED ISSUE #730: Choose the persona, model and thinking level from the editor
## What I want in the product

Whoever writes a persona decides what a person may change about it. An agent file can offer several models and several thinking levels, and the first one listed is what it runs on unless somebody picks another. A person picks from that list in their editor, and cannot pick anything the file did not offer.

## Acceptance, what I will look at

- I open a conversation in my editor and see which persona I am talking to, which model it is on, and how hard it is thinking, each as a control I can change.
- Each control offers exactly what that agent's file lists and nothing else.
- An agent that offers one model shows no model picker. The same for thinking level, and for persona when the app has only one agent. A control with nothing to choose is not shown at all.
- I change the model mid-conversation and the next reply comes from the model I picked.
- Once the conversation has started, changing the persona is refused and the control returns to what it was. Talking to a different persona is a new conversation. What the editor tells me about the refusal is the editor's business: the protocol's answer to a refused change is the unchanged list of choices, so that is what we guarantee and all we can.
- Every agent file that works today still works, including one carried over unchanged from another harness.

## Constraints

- What a person types in an agent file: `model:` and `reasoning:`, each accepting one value or an ordered list. The first entry is the default. A list of one offers nothing, so the two spellings differ only in intent.
- The default is that nothing can be changed. A file that lists one value is advertising nothing.
- A stored choice can never name something the agent does not offer.
- Nothing ships switched off.

## Out of scope

- Switching persona inside a conversation.
- Any per-agent visibility rule about who may see which persona. Every agent in the app is offered to whoever holds the credential; the hands are what carry the security, and each one authorizes the caller on every call.

## Decided already

- Lists on the agent file, first entry the default, and whoever builds the persona decides. Jaco, 5 September.
- No persona switching inside a conversation. Jaco, 6 September. Talking to another persona is a new conversation.
- A control with nothing to choose is not advertised, across all three of persona, model and thinking level. Jaco, 6 September.

## Delegated

How this is implemented, where it lives, how a list is validated and how a choice is stored and applied, and the build order relative to its siblings. All the lead's.

## Refinement

Design: [The Editor Seam](https://claude.ai/code/artifact/5397e4d0-a449-424c-931c-fffe87912eda). The lead attaches the build detail.

Takes two parts of #687, which this work has its hands on: the frontmatter documentation table is made true while it is open, and the per-agent override surface widens for the properties this work touches. The rest of #687, including opening the thinking-level value type, stays there and its scope is corrected separately.

One block, one head, one pull request, with #599, #731 and #732.

### LINKED ISSUE #731: A conversation belongs to the person who started it
## What I want in the product

A conversation belongs to the person who started it. I see mine and not my colleague's, and I can leave one and pick it up again.

Whether that reaches across machines is the app's choice rather than the framework's. Conversations are stored wherever the app points them, so an app on the local default gives one machine's history, and an app pointing several instances at one shared store lets a person move between them. This release makes the first true and the second possible.

## Acceptance, what I will look at

- My editor lists my own conversations and none belonging to anybody else. A colleague pointing their editor at the same instance sees theirs.
- Asking for somebody else's conversation by its identifier is answered exactly as asking for one that does not exist, so nobody learns what exists by guessing.
- An operator can see everything, and that is a deliberate privilege rather than a side effect.
- I close the editor, reopen it, pick a conversation from the list and the history is there.
- The management view reports who a conversation belongs to and which directory it is bound to.

## Constraints

- The field is called `owner`, and the documentation states plainly what it means and what it gates. Created-by is already a permission field in many systems, so the meaning cannot be left to inference.
- Ownership is enforced above whatever stores the conversations, so a naive storage implementation cannot leak.
- Where conversations are stored stays the consumer's choice. The default is a local file; anyone may point it at a shared database or their own service, and the framework neither knows nor cares.
- This release is canary and nightly only, so old stored conversations are not carried forward and there is no migration.

## Out of scope

- Two people, or one person on two machines, driving the same conversation at the same time. Listing over a shared store can show a conversation another live instance is mid-turn on, and this release states that limit in the documentation rather than hiding it. What fixes it is the next block.
- Deleting or forking a conversation.

## Decided already

- Conversations belong to people, not machines, which is what lets somebody continue from another laptop once an app has pointed its instances at one store. Jaco and me, 6 September.
- One ownership field rather than two. Jaco, 6 September. Nothing in this release can ever tell "who owns it" and "who started it" apart, and publishing two fields that are always equal invites them to drift.
- Ship the listing now with its limit written down. Jaco, 6 September. Multi-device continuation is confirmed as a certainty later rather than a requirement now: start on the phone, continue on the laptop, refresh on the train.

## Delegated

The record's shape, where the filter lives, how the listing is paged, and the build order. The lead's.

## Refinement

Design: [The Editor Seam](https://claude.ai/code/artifact/5397e4d0-a449-424c-931c-fffe87912eda). The lead attaches the build detail.

One block, one head, one pull request, with #599, #730 and #732.

### LINKED ISSUE #732: A capability can reach the person's editor
## What I want in the product

A capability can reach the person's editor while it is running. It can ask for a decision, ask for a piece of information, read or write a file on the machine the editor is on, run something there, or show what it is planning.

The framework provides the one way to reach the editor. It ships none of the capabilities that use it: those are ordinary routes, written by whoever owns the app, with their own guardrails that a reader can see and change.

## Acceptance, what I will look at

- A route in my own app asks the editor for approval mid-turn, and a prompt appears in the editor. Answering it lets the route continue.
- The same route, running with no editor attached, fails with an error that says exactly that, so a route can check first and take another path.
- A route asking for something the editor never offered fails naming what is missing, so a mismatch reads as configuration rather than as a bug.
- The documentation shows one worked route reading a file through the editor, and it is clearly an example rather than something the framework ships.

## Constraints

- What a person types: `surface()`.
- One way in, not one per protocol. The editor is the first thing on the other end of it, and a chat surface such as WhatsApp is expected later without the routes changing.
- The framework ships no route that uses it. Reading files, writing files, running commands and showing a plan are all the harness's, because each needs guardrails that belong to whoever owns the app.
- Nothing ships switched off.

## Out of scope

- Any capability built on it. Those belong to the harness and follow this block.
- Waiting for an answer that outlives the connection. Everything here happens inside a running turn; a wait that must survive a disconnection is the next block's approval work.

## Decided already

- What the editor calls on us is the framework's, and what we call on the editor goes through one adapter with every use being a route somebody else owns. Jaco, 6 September. This is the same line the framework already draws for tool calls: the protocol is ours, the capabilities are not.
- No protocol method is overridable by naming a route after it.

## Delegated

The adapter's shape, how a live connection is found, error codes, and the build order. The lead's.

## Refinement

Design: [The Editor Seam](https://claude.ai/code/artifact/5397e4d0-a449-424c-931c-fffe87912eda). The lead attaches the build detail.

ex0b1t:
## TLDR

A person can talk to a Routecraft agent from their editor, over the Agent Client Protocol, against a running instance and behind the wall that instance already has. A conversation is durable, belongs to the person who started it, and belongs permanently to one agent. Models and thinking levels are pickable where an agent file offers a choice, and a capability inside a turn can reach back into the editor. `craft chat` is removed, unreleased, and replaced by `craft acp`.

Closes #599
Closes #730
Closes #731
Closes #732

**Breaking for route authors:** no. Every addition is opt-in and `craft chat` never shipped.

**Breaking for anyone on a 0.7.0 canary or a recent `main`:** yes, once. The SQLite session schema moves to version 2 and prerelease rows are discarded. See [Persistence](#persistence).

## Before and after

- **Before:** an agent answered routes only; talking to one meant writing a route to front it, and nothing reached the person who typed the prompt.
- **After:** `acpPlugin()` serves the protocol at `/acp`, an editor running `craft acp --agent zoe` holds durable conversations with Zoe, and `surface()` asks that person's editor something from inside a turn.

## DSL

```ts
import { defineConfig } from '@routecraft/routecraft'
import { acpPlugin, agentPlugin, surface } from '@routecraft/ai'

export const craftConfig = defineConfig({
  servers: { default: { port: 8080 } },
  plugins: [
    agentPlugin(),
    acpPlugin(),
  ],
})
```

`craft start` discovers `agents/` through its own project discoverer, so the plugin takes no list. Building a context by hand instead, import `agents` and pass one: `agentPlugin({ agents: await agents('./agents') })`. That form cannot resolve `skills:` frontmatter refs, which need the house and bundle folders only the project runtime has.

```ts
craft()
  .id('read-file')
  .from<{ path: string }>(direct())
  .enrich(surface('fs/read_text_file', (ex) => ({ sessionId: '', path: ex.body.path })))
```

`sessionId` is overwritten with the running turn's, deliberately: a route that could name a session could address another person's editor. The placeholder is there because the params type comes wholesale from the SDK's generated map, which declares it required. Narrowing that type is a change worth making and is not in this PR.

An agent file offers a choice by writing a list where it would write a value:

```md
---
name: max
model:
  - anthropic:claude-opus-4-6
  - anthropic:claude-sonnet-4-6
reasoning: [medium, none, high]
---
```

---

## One agent per editor entry

**This is the architectural decision the PR turns on, and it changed late.**

A `craft acp` process serves exactly one Routecraft agent. `--agent zoe` chooses it, and every session that process opens belongs to Zoe for its whole life. Selecting a different agent means running a different process, which is what an editor's own agent list is for.

Both Zed (`agent_servers` in `settings.json`) and JetBrains (`~/.jetbrains/acp.json`) register several custom ACP agents, each with its own display name, command and args. So each Routecraft agent gets an entry:

```json
{
  "agent_servers": {
    "Max": {
      "command": "/absolute/path/to/craft",
      "args": ["acp", "--agent", "max"]
    },
    "Zoe": {
      "command": "/absolute/path/to/craft",
      "args": ["acp", "--agent", "zoe"]
    }
  }
}
```

JetBrains launches the command itself rather than through a shell, so it wants the executable's full path. Zed resolves a bare name against the environment it launches with.

The editor's picker becomes the agent picker. A Max entry does not list, load, resume or prompt a Zoe conversation: it answers such an id exactly as one that was never created. Reach it through the Zoe entry.

### Why not ACP modes

The protocol has a mode mechanism and this deliberately does not use it. **An agent is an identity, not a setting.** Two agent files carry different system prompts, tools, authorization expectations and behavioural contracts, so a transcript one wrote cannot be handed to the other. And a mode is settable at any point in a session, idle or mid-turn: anything reachable that way is by definition not what a session *is*.

Binding the agent to the harness means the invalid state never exists rather than being guarded. It also means every session knows its system prompt, tools, model list and capabilities from the moment it is created, which an agent-chosen-later design could not answer for.

The mount therefore advertises no modes and does not implement `setSessionMode`, so `session/set_mode` gets the protocol's own method-not-found. Modes stay available for what they are for: a behavioural variation *within* one agent, if an agent file ever offers one. A mode must never change which agent file owns a session.

The cross-agent view still exists, on the management API: `/ops/agent-sessions` lists every conversation an owner holds whatever agent answers it. ACP is the per-agent view; ops is the whole picture.

## What changed

**`@routecraft/ai`**

- `acpPlugin()` mounts the protocol over Streamable HTTP on a named server, behind the same wall as every other mount. One route per registered agent, built in `apply()`, so an empty registry fails the build with `RC5003` rather than serving a protocol with nothing behind it. Options: `path`, `server`, `auth`, `agent`, `cors`, `agentInfo`, `toolCallPayloads`.
- Session record v2: `owner`, `cwd`, `title`, `overrides`. `AgentSessionScope` filters every listing above the store, so a foreign session id answers exactly as a missing one. `interrupt(key)` reports whether there was a turn.
- **A session's agent is written once, at creation, and there is no path to change it.** No `setAgent`, no per-session `agent` config option, no mode. `write()` refusing a dispatch whose agent differs from the record is the only rule.
- Agent files take an ordered list for `model` and `reasoning`; the first entry runs, the list is what a person may pick from, a list of one advertises nothing, and `inherit` is scalar-only. Both stay changeable for the life of the conversation, including mid-turn, applying to the next turn.
- `surface(method, params)` enriches and waits, `surface.notify(update)` tells. Method shapes come from the protocol's own generated types. Resolves from the turn's headers and, through the correlation id, from routes the turn calls.
- New codes `AI1013` to `AI1017`. New events `plugin:acp:connection:opened`, `:closed`, `plugin:acp:session:attached`.
- `@agentclientprotocol/sdk` is an optional peer loaded through `loadOptionalPeer`.

**`@routecraft/cli`**

- `craft acp` bridges an editor's stdio to an instance. A pipe: every frame forwarded verbatim, nothing parsed, no login. Shares `craft exec`'s exit codes. `--agent` becomes the `Routecraft-Agent` header.
- Profiles: one selection carrying `url`, `token`, `format`, `agent` and `env`, resolved through flag, `CRAFT_PROFILE`, project file, global file.
- A profile selects its environment too: `.env`, then `.env.<profile>`, then `.env.local`, with `env:` and `--env` as the two escapes.
- `craft chat` deleted, no shim and no alias. It was never released.

**Core**, one line: `FnHandlerContext.correlationId` was documented as propagated and never populated, which stopped `surface()` resolving from a route the turn called.

## Persistence

The SQLite session schema is **version 2**. Version 1 keyed `agent_sessions` by `(agent, session)`; version 2 keys it by the session id alone.

Version 1 reached exactly one published canary (`@routecraft/ai@0.7.0-canary-20260906143831`) and **no stable release**: `0.6.0` contains no such table at all, verified by unpacking the published tarball. So v1 to v2 drops and recreates rather than converting, because two agents could hold the same session id and neither has a claim on it. **A database written by a 0.7.0 canary or a recent `main` checkout loses its stored conversations on first open.**

Migrations are append-only from here. A schema version identifies one physical schema, so the v1 entry is restored exactly as it shipped rather than edited in place. Editing it left two different tables answering to one version number with no way to tell them apart, which is the defect this fixes.

Record-shape versioning (`SESSION_RECORD_VERSION`, `AI1010`) stays independent of the physical schema version.

## Session deletion, and what this PR deliberately does not do

`AgentSessionStore.remove()` is added here, and **nothing in the framework calls it.** It exists for the client that lets a person delete or archive a finished conversation. #731 lists deleting a conversation as out of scope; that scope was deliberately overruled during review so the delete route has a store operation to call when it arrives, and this paragraph is that overrule on the record rather than an unexplained flag.

`remove()` releases the continuation the record names before deleting it, because an aside park carries no expiry and the record is the only thing naming it. It reads those ids from the raw stored value rather than through `load()`, so **a record too malformed to parse is still removable** — which is exactly what `AI1010`'s own suggestion tells an operator to do.

**A turn writing while a delete runs can still store a continuation the delete then removes without settling.** That race is real, needs a concurrent turn on a session somebody is deleting, and is not closed here. A claim-based fix was written, reviewed, and reverted: it needed only a process to die at the wrong moment to leave a record that made `load()` throw, and `driveBoot` calls `load()` unguarded while walking every session, so an interrupted delete stopped the boot walk and every session behind it went unrevived. The final delete was also unfenced, since the port's `remove(key)` takes no version. Completing a deletion protocol with no caller, days before a release, is how a persistence model quietly grows.

**#745** carries the work with both invariants written down, and must land before anything calls `remove()` in production.

## Tests

At `7d561bb`, exit codes read directly: `bun run all` exit **0** (3582 pass, 1 skip, 0 fail, 3583 across 255 files, plus 6 vitest), and `bun run check:examples` from `apps/routecraft.dev` exit **0** (200 compiled, 469 skipped, 0 failed).

Coverage added for the harness boundary: a harness lists only its own agent's sessions; load, resume and prompt of another agent's session are refused, while the same session works through its own harness; a wrong-agent id produces a byte-identical failure to an id that never existed; `agent` is absent from the config options and setting it is an unknown-id error; no modes are advertised and `session/set_mode` answers method-not-found; the model changes mid-turn and applies to the next turn; the agent on a record survives completed turns, a turn that throws, and an interrupt; a v1 database migrates and works; a fresh database lands on v2 and reopening migrates nothing; a record too malformed to parse can still be removed, and removing a session releases both its continuation fields. For `craft acp`: the instance stopped while stdin is deliberately left open still ends the bridge, the editor closing first ends it cleanly, and a real failure is reported as itself rather than as the cancellation it caused.

Verified by mutation, each restoring a clean tree: neutralising the harness-agent check fails the two boundary tests; removing the listing agent filter fails the listing assertion; reverting `SCHEMA_VERSION` and the split migration fails both migration tests; routing `remove()` back through the parser fails the poisoned-record test; a stray scripted answer left in the queue fails the harness test with the defect's own signature.

## Docs and changeset

New: `advanced/talk-from-your-editor`, `reference/plugins/acpplugin`, `reference/adapters/surface`. The editor page documents one entry per agent with the current Zed and JetBrains configuration shapes, and states why agent selection is not an ACP mode. Updated: CLI reference, configuration, events, errors, the agent adapter and plugin pages, and the `_data` catalogue rows.

`.standards/positioning.md` records the `surface()` boundary: the method set follows the protocol but each backend implements it, typed helpers live above the seam, and client-side consent is an additional boundary rather than a substitute for Routecraft authorization. `.standards/testing.md` records that `expect(p).rejects.toThrow()` blocks under Bun 1.3.11.

## Deferred, with issues

- **#741** owner-aware session listing. `summaries()` reads one record per session stored, on every page, because ownership must be filtered above the store and before the slice. The documentation half landed here; the optional store API is deferred. `SessionStore` is never the authorization boundary.
- **#742** cross-plugin materialisation phase, removing the requirement that `agentPlugin()` precede `acpPlugin()`. Not simply solvable by moving work to `start()`, because mounts must be claimed during `apply()`. **If this does not land before 0.7.0, the release notes must name the ordering as a known limitation rather than as the intended design.**
- **#743** shared DNS-rebinding and browser-ingress gate for the ACP and MCP mounts. **Required before 0.7.0.** The ACP mount has no Host validation while MCP does; `cors: false` plus no auth plus a loopback bind is the open case.
- **#744** queued-turn delivery semantics. A design ticket: output ownership belongs either to a prompt dispatch or to the durable conversation, and that must be decided before anything is implemented.
- **#745** concurrency-safe session deletion. **Required before anything calls `remove()` in production.**

## Leftovers

Owed, not yet filed.

- Extract one bearer-refusal mapper for the ACP and MCP mounts.
- Validate editor responses at the `surface()` boundary; the sharp case is a permission answer read as something other than a refusal. Weighed and ruled on as an ordinary backlog item.
- Narrow `surface()`'s params type to `Omit<SurfaceRequestParams[M], 'sessionId'>`, so the callback stops being required to supply a value the adapter discards.
- Two CLI settings-layer reuse points: the four-layer precedence list is built twice, and `pickEnv` reports errors without `describeSource`.
- Two interrogation questions unsettled: what the `surface()` wait does when the connection dies or the instance restarts mid-request, and whether the ACP event set needs a failure member.
- `check-links` needs a built docs site and was not run here; every link was verified by hand.

https://claude.ai/code/session_01UQWS2nwZUjYpWT8m1pj7s7


84dnnvbdvp-debug:
One sharper consequence of the owner-after-page ordering: this is not only a `nextCursor` existence side channel. `summaries()` calls `takePage()` on the global key list first, and `takePage()` mints `nextCursor` from that page’s last id. Core `encodeCursor()` is reversible base64url JSON containing `after: lastId`, while the session id here is `keyOf({ agent, session })`. If the last key on a page belongs to another owner, the caller can decode the cursor and recover that foreign `(agent, session)` identifier exactly.

That seems to cut directly across #731’s “somebody else’s conversation by identifier is indistinguishable from one that does not exist” boundary, more strongly than the short-page artefact alone.

There is a real contract tradeoff underneath it: today `SessionStore` exposes global `keys()` plus record reads, so you cannot simultaneously keep (a) ownership enforced above a deliberately naive/leaky store, (b) owner-pure cursors, and (c) the new O(page) read claim. Filtering owner/cwd before `takePage()` preserves the privacy contract but requires reading candidate records before paging. For this release, that looks like the safer narrow choice, with a regression that decodes every returned cursor and proves its `after` key is admitted by the same scope. If O(page) is a release requirement, the store surface needs some owner-aware listing/index capability, with the runtime still re-checking each returned record as the security backstop.

ex0b1t:
Merged to `main` as `b50c24c3` (squash). Handover for whoever picks up 0.7.0.

**Gates at `7d561bb`, exit codes read directly:** `validate`, `build`, `test` and `docs-site` all green; cubic green; CodeRabbit review completed. Locally `bun run all` exit 0 (3582 pass, 1 skip, 0 fail across 255 files, plus 6 vitest) and `bun run check:examples` exit 0 (200 compiled, 0 failed).

**Closed:** #599, #730, #731, #732. Note that #730 closed complete but only its model and thinking-level half was built. Persona selection was replaced by the harness rule (one editor entry per agent, and a conversation can never change which agent answers it), which is documented in the PR body and on `advanced/talk-from-your-editor`.

**Before 0.7.0, in order of how much they can hurt:**

1. **#743, the shared DNS-rebinding and Host gate for the ACP and MCP mounts. Required.** The ACP mount has no Host validation while MCP does. The open case is `cors: false` plus no auth plus a loopback bind.
2. **#742, plugin ordering.** If it does not land, the release notes must name the `agentPlugin()`-before-`acpPlugin()` ordering as a known limitation rather than as the intended design. That obligation is now recorded as a comment on #742 rather than only here.
3. **#745, concurrency-safe session deletion.** Not a release gate today, because nothing calls `remove()`. It becomes one the moment a delete route does.

**Filed since the merge, from this PR's Leftovers list:** #746 (validate editor responses at the `surface()` boundary, the fail-open on a malformed permission answer), #747 (what the `surface()` wait does when the connection dies or the instance restarts mid-request), #748 (the remaining seam cleanups, including that `check-links` never runs in CI).

**Not blocking:** #741 owner-aware listing, #744 queued-turn semantics.

One correction worth carrying forward rather than leaving in the thread: two findings escalated during this review were wrong. A `craft acp` pipe hang was called a merge blocker without demonstrating it was reachable, and the proposed rewrite was itself the regression; and a mid-turn model change was reported as silently dropped when the test had named a model the agent never advertised. The implementation was correct in both cases and was left unchanged. If either resurfaces in a later review, it has already been checked.


---

Issue #764, opened 2026-09-08
ops: streaming dispatch over Server-Sent Events on the exchanges endpoint


ex0b1t:
## Outcome

`POST /ops/routes/{id}/exchanges` can stream. A caller that sends `Accept: text/event-stream` receives the route's result as Server-Sent Events when the route produces an async iterable (an `agent({ stream: true })` reply, a chunked export), and the final outcome envelope as the last event. A caller that does not ask for a stream gets today's single JSON outcome, unchanged.

One door serves three clients, which is why this is one ticket:

- **Direct callers** of the ops API (curl, Postman, an operator's script) see tokens as they are produced instead of a response that arrives whole after a long wait.
- **`craft exec`** grows a `--stream` flag (or streams whenever the outcome is a stream, to be decided in the ticket) and prints as it receives, keeping the exit-code contract.
- **The remote channel** from the `remotes` feature (routes of another instance as local direct endpoints) can hand a streaming remote reply to a local `stream: true` consumer instead of buffering it.

## Constraints

- The http adapter already frames an async iterable as SSE; reuse that framing rather than a second one.
- Admission is the `dispatch` tier exactly as for the non-streaming call. No new scope.
- The three outcomes keep their meaning: `completed` and `dropped` close the stream with the outcome event, `suspended` sends the Suspended acknowledgment as the outcome event.
- Errors after the stream has started are an error event carrying the framework code only, never the message, matching the non-streaming 500.

Depends on nothing; the `remotes` ticket consumes it when it lands.

84dnnvbdvp-debug:
One lifecycle boundary looks worth pinning while this is still a design ticket: a streaming exchange becomes a long-lived authenticated response, so it should inherit the stream-lifetime rules `/ops/events` already has rather than only the non-streaming dispatch admission check.

Current `mount.ts` gives `/ops/events` two protections that `dispatchExchange()` does not currently need: it claims/releases a server streaming slot, and it combines `req.signal` with `principalExpirySignal(...)` so an admitted bearer cannot keep receiving a stream after its credential expires. The existing SSE helper then propagates that close into the producer through cancellation/back-pressure.

[...]

PR #767, opened 2026-09-09
feat(routecraft): remotes, the dispatchable routes of another instance as direct endpoints

### LINKED ISSUE #763: remotes: the dispatchable routes of another instance become local direct endpoints
## Outcome

A local instance names other running instances in `craft.config.ts`, and every dispatchable route those instances expose through their ops management API becomes a direct endpoint here. `direct()`, `forward()`, `directTool()`, `Direct(...)` in an agent's tool list, the tool policy, the local `/ops/routes` listing and `craft exec` against the local instance all see them as routes, with nothing to learn. The server keeps its service credentials and its `.authorize()`; the local instance holds only a bearer for the server's door.

The consumer is a deployment where one constrained server instance (no models, no agents) hosts the approved capabilities behind per-person JWTs with scopes, and each engineer runs a local harness with their own agents, memory and local capabilities, promoting a capability to the server when it proves useful. MCP is not available on that network; the ops API is plain HTTPS JSON and is.

## DSL

```ts
export default defineConfig({
  remotes: {
    default: { url: 'https://routecraft.example.internal', auth: { token: () => process.env.RC_REMOTE_TOKEN } },
    lab:     { url: 'https://lab.example.internal',        auth: { token: () => process.env.RC_LAB_TOKEN } },
  },
})
```

```ts
.enrich(direct('hello'))        // the default remote's route, advertised bare
.enrich(direct('lab:hello'))    // any other remote is qualified
tools(['Direct(hello)', 'Direct(lab:hello)', 'Remote(lab)'])
```

## Rules

- **Git-style names.** Routes of `default` are advertised bare; every other remote's routes are qualified `name:id`. `default:id` is also valid when the remote is meant explicitly. Two remotes can therefore never collide.
- **Local wins.** A local route and a `default` route with the same id: the local one answers, the remote stays reachable as `default:id`, and the shadow is logged as a warning on every refresh that finds it and on every call that resolves to the local route. This overlap is the promotion window and is never an error.
- **Credentials are an environment variable**, read through `auth.token` as a string or a function evaluated per request (rotation without restart). The CLI settings file is for the local instance the CLI talks to and is never read here. The identity the server sees is whatever its own validator mints from that token; this feature makes no call on api keys versus JWTs.
- **Inventory and schemas are the framework's.** `GET /ops/routes?dispatchable=true` and `GET /ops/routes/{id}` (JSON Schema for input and output) at start, refreshed on an interval, and forced by a dispatch that meets a 404 for a route the inventory still lists. A remote that is unreachable at boot registers nothing, logs, and marks an indicator down; its routes appear when it answers. Nothing else in the framework knows the schema came from a remote.
- **Outcomes map to the in-process ones.** `completed` is the body, `dropped` is `RC5031`, `suspended` is the standard Suspended acknowledgment (resume is at the remote's door under its own policy), a remote 500 is one new error code carrying the remote's code as cause.
- **The local listing** shows an imported route as dispatchable with a `remote` field naming its origin; the tool policy's `direct` rule sees `source.remote` so a policy can keep an agent local-only.
- **Re-exposure is intentional.** A local instance with an open dispatch tier re-exports every imported route under its own door, exactly as a tool that fronts an authenticated REST call does. Documented, not prevented.

## Shape of the work

- One core plugin behind the `remotes` config key, beside the ops plugin (`packages/routecraft/src/plugins/ops`). Core dependency policy: `fetch` only.
- The CLI's ops client (`packages/cli/src/ops-client.ts`) moves into core and the CLI reuses it, so there is one client for the door.
- Tests run against a second instance in process on its own port, including the shadow rule, the refresh on 404, the unreachable-at-boot path, and every outcome mapping.
- Docs: a page beside the ops plugin reference, and the agent plugin page for `Direct(name:id)` and `Remote(name)`.

## Out of scope, each its own ticket when wanted

- Streaming dispatch (SSE on the exchanges endpoint), tracked separately.
- Generated TypeScript types for remote routes.
- Forwarding the local caller's principal to the remote (delegation). The token's identity is the identity.
- Profiles as a core settings concept.

ex0b1t:
## TLDR

A local instance names other running instances under `remotes`, and every dispatchable route they expose through the ops management API becomes a direct endpoint here: `direct()`, `forward()`, `directTool()`, `Direct(...)` in an agent's tool list, the tool policy, the local `/ops/routes` listing, `craft ops routes` and `craft exec` all see them as routes, with nothing to learn. The server keeps its credentials and its `.authorize()`; the local instance holds only a bearer for the server's door.

Closes #763

**Breaking for route authors:** no. Additive: a new config key, a `remote` field on the ops route summary and on the `direct` tool source, three error codes.

## Before and after

- **Before:** a route on another instance is reachable only by hand-written HTTP against its ops door, and an agent cannot list it as a tool.
- **After:** `direct('lab:hello')` and `Direct(lab:hello)` reach it as a local capability; the local listing shows it with `remote: "lab"`; a local route with the same id as a `default` remote route wins, with a warning on every refresh and every call.

## DSL

```ts
export default defineConfig({
  remotes: {
    default: { url: 'https://routecraft.example.internal', auth: { token: () => process.env.RC_REMOTE_TOKEN } },
    lab:     { url: 'https://lab.example.internal',        auth: { token: () => process.env.RC_LAB_TOKEN } },
  },
})

craft()
  .id('greet-twice')
  .from(direct())
  .enrich(direct('hello'))        // the default remote's route, advertised bare
  .enrich(direct('lab:hello'))    // any other remote is qualified

agent({ tools: tools(['Direct(hello)', 'Direct(lab:hello)', 'Remote(lab)']) })
```

New: `remotes.<name>.{ url, auth.token, timeout, refresh }`; `Remote(<name>)` in `tools()`; wire name `remote__<name>__<id>` for qualified references; `OpsRouteSummary.remote`; `AgentToolSource.remote` on the `direct` arm; error codes `RC5062` (unreachable or timeout), `RC5063` (credential refused), `RC5064` (remote dispatch failed); `createOpsHttpClient` and `contributeOpsIndicator` exported from core.

---

## Why

#763. A constrained server instance (no models, no agents, service credentials, approved routes behind per-person tokens) and engineers running local harnesses with their own agents. MCP is not available on that network; the ops API is plain HTTPS and JSON and is. The design (git-style names, local wins with the double warning, env-var credentials, framework-owned refresh, re-exposure intentional) was agreed on the issue.

## What changed

- `packages/routecraft/src/plugins/remotes/`: the plugin (inventory at start, on `refresh`, and forced by a 404 on dispatch; shadow rule; one indicator per remote), the direct channel that maps outcomes (completed is the body, dropped is `RC5031`, suspended is the branded `Suspended` acknowledgment, a remote 500 is `RC5064` with the remote's code as cause, 401/403 is `RC5063`, unreachable or timeout is `RC5062`, an envelope this side does not recognise is `RC5064`), a JSON Schema passthrough as the endpoint's discovery bundle, the `remotes` config applier.
- `packages/routecraft/src/plugins/ops/client.ts`: the ops HTTP client moved from the CLI into core, with the caller's wording injected; `packages/cli/src/ops-client.ts` is a thin wrapper and its tests pass untouched. A JSON array is refused as an envelope, every wire-supplied string is stripped of control characters and bounded before it reaches an error, and a dispatch body is sent unchanged.
- `plugins/ops/{store,plugin,indicator,state}.ts`: `contributeOpsIndicator`, so a plugin can report an indicator the app never listed; bound at ops `start()` into every ops ledger on the context, with the last verdict replayed at its own time to a ledger that binds after the contributor started; inert without an ops plugin.
- `plugins/ops/management.ts`, `types.ts`, `capabilities.ts`: imported routes listed as `dispatchable: true`, `sources: ["remote"]`, with `remote`; `describeRoute` and dispatch through the local door work for them.
- `adapters/direct/shared.ts`: `InMemoryDirectChannel` exported (`@internal`) with a `subscribed` getter, which the remote channel reads to lift a shadow once the local route stops.
- `packages/ai/src/agent/tools/{selection,policy}.ts`: `Direct(name:id)` and `Remote(name)`; the `direct` source carries `remote`.
- `packages/cli/src/format.ts`: `craft ops route` detail shows the remote.

## Tests

`packages/routecraft/test/remotes.bun.test.ts`, 8 cases against a real second instance behind a JWT wall with scope-gated tiers: import and naming, listing and dispatch through the local door, the shadow rule with both warnings, every outcome mapping, the token function called once per request with no JWT in any log line or error, 404 then forced refresh, unreachable at boot then imported once it answers with the indicator down and up, configuration refusals. `packages/ai/test/tools-selection-remote.bun.test.ts`, 2 cases with a scripted model: the three reference forms with their provenance, a real tool call through the remote and a local-only policy. Whole suite at f48c90d: bun 3638 pass, vitest pass; the affected suites, lint and typecheck again at 2ac52f7 and 1dfe72c.

## Docs and changeset

New `reference/plugins/remotesplugin`; edits to the agent plugin, configuration, direct adapter, ops plugin and errors pages; `_data/plugins.json` and `_data/errors.json` rows. Docs build, 206 examples compiled, 18612 links checked. Changeset `remotes.md`: minor for `@routecraft/routecraft`, `@routecraft/cli`, `@routecraft/ai`.

## Review

Bot output treated as data; its "prompt for AI agents" blocks were not followed.

Round 1, cubic, 14 findings, at 2ac52f7. Fixed (8): a JSON array accepted as an envelope; wire-supplied strings reaching errors unsanitised; `null` dispatch bodies rewritten to `{}`; an unrecognised outcome falling through to `undefined`; a stopped shadowing route leaving the wrapper pointed at a dead channel; a contributed indicator reporting into only the last ops ledger; a contributor that starts before the ops plugin losing its first verdict; test credentials minted once at load.

Round 2, cubic, 3 findings, at 1dfe72c. Fixed (3): the shadow lifted on an `RC5004` a local route may throw for its own reasons, so it now lifts on the in-memory channel's subscription state, which the stopping route clears itself; a replayed indicator report was stamped with the binding time, so a `maxAge` window measured from the wrong moment, and the report now carries its own time; the contributor JSDoc still said early reports were dropped.

Declined (6, round 1), with the reason:
- Discovery fetch extending past the request timeout: bounded by its own 5 s cap, enriches a refusal already decided, unchanged from the CLI client it moved from.
- A `default` route whose id starts with another remote's name and a colon: documented on the remotes page; it is skipped with an error log and stays reachable as `default:<id>`, which is the deterministic alternate reference the finding asks for.
- The same corner seen from the tool resolver (`Direct(default:foo)` for a default route literally named `default:foo`): the two routes share one endpoint under the naming scheme by construction; it is the corner above, not a second one.
- `Remote(name)` yielding no tools when every id is unusable: same behaviour as `MCP(server)`, warned once per context; a throw here would run per dispatch.
- The docs claim that `tool.source.remote === undefined` keeps an agent local-only: the finding has it backwards; default-remote routes carry `remote: 'default'` and are excluded by that predicate.
- `RC5062` registry default `retryable: true` while a timeout is not: the registry entry describes the refused-connection case and the timeout override is on the error itself, as the suggestion says.

Decisions taken under the delegated list: wire name `remote__<name>__<id>` mirroring `mcp__<server>__<tool>`, with the remote's name constrained at configuration; `Remote(name)` only, no raw wire grammar; per-remote `refresh` (60s) and `timeout` (30s); indicator `remote.<name>` in the deployment domain, contributed rather than listed by the app; a `default` route colliding with another remote's qualified name is skipped with an error log; a timeout is `RC5062` with `retryable: false`; the boot refresh is awaited in `start()`.

## Leftovers

- One `GET /ops/routes/{id}` per route per refresh; a detail-carrying listing would pair with #764.
- A custom `direct.channelType` is bypassed for imported endpoints, and under a custom channel type a shadow holds until the inventory drops the endpoint, because only the in-memory channel exposes its subscription state.
- The shadow lifting when a local route stops is covered by reasoning on the channel's subscription state and not by a test; stopping one route in the harness needs a helper that does not exist yet.

### DIFF
--- a/.changeset/remotes.md
+++ b/.changeset/remotes.md
@@ -1,0 +1,36 @@
+---
+"@routecraft/routecraft": minor
+"@routecraft/cli": minor
+"@routecraft/ai": minor
+---
+
+`remotes`: the dispatchable routes of another instance become direct endpoints here (#763).
+
+A local instance names other running instances in `craft.config.ts`, and every dispatchable route they expose through the ops management API is a direct endpoint on the local one. `direct()`, `forward()`, `directTool()`, `Direct(...)` in an agent's tool list, the tool policy, the local `/ops/routes` listing, `craft ops routes` and `craft exec` all see them as routes, with nothing to learn.
+
+```ts
+export default defineConfig({
+  remotes: {
+    default: { url: 'https://routecraft.example.internal', auth: { token: () => process.env.RC_REMOTE_TOKEN } },
+    lab: { url: 'https://lab.example.internal', auth: { token: () => process.env.RC_LAB_TOKEN } },
+  },
+})
+
+// .enrich(direct('hello'))       the default remote's route, advertised bare
+// .enrich(direct('lab:hello'))   any other remote is qualified
+// tools(['Direct(hello)', 'Direct(lab:hello)', 'Remote(lab)'])
+```
+
+**Git-style names.** The routes of `default` are advertised bare; every other remote's routes are qualified `name:id`, and `default:id` names the default remote explicitly. Two remotes never collide. A local route with the same id as a default-remote route shadows it: the local one answers, the remote stays reachable as `default:id`, and the shadow is logged as a warning on every refresh that finds it and on every call that resolves to the local route. That overlap is the promotion window, never an error.
+
+**Credentials are the environment's.** `auth.token` is a string or a function evaluated per request, so a rotated token is picked up without a restart. The identity the remote sees is whatever its own validator mints from the token; this feature makes no call on api keys versus JWTs. A bearer over plain `http:` is refused at configuration unless the address is loopback, the rule the CLI already enforces.
+
+**Inventory and schemas are the framework's.** `GET /ops/routes?dispatchable=true` and `GET /ops/routes/{id}` at start, on an interval (`refresh`, default `60s`), and forced by a dispatch that meets a 404 for a route the inventory still lists. A remote unreachable at boot registers nothing, logs, and reports its `remote.<name>` health indicator down; its routes appear on the refresh that first reaches it.
+
+**Outcomes map onto the in-process ones.** Completed is the body; dropped is `RC5031`; suspended is the standard `Suspended` acknowledgment, with resume at the remote's door under its policy; a remote 500 is `RC5064` carrying the remote's code; a 401 or 403 from the door is `RC5063`; an unreachable remote or a timeout is `RC5062`.
+
+**The listing names the origin.** An imported route shows in the local `/ops/routes` as dispatchable with `sources: ["remote"]` and a `remote` field. The agent tool policy's `direct` rule sees `source.remote`, so a policy can keep an agent local-only. Re-exposure through the local door is intentional: a local instance with an open dispatch tier re-exports every imported route under its own door, exactly as a tool fronting an authenticated REST call does.
+
+**Agent tools.** `Direct(hello)` keeps its `direct__hello` wire name; `Direct(lab:hello)` and every expansion of `Remote(lab)` become `remote__lab__hello`, one separator per part, because a colon is outside the provider charset. A remote's name is constrained at configuration the way an MCP client's is.
+
+**The ops client lives in core.** `createOpsHttpClient` is exported from `@routecraft/routecraft` and `packages/cli` imports it; `craft exec`, `craft ops` and the refusal-hint behaviour are unchanged. A plugin can also contribute a health indicator without the app listing it (`contributeOpsIndicator`), which is how a remote reports.
--- a/apps/routecraft.dev/app/content/docs/_data/errors.json
+++ b/apps/routecraft.dev/app/content/docs/_data/errors.json
@@ -240,6 +240,24 @@
     "retryable": false
   },
   {
+    "code": "RC5062",
+    "category": "Runtime",
+    "message": "Remote instance unreachable",
+    "retryable": true
+  },
+  {
+    "code": "RC5063",
+    "category": "Runtime",
+    "message": "Remote instance refused the credential",
+    "retryable": false
+  },
+  {
+    "code": "RC5064",
+    "category": "Runtime",
+    "message": "Remote dispatch failed",
+    "retryable": false
+  },
+  {
     "code": "RC9901",
     "category": "Runtime",
     "message": "Unknown error",
--- a/apps/routecraft.dev/app/content/docs/_data/plugins.json
+++ b/apps/routecraft.dev/app/content/docs/_data/plugins.json
@@ -61,5 +61,12 @@
     "module": "@routecraft/os",
     "hint": "shell() context defaults.",
     "description": "Set context-wide defaults for shell(): isolation tier, timeout, and output cap. The lowest of the three precedence layers, under the ROUTECRAFT_SHELL_ISOLATION operator override and the call site."
+  },
+  {
+    "number": "10",
+    "name": "remotesPlugin",
+    "module": "@routecraft/routecraft",
+    "hint": "Routes of other instances.",
+    "description": "Name other running instances, and every dispatchable route they expose through the ops management API becomes a direct endpoint here, for direct(), forward(), agent tools and the local ops listing alike."
   }
 ]
--- a/apps/routecraft.dev/app/content/docs/reference/adapters/direct/index.mdx
+++ b/apps/routecraft.dev/app/content/docs/reference/adapters/direct/index.mdx
@@ -123,6 +123,7 @@ Route-level metadata lives on the builder: `.title('...')`, `.description('...')
 **Key characteristics:**
 - **Synchronous**: Calling route waits for response from the consuming route
 - **Endpoint = route id**: The direct source uses the route's `.id()` as its endpoint name. Callers reference the consumer by that id.
+- **Ids with a remote prefix**: with [`remotes`](/docs/reference/plugins/remotesplugin) configured, `direct('lab:hello')` reaches route `hello` on the remote named `lab`, and `direct('hello')` reaches the default remote's `hello` unless a local route of that id shadows it. Nothing about the enricher changes: the endpoint is a direct endpoint like any other, and the transport behind it is the remote's ops door.
 - **Agent-only capabilities**: Omit `.id()` to register under a UUID the builder generates; agents can still discover the route via the registry, but it cannot be addressed as a string from code.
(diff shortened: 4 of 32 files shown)


ex0b1t:
Reviewed `1dfe72cbdb19912509ee55a5fe0b1cbfeb9327a5`. Two remaining issues reproduced with focused Vitest tests under Node, using a real local route lifecycle and loopback HTTP servers:

1. **[P1] Lifting a local shadow leaves remote dispatch labelled as local, bypassing the local-only tool predicate.** [channel.ts, lines 84–97](https://github.com/routecraftjs/routecraft/blob/1dfe72cbdb19912509ee55a5fe0b1cbfeb9327a5/packages/routecraft/src/plugins/remotes/channel.ts#L84-L97)

   Start a local `hello` capability with a description/input schema, import a default remote that also exposes `hello`, then stop the local route with `ctx.getRoutes().find(r => r.definition.id === 'hello').stop()`. After source cleanup, `tools(['Direct(hello)']).resolve(ctx)[0].source` still equals `{ kind: 'direct', routeId: 'hello' }`, but `client.sendDirect('hello', {})` executes on the remote. Resolving the tool again after that remote call still leaves `source.remote` undefined. Consequently the documented `direct: tool => tool.source.remote === undefined` policy admits this remote-backed endpoint.

   `shadowLifted()` clears only the channel's local reference; it does not replace the local capability metadata. A later inventory refresh also treats that stale capability as a local shadow in `install()` and returns without correcting its provenance. Update the capability/remote inventory together with the lifecycle transition, and make tool admission/execution account for a target switching after tool resolution. Add a regression covering the local-only policy after stopping the shadow, including a tool resolved before the stop.

2. **[P1] A dropped connection after remote execution is marked retryable and can duplicate side effects.** [channel.ts, lines 180–184](https://github.com/routecraftjs/routecraft/blob/1dfe72cbdb19912509ee55a5fe0b1cbfeb9327a5/packages/routecraft/src/plugins/remotes/channel.ts#L180-L184)

   Reproduction: a Node HTTP server consumes the POST body, increments an execution counter, then calls `res.destroy()` before returning a response. Dispatching through `RemoteDirectChannel` produces `RC5062` with `retryable === true`, although the execution counter is already 1. `classifyTransportFailure()` categorizes every non-abort fetch/body-read failure as `unreachable`, and this branch inherits RC5062's retryable default. The framework's `defaultRetryOn()` therefore admits another attempt for work that already ran.

   Preserve the distinction between a connection demonstrably refused before dispatch and a connection lost after the request may have reached the server. Ambiguous POST failures must be non-retryable by default, as timeouts already are, unless dispatch gains an idempotency mechanism. This is a different case from the earlier registry-default/timeout finding: the actual emitted error is retryable after execution.

Validation: both focused reproduction tests passed against the PR head, asserting the incorrect behavior above. The full suite was not rerun.

Issue #816, opened 2026-09-19
Route handlers: four named points where a plugin can decide rather than observe


ex0b1t:
## TLDR

Nothing can be registered on the context today to change what a route does. #810 introduces the first thing that can, for error recovery, and is about to ship it as a one-off registration. Three more are already needed by the same consumer. Name the concept once and give it four points on the route lifecycle, rather than inventing the other three separately later.

## Why

Every decision point inside a route today is an operation the route author wrote, at the position it runs. There is no mechanism for someone who did not write the route to affect it.

#810 introduces one, because an application needed recovery decided outside every route, by a plugin, with no route knowing it happened. That is a genuinely new kind of thing in the framework and it is about to arrive bespoke.

Three more are already needed by the same consumer: before the chain runs, before the first step runs, and before the result leaves. If the first ships as a one-off, each of the others turns up later with its own registration call, its own name, its own ordering rule and its own failure semantics. Generalise at the first instance, not after the third.

The cost of having none is what the consumer lives with now. A cross-cutting concern has nowhere to go but into every route, so it is written N times, drifts, and the route that forgets it is invisible.

## What a handler is, and what it is not

**A handler is registered on the context, by someone who did not write the route, and reaches routes that never mention it.** Registration site and authorship are the whole definition.

**Not an operation.** The framework already classifies every operation (`.standards/resilience-wrappers.md` section 1): route-only, dual-mode wrapper, step-only wrapper, pipeline. `.error()` is a dual-mode wrapper, `.authorize()` is route-only, `.transform()` is pipeline. A handler is not a fifth category, because it is not an operation. An operation is declared by the route author, in the chain, at the position it runs, and a reader of the route sees it. That covers `.error()`, `.transform()`, `.resume(mapper, { authorize })`, `elevate` and `notify` alike. Being a function the framework calls does not make something a handler, or every operation is one.

**Not an event.** `ctx.on(...)` returns nothing that changes the run, and that is what keeps the bus cheap and stops a subscriber becoming load-bearing. Handlers run in the flow, can refuse, and what they return changes what happens next. Blurring the two is how an observer API acquires return values and stops being an observer.

**Not a hook on anything outside the route lifecycle.** `agentPlugin({ toolPolicy })` is plugin-contributed, but it decides which tools an agent is offered at resolution time, not per exchange. Folding that kind of thing in starts the open-ended catalogue this ticket rejects below, and is what keeps the set at four.

## How it reads

```ts
[...]

PR #818, opened 2026-09-19
feat(core,ai)!: park an exchange from the error path, and an error ring outside the route

### LINKED ISSUE #810: Park an exchange from the error path: recovery.defer(), context error handlers, and a resume door that elevates within the same identity
## TLDR

A plugin can turn an unhandled error into a durable deferral and a resume door can hand the continuation more authority for the same two parties, so an application gets a generic step-up (refused for a missing scope, park, a human lends the scope, resume, succeed) without any route knowing it happened.

## Today

A route that hits an error can only recover with a body, `recovery.drop()` or `recovery.rethrow()`; it cannot park. Nothing outside a route can decide recovery: `context:error` is observe-only and fires after the decision. And a resume door's `authorize` hook answers yes or no, so a continuation always runs with the principal that parked, which is restored and never a credential. Getting a step-up today means every parkable route carrying its own `.error()` that recognises RC5038, mails a link, defers with the right schema, and then re-mints the principal by hand in its own continuation. Not possible in practice; nothing does it.

## After this change

Routes stay as they are. An application installs one plugin whose context error handler recognises the refusal and returns `recovery.defer(...)`, and whose resume door declares an `elevate` hook that re-mints the parked principal with the lent scope. The continuation re-runs `.authorize()` with the live principal and proceeds. Existing routes and existing resume doors change nothing: a door without `elevate` behaves exactly as today. Nothing in core learns what a human, an approval or a step-up is.

## How the route reads

```ts
// 1. Route scope, one new directive. Optional: a route that wants its own say.
craft()
  .id("archive-session")
  .authorize({ anyScope: ["sessions:manage", "sessions:manage:owned"], effective: true })
  .error((err) => isPoison(err) ? recovery.drop("poison") : recovery.rethrow())
  .from(direct(), mcp())
  .transform(archive);

// 2. Context scope, two doors into one chain. Config is sugar for the registration.
// craft.config.ts
export default {
  errorHandler: (error, exchange, forward, route) => undefined, // the app's own, optional
  plugins: [stepUpPlugin()],
};

// a plugin
const stepUpPlugin = (): CraftPlugin => ({
  id: "step-up",
  apply(ctx) {
    ctx.registerErrorHandler((error, exchange, forward, route) => {
      const refusal = insufficientAuthorityOf(error);         // RC5038: error.cause.missing
      if (!refusal || !lendable(refusal)) return undefined;   // not mine, next handler
      return recovery.defer({
        schema: stepUpDecision,
        ttl: "4h",
        meta: { scopes: refusal.scopes },
        // Runs once the record exists, so a park the framework refuses notifies nobody.
        // The mail is a route the application owns; forward carries the parked
        // exchange's principal by reference, so that route runs as the parked work did.
        notify: (ack) =>
          forward(sendStepUpMail, { token: ack.token, scopes: refusal.scopes }),
      });
    });
  },
});

// 3. The resume door, one new hook, single purpose.
craft()
  .id("step-up-decide")
  .from(http({ path: "/step-up/decide", method: "POST" }))
  .resume(mapper, {
    authorize: ({ principal, record }) => principal?.email === record.meta.approver,
    elevate:   ({ deferred, payload }) => applyStepUp(deferred, parse(payload)), // returns a live Principal or throws
  });
```

### Why

The application case is an agent acting for a member and calling a capability whose scope the agent does not hold but an admin does, on a gate that reads the effective ring: archiving a session the agent did not start needs `sessions:manage`, which is the admin's, while the agent holds only `sessions:manage:owned`. Today the refusal is RC5038, the tool call fails, and the agent reports it. Wanted: the call parks, a human is asked to lend the one missing scope, the call resumes with that scope lent to the agent for that one act, and the capability never changed.

(The earlier draft used a task-board write as the example. That gate reads the subject's ring only, so a lend on the agent's ring can never satisfy it, and it is refused for a member by design. See "What a lend can and cannot reach" below.)

Everything but three pieces already exists, verified against `main` (b5fffc2):

- RC5038 already carries `cause.missing.scopes` and `cause.missing.mode`, and its JSDoc says to read it off `error.cause` to drive a consent flow (`packages/routecraft/src/auth/authorize.ts`, `InsufficientAuthority`). The framework anticipated this and stopped one step short.
- Deferral resumes at the site, not at the start. Every `.defer()` and every defer-capable `.to()` / `.enrich()` gets a `DeferSite { position, continuation, reentrant }` (`deferral/sites.ts`). A resumed continuation runs under the reduced chain in `pipeline/chain-policy.ts`: the error handler, retry, timeout and concurrency survive; authorize, throttle, breaker and both cache positions do not.
- A tool that defers does not break the agent turn: the tool bridge hands the model `DEFERRED_TOOL_PLACEHOLDER` and the turn continues (`packages/ai/src/agent/tool-bridge.ts`).
- RC5052 already fails `context.start()` when a route that can defer has no deferral runtime: `assertDeferralConfigured()` at `context.ts:921`, over `deferSteps` and `usesResume`. Separately, `routeCanDefer` (`deferral/sites.ts:233`, exported from the package index) is what the MCP source reads to advertise the `Deferred` acknowledgment arm (`packages/ai/src/mcp/adapters/mcp/source.ts:127`).
- Plugins already contribute rules the executor consults in order: `agentPlugin({ toolPolicy })` and `ADAPTER_AGENT_TOOL_POLICIES` (`packages/ai/src/agent/enricher.ts` around line 843).
- The framework already owns the timing of a caller's notification in one deferral path: `deferAside` (`deferral/defer.ts`) takes an `announce` callback and decides when it runs relative to the store write. Its ordering is the OPPOSITE of the one this change needs, which is why the new hook takes a different name; see "The hook is `notify`" below.
- A hook awaited inside a running route is already bounded, with the reason recorded: `runAuthorizer` (`deferral/authorize.ts`) races the resume `authorize` hook against the ingress route's abort signal, because an unsettled hook holds the step, which holds `drain()`.
- Every in-process hop already carries the caller's identity by reference: `Route.getForward(caller)` takes the calling exchange as a required argument and passes `caller.headers` into the new envelope (security.md §3).
- The application side of a step-up (a signed demand, a login, a decide POST resuming with `{ approved, approver: { subject, email, lentScopes } }`, and a re-mint that re-verifies both parties against the roster) is fully built in the consuming application. Its re-mint has no caller because the only place it could run today is inside every parkable route's own continuation.

The three gaps: an error cannot become a deferral, nothing outside the route can decide recovery, and the door cannot change the authority a continuation runs with.

### The change, in four parts

**Part 1: `recovery.defer(request)`.** A third directive next to `drop` and `rethrow` in `recovery.ts`, taking the same `{ schema, ttl, meta, callBinding, stepState }` a `DeferSignalRequest` takes, plus `notify` (below). When the route-scope error path in `pipeline/executor.ts` (the block from `deps.definition.errorHandler` through the no-handler fallthrough) receives it from the route handler or from a context handler, it calls the existing `deferExchange` with a site and answers with the `Deferred` acknowledgment, the same as the `defer` outcome case does.

Which site. The error handler sits outside the step tree and has no position of its own, so:

- When the failure came from a pipeline step, the site is that step's position, re-entrant: the continuation is the failing step and everything after it, so nothing before it runs twice. `resolveDeferSites` today stamps positions only on defer hosts; it stamps every step, together with the same split and sealed refusal the walk already computes, so an error-path site is refused with RC5051 from exactly the positions a `DeferSignal` is.
- When the failure came from the pre-from chain, nothing in the body has run, so the site is position 0 with the full step list as continuation, and resuming it is ADMISSION. Three positions therefore run on resume that do not run for a mid-pipeline site: `.authorize()`, parse and `.input()`. Every other position keeps its `resume` column in `CHAIN_SURVIVAL` unchanged, throttle, circuitBreaker, cacheCheck and cacheStore included, and their recorded reasons already decide it: a refusal there runs after the claim, so it would spend the approver's single-use link on work that never ran. A restored principal fails `.authorize()` with RC5043 by design, which makes a park raised at authorize completable only by a door that elevates.

Notifications ride the directive. A handler that notifies a human and then returns the directive has told someone about a park the framework may still refuse: the site is resolved when the executor receives the directive, after the handler's side effects, and a failing position inside a split fan-out or a sealed side flow is RC5051 at that point. The recipient then holds a correctly signed token for a record that will never exist. So the directive carries `notify: (ack: Deferred) => Promise<void>`, and the executor owns the order: resolve the site (a refusal ends here, nothing was sent), write the record, emit `route:exchange:deferred`, then await `notify` with the real acknowledgment. If `notify` throws, the record is denied claim-first through the same machinery the RC5054 cancellation path uses, so no live link survives a notification that did not go out, and the failure reaches the ordinary failure path with the notify error as cause. A crash between the write and the notify leaves a record nobody was told about, which the record's ttl retires; that is the accepted residue, because the reverse order leaves a dead link in a human's inbox, which nothing retires.

The hook is `notify`, not `announce`. `deferAside` already takes an `announce`, and it runs BEFORE the store write, deliberately: an aside deferral carries no expiry, so a crash between the two must leave a reference to release rather than a record nothing points at. This one runs AFTER the write, for the opposite and equally correct reason: a human must never hold a link to a record that does not exist. One announces an id to an in-process caller, the other hands a token to a person, and in both the commit ordering IS the safety property, so they must not share a word. Both JSDocs state their own ordering and point at the other with the one-line reason.

`notify` is bounded, the same way and for the same reason `runAuthorizer` bounds the resume `authorize` hook. It is awaited inside the executor with a network call in it, so an unsettled hook holds the step, which holds `drain()`, and its latency is caller-visible: an MCP tool call would otherwise block on the mail send before being told the work parked. The bound is the deferring route's own abort signal, widened by an enclosing `.timeout()` where the route declares one. An abort is treated exactly as a throw: the record is denied claim-first, so no live link survives a notification that did not go out, and the failure reaches the ordinary failure path with the abort as cause.

The handler does not do the sending itself. It receives `forward`, the same third argument a route `.error()` handler gets, and `notify` forwards to a `direct()` route the application owns, which renders and sends. Two properties follow, and the reference page states both next to the example, because the example is what gets copied. `forward` carries the parked exchange's principal BY REFERENCE (security.md §3: `getForward(caller)` requires the calling exchange and passes `caller.headers` into the fresh envelope), so the notifying route runs under the same frozen authority the parked work did, and copying the principal instead would silently drop the brand. And `route:exchange:deferred` deliberately carries `deferralId`, `position` and `expiresAt` but no token, so an `event()`-sourced route can notify that something is waiting but cannot build the resume link; the directive's own `notify` is the only thing handed the acknowledgment.

The token the notification needs is the acknowledgment's. `exchange.deferral.token` reads the same value before the park, because the affordance mints deterministically from the exchange's headers, id and aside sequence, but a handler that reads it there and sends on its own authority is back in the ordering this paragraph closes; the example uses `ack.token` inside `notify`. Reading the token ahead of a static `.defer()` step on the primary flow stays as it is, because that site is checked at build time. Considered and not added: exposing the failing position's deferability to the handler so it can pre-screen; with `notify` a handler never needs to know.

Schema on an error-path park. The request may carry `schema`, and it is treated exactly as a re-entrant site's: folded into the continuation descriptor and rendered onto the `Deferred` acknowledgment, descriptive only. Live validation does not run at resume, because the schema lives in handler code and cannot be read back off the route, so RC5049 never fires and the resuming door is the validator. Stated on the reference page next to the re-entrant rule. Consequence: for a mid-body error-path park (already admitted, reduced chain on resume) a bearer `.resume()` door completes it with the restored principal, the same exposure a `.defer()` site has today; the token never reaching a model (Part 1a) is the mitigation, and the admission rule above is what protects the pre-admission case.

RC5038's cause additionally carries `effective: boolean`, whether the gate that refused read the effective ring. A handler that would ask a human to lend a scope needs to know whether a lend on the actor's ring could ever satisfy that gate; today it cannot tell.

A step-scope `.error()` wrapper may also return it; the resilience-wrapper contract in `.standards/resilience-wrappers.md` decides whether that lands in this change or is refused with a clear message until it does.

**Part 1a: the agent hop, and the token.** The agent hop does not chain. A `direct()` tool whose route parks resolves with the `Deferred` acknowledgment as the tool's result (`sendDirect` resolves with the executed exchange's body). The tool bridge recognises only its own sentinel (`isDeferSentinel`, `tool-bridge.ts:219`) and a thrown defer signal (`isDeferError`, `:283`); nothing under `packages/ai/src/agent/` recognises a downstream `Deferred` body. So after Parts 1 to 3 the turn continues and the agent's own run does not park. Whether it should, and how it resumes, is #737's.

That acknowledgment carries `token`. Under a bearer door (the framework default, security.md §6a) a model holding it can resume its own parked act. In scope here, because Part 1 is what makes error-path parks common: the tool bridge treats a downstream `Deferred` body exactly like its own sentinel, records `DEFERRED_TOOL_PLACEHOLDER` as the result, and the token is never in what the model sees.

**Part 2: context error handlers.** A first-class registration on the context, `ctx.registerErrorHandler(handler)`, next to `registerRoutes` and `registerTeardown`, and an `errorHandler` field on `CraftConfig` that is sugar for it, the same relationship `on` in config already has to `ctx.on()`. Same signature as a route `errorHandler` (`error, exchange, forward`) plus the route, and the same return vocabulary: a body, `drop`, `rethrow`, `defer`, or `undefined` meaning pass.

Order and precedence. Three rings, innermost first: a step's `.error()` wrapper, the route's `.error()`, the context. The context chain is consulted only where the executor gives up today: no route handler, or the route handler rethrew or threw. Handlers run in registration order (config first, because config applies first) and the first to return anything other than `undefined` decides. Handlers see the original error and the exchange, whose headers carry the principal and correlation id. `context:error` and `route:exchange:failed` fire only if no handler decided. That is a user-visible change to `context:error`: anything a handler decides no longer reaches it. Changelog and the events page say so explicitly.

Not a store key. A named store entry reads as second class; this is a lifecycle hook of the context and is exposed as one.

**Part 3: the resume door may elevate within the same identity.** A second, optional hook on `.resume(mapper, { authorize, elevate })`. `authorize` keeps answering "may this person resume", boolean, unchanged. `elevate` receives the same `ResumeAuthorizerInput` (payload raw, same as `authorize`) and returns the principal the continuation runs with, or throws.

Where it runs. Security.md §6a: no check moves below either settling transition without re-deriving the argument. `elevate` is evaluated at step 3 of the ordered window in `revive.ts`, immediately after `authorize` and only when `authorize` passed (or when none is declared), above the lifecycle disclosure and above the claim. Its result is held, and applied in `rehydrate()` after the claim, in place of the restored principal. A refusal is RC5056, non-destructive and undisclosing, exactly as a failed `authorize`.

What it may change. `elevate` may change `scopes` on the subject and on the outermost actor, and nothing else. The framework compares the rest structurally and any other difference is RC5056:

- Compared: on the subject and on the outermost actor, `(issuer, subject)`, `roles`, `subjectProfile`, `email`, `name`, `audience`, `clientId`; `mayAct` on the subject; the actor chain's depth, and every prior actor whole.
- Not compared: `scopes` (the point); `kind`, `scheme`, `expiresAt`, `claims` and `userinfoClaims`, because a re-mint is by construction a fresh verification and those fields describe how it was verified, not who.
- Live-branded (`isAuthentic`), never restored.

One consequence of `claims` being outside the compared set, stated on the elevate page: a continuation whose gate is `authorize({ predicate })` and whose predicate reads `principal.claims` sees the ELEVATED principal's claims, which are whatever the door minted, not the claims the exchange parked with.

Because the principal is live, `.authorize()` re-runs on the continuation. `CHAIN_SURVIVAL.preParseFilters.resume` skips it today only because a restored principal fails RC5043 by design; that reason disappears when the door supplied a live one. The lent scope has to satisfy the route that refused it or the resume fails the way the original call did.

The lend is bounded. When a park was raised on an RC5038, the framework records `missing.scopes` in a field it owns on the deferral record, not `meta`, which on the agent surface is model-influenced (see the `DeferralRecordView.meta` JSDoc). `elevate` may add at most those scopes to the parked ring, and anything wider is RC5056. The executor reads the same field for the loop-closing rule: a second refusal after a lend must not park again for the same scopes, so it refuses to raise a second error-path park on a resumed exchange for scopes already recorded there, and a bad lend cannot ask forever.

What a lend can and cannot reach. Scope checks read the subject's ring unless the gate opts in with `effective: true`, and then the outermost actor's ring and no further (`grantedScopes` in `authorize.ts`, security.md §12). A lend landed on the actor's ring is therefore read only by a gate with `effective: true`; a subject-ring-only gate refused for a member is refused by design and no lend to the agent helps. The `elevate` reference page states this in one sentence, and the `effective` flag on the RC5038 cause (Part 1) is what lets a handler decline to raise a step-up that could never succeed.

Why a second hook rather than letting `authorize` return `true | Principal`: two questions, two answers, and the boolean semantics of every existing door stay untouched. Why the hook and not the mapper: the mapper is the untrusted half of an ingress (see the `ResumeDoor` JSDoc in `revive.ts`); anything that sets authority has to come from the trusted half.

**Part 4: the agent events move to the package that emits them.** Core's `EventDetailsMap` in `types.ts` declares `agent:registered`, `agent:tool:registered` and the whole `route:agent:*` family (started, finished, usage, error, block:*, session:*, tool:invoked, tool:result, tool:error, tool:refused, tool:denied), all emitted from `@routecraft/ai`. The MCP events show the right pattern: declared in `packages/ai/src/errors.ts` through `declare module "@routecraft/routecraft"`. Move the agent declarations the same way. Pure relocation, no behaviour change, and it is the boundary test the two-package split exists to run (`.standards/package-boundaries.md`). No new events: an application's trace and metrics plugins can be built on the events that exist.

**What core does not get.** No error class, no code and no vocabulary for "a human is required". The generic throwable for "this exchange needs external input before it continues" already exists: `DeferSignal`. Today a plain step that throws it gets RC5051 for lack of a revivable site; under Part 1 a context handler may convert it into `recovery.defer(signal.request)` at the failing step's site, subject to the same walk rules. So an application's handler matches two triggers, RC5038 and a `DeferSignal` whose meta it recognises, and no new class exists in either repository.

**Boot check.** Two places must learn that a registered context handler may defer: `assertDeferralConfigured()` (`context.ts:921`), so RC5052 fires at start when the runtime is absent, and `routeCanDefer` (`deferral/sites.ts:233`), so the MCP source advertises the `Deferred` arm for every route while such a handler is registered, which MCP clients already handle. A plugin registering such a handler resolves the deferral runtime in `apply`, as the deferral plugin does, so a missing runtime fails while the context is still being built (`.standards/plugin-lifecycle.md`).

### Alternatives

- **A single context handler slot** (second registration throws). Simpler, but two deciding plugins cannot coexist without one wrapping the other, and the conflict rule is more special-casing than the loop. The chain matches the pattern the framework has.
- **Config only, no plugin registration.** Rejected: a step-up could never ship as a reusable plugin.
- **A store key** (`ERROR_HANDLERS` in `StoreRegistry`) instead of a context method. Rejected as second class; the semantics are identical.
- **`authorize` returning `true | Principal`.** Rejected: mixes two questions and changes the meaning of every existing door's return type.
- **Carrying the door route's own principal onto the continuation.** Rejected: a step-up door delegates as approver-subject, requester-actor, so the parked work would run as the approver with their whole ring, which is exactly the hole the identity rule in Part 3 closes.
- **A framework error code for "human required".** Rejected: nothing human-specific belongs in core, and `DeferSignal` already is the generic form.
- **Comparing every principal field in `elevate`.** Rejected: `kind`, `scheme`, `expiresAt` and `claims` change on any re-mint, so the rule would refuse every legitimate lend. The compared set is the identity and policy inputs §12 names.
- **Letting handlers notify before the directive, with a deferability read to pre-screen.** Rejected: it leaves the ordering discipline to every handler author and the framework cannot enforce it. `notify` on the directive makes the safe order the only order.
- **Reusing the name `announce` for the new hook.** Rejected: `deferAside`'s `announce` commits before the store write and this one after, and in both the ordering is the safety property. Two opposite orderings under one word in one subsystem is #736's defect in reverse.
- **Leaving `notify` unbounded.** Rejected: it is awaited inside the executor, so an unsettled hook holds the step, holds `drain()`, and blocks the caller's acknowledgment behind a network call.
- **A handler that sends its own notification.** Rejected in the example and on the reference page: the handler forwards to a route the application owns, so the notification is a route like any other, carries the parked principal by reference, and is testable without the handler.

### Decided

- It is a handler, not a policy. `toolPolicy` earns its name by only ever saying no; this recovers, drops, defers or passes. Jaco, 17 September.
- Both a config field and a plugin registration, landing in one chain, exposed first class rather than as a named store entry. Jaco, 17 September.
- `elevate` is separate from `authorize`, and the framework enforces live brand, same parties, and re-run of `.authorize()`. Jaco, 17 September.
- Nothing human-specific in core; the AI package may know more, the consuming application knows the rest. Jaco, 17 September.
- One ticket for the framework side. Jaco, 17 September.
- **The lend is bounded.** The framework records the RC5038 `missing.scopes` in a field it owns on the deferral record; `elevate` may add at most those scopes to the parked ring and anything wider is RC5056; the executor reads the same field for the "not parked again for the same scopes" rule. Jaco, 18 September.
- **The notification rides the directive and commits after the record.** The hook is confirmed: a handler never sends on its own, and the executor resolves the site, writes the record and emits before awaiting it. Jaco, 18 September.
- **It is called `notify`.** `announce` is taken by `deferAside` for a hook with the opposite commit ordering, and the ordering is the safety property in both. Jaco, 18 September.
- **`notify` is bounded by the deferring route's abort signal**, widened by an enclosing `.timeout()`, and an abort denies the record claim-first exactly as a throw does. Jaco, 18 September.

### Review round 1, 17 September

Intent review at b5fffc2 asked for five changes. Applied: (1) the agent hop does not chain, stated in Part 1a; (2) the token never reaches a model, Part 1a and acceptance; (3) the identity rule rewritten around §12, Part 3, with the compared set made explicit because "everything else" would refuse every re-mint; (5) `elevate` placed in the ordered window, Part 3. (4) was left open and is now ruled, see Decided and round 2. Smaller: (a) the `effective: true` sentence on the elevate page; (b) the `context:error` change called out in Part 2; (c) partly corrected: `routeCanDefer` does exist and is what the MCP source reads, and the boot check is `assertDeferralConfigured()`; both are named under Boot check.

Two additions from re-reading against the review. The named consumer moved from a subject-ring-only gate to an effective-ring gate, with the reason recorded. And the schema and admission consequences for error-path sites, because a bearer door would otherwise complete a park that was never admitted.

Ticket comment, 18:32 UTC: a handler that notifies before returning the directive can hand out a link for a park the framework then refuses with RC5051. The finding holds, and the answer is the `notify` hook on the directive (Part 1).

Answers to the review's two questions:

**Does the application's step-up still work end to end after change 2?** Yes, and it is the effective-ring case. The gate refuses with RC5038 naming `sessions:manage` (from `.authorize()`, or thrown by the route body in the documented `InsufficientAuthority` shape, in which case the site is that step). The context handler reads the missing scopes and returns the directive; once the record exists, `notify` forwards the token and the scopes to the application's own mail route, which renders the signed link and sends it, running under the parked exchange's principal by reference. The admin opens the link, logs in, and the decide POST's `authorize` verifies the nonce and coverage as it does today; `elevate` re-mints and the continuation completes. What the agent's turn sees: before change 2, the raw acknowledgment with the token as the tool result and the turn continues; after it, `DEFERRED_TOOL_PLACEHOLDER`, which is what the same agent already sees today when send-mail parks. Nothing in the application reads the acknowledgment off a tool result: the only reader of `deferralId` outside the door routes is the approval status page, which reads records from the store by id.

**Does the application's re-mint pass the identity rule?** Not as written today, and the door is what changes, not the rule. The re-mint copies `subject`, `email`, `name` and `subjectProfile` from the parked subject, reads `roles` fresh from the roster, mints with `kind: "custom"`, `scheme: "step-up"`, and carries no `issuer` or `mayAct`. Against the compared set: the actor passes, because the parked actor and the re-minted one both come from the same registered agent claims, same `(issuer, subject)`, same `roles: ["agent"]`; `kind` and `scheme` differ and are excluded by design; the subject fails on `issuer` (dropped) and on `roles` whenever the roster changed during the park. So the door copies `issuer`, `mayAct`, `audience` and `clientId` from the parked principal verbatim, and re-verification against the roster becomes a gate rather than a source: if the person is gone, or their roles or the agent's registration differ from what parked, `elevate` throws and the resume is refused; it never mints from the changed values. A renamed agent or a role added during the park is refused, and the requester asks again, so the new park carries the new ring. Refuse, never reconcile, is the rule for the door.

### Review round 2, 18 September

Jaco ruled on the open item and intent raised three corrections plus a docs line. All applied.

- **The ruling.** Bound the lend, recorded under Decided. The open paragraph in Part 3 and the conditional acceptance line are resolved into the settled rule; the framework-owned field now serves both the bound and the loop-closing rule.
- **Correction 1, position 0 is admission rather than the whole chain.** The "FULL chain" wording was wrong and is replaced by the three positions that actually run, `.authorize()`, parse and `.input()`, with every other position keeping its `CHAIN_SURVIVAL` resume column and the recorded reason restated: a refusal below the claim spends the approver's single-use link.
- **Correction 2, one story for schema.** Part 1 accepted `schema` while the acceptance line said the acknowledgment sets none. Settled the way the reviewer chose, which is also how a re-entrant site already behaves: accepted, folded into the descriptor, rendered descriptively, no live validation, the door validates. The acceptance line now matches.
- **Correction 3, the example handler.** The imperative `mailStepUpLink(...)` is gone. The handler forwards to a `direct()` route the application owns, and both stated facts are now in Part 1 and on the reference page: `forward` carries the parked exchange's principal by reference (security.md §3, `getForward(caller)` requires the calling exchange), and `route:exchange:deferred` carries no token, so an `event()`-sourced route can notify but cannot build the link. Verified against the tree: `ForwardFn` is `(endpoint, payload) => Promise<unknown>`, and the deferred event's payload is `ExchangeScoped & { deferralId, position, expiresAt? }`.
- **Docs line.** `claims` is outside the compared set, so a `predicate` reading it on the continuation sees the door's claims. One sentence added to Part 3 for the elevate page.

A divergence raised here (whether correction 3 asked for the notification to be sent before the park) was settled in round 3: it did not, the post-record ordering stands, and the forward belongs inside the hook.

### Review round 3, 18 September

Confirmation and two findings, both accepted, both applied.

- **The hook is confirmed.** The notification rides the directive and commits after the record; correction 3's pre-park phrasing was not an instruction to notify first, and the round-1 fix stands. Recorded under Decided; the divergence note from round 2 is withdrawn.
- **Finding A, the name collided with an opposite ordering.** `deferAside`'s existing `announce` is awaited BEFORE the store write ("The caller learns the id before the record exists ... a crash between the two leaves a reference to release, not a record nothing points at, and an aside deferral has no expiry to retire it otherwise"), while this one must commit AFTER it so a human never holds a link to a record that does not exist. Same subsystem, same word, opposite ordering, and the ordering carries the safety property in both. The new hook is `notify`, verified free as an option name across `routecraft/src` and `ai/src`. Both JSDocs state their own ordering and point at the other with the one-line reason.
- **Finding B, the hook had no bound.** It is awaited inside the executor error path with a network call in it, so an unsettled hook holds the step, which holds `drain()`, and its latency is caller-visible: an MCP tool call would block on the mail send before being told the work parked. `notify` now takes the same treatment `runAuthorizer` already gives the resume `authorize` hook: bounded by the deferring route's own abort signal, widened by an enclosing `.timeout()`, with an abort denying the record claim-first exactly as a throw does, so no live link survives a notification that did not go out. One acceptance line added.

### Related

- #737 is complementary, not a dependency. It is about where the person is asked (in the open conversation, on the editor surface). This is about parking and un-parking without the capability knowing, and it works with a mailed link today. One known limitation, the same as any mailed approval today: after the lend the parked work completes, but the agent that asked is long gone and nobody tells the requester. Closing that loop, and whether the agent's own run parks when a tool parks (Part 1a), is #737's job.
- #736 supplies the vocabulary (defer, deferral, Deferred) this is written in, and the naming rule behind `notify`: one word must not carry two meanings in one subsystem.
- Standards that decide it: `pre-from-filter-chain.md` (the error position is the outermost try/catch, so RC5038 from authorize reaches the handler), `plugin-lifecycle.md` (apply resolves the runtime), `type-safety-and-schemas.md` (plugin vs config vs store), `package-boundaries.md` (Part 4), `security.md` §3 (forward carries identity by reference), §6a and §12 (Parts 1a and 3), `error-and-logging-policy.md`.

### Acceptance

Part 1

- [ ] `recovery.defer(request)` exists, is branded like `drop` and `rethrow`, and `isRecovery` accepts it.
- [ ] Returned from a route-scope `.error()`, it parks the exchange at the failing step (re-entrant) or at position 0 when the failure came from the pre-from chain, and answers with the `Deferred` acknowledgment; `route:exchange:deferred` carries the position.
- [ ] A position-0 site runs `.authorize()`, parse and `.input()` on resume; throttle, circuitBreaker, cacheCheck and cacheStore keep their `CHAIN_SURVIVAL` resume column unchanged. A mid-pipeline site runs under the reduced resume chain and does not re-run the steps before it.
- [ ] A position-0 park presented to a door with no `elevate` is refused with RC5043 at `.authorize()`, non-destructively.
- [ ] An error-path site accepts `schema`, folds it into the continuation descriptor and renders it on the acknowledgment; no live validation runs at resume and RC5049 never fires, and the reference page says the door is the validator, next to the re-entrant rule.
- [ ] `notify` runs only after the record is written and `route:exchange:deferred` has fired, receives the acknowledgment the caller receives (same `deferralId`, same `token`), and a token it hands out verifies against that record.
- [ ] An error-path park refused with RC5051 (split fan-out, sealed side flow) runs no `notify`; nothing external is sent for a park that does not exist.
- [ ] A `notify` that throws denies the record claim-first, so a replay of its token reads RC5050, and the exchange reaches the failure path with the notify error as cause.
- [ ] `notify` is raced against the deferring route's abort signal, widened by an enclosing `.timeout()`; a hook that never settles neither holds `drain()` nor delays the caller's acknowledgment indefinitely, and an abort denies the record claim-first exactly as a throw does.
- [ ] `notify` and `deferAside`'s `announce` each carry a JSDoc stating their own commit ordering and naming the other, so the two are never conflated.
- [ ] A handler that forwards from inside `notify` reaches the target route with the parked exchange's principal, by reference: the target's `.authorize()` sees the same brand state the parked exchange carried, restored or authentic, and never a copy.
- [ ] `InsufficientAuthority.missing` carries `effective`, set by `authorize()`.
- [ ] A cancelled run (`abortSignal.aborted`) refuses to defer with RC5054, the same as the `defer` outcome case.

Part 1a

- [ ] A `direct()` tool whose route answers with a `Deferred` body is recorded by the tool bridge as `DEFERRED_TOOL_PLACEHOLDER`, emits `route:agent:tool:result` with the placeholder snapshot, and the token appears nowhere in the model's context, the session thread, or the telemetry snapshot.
- [ ] The agent's own run does not park on a downstream `Deferred`; the turn continues. Documented on the tool-bridge and agent reference pages, pointing at #737 for the in-conversation case.

Part 2

- [ ] `ctx.registerErrorHandler(handler)` exists and returns an unregister function, the same shape as `ctx.on()`.
- [ ] `CraftConfig.errorHandler` registers the same handler before any plugin's `apply` runs.
- [ ] A context handler receives `forward` bound to the failing exchange, the same as a route `.error()` handler does.
- [ ] Handlers run only when the route has no handler or its handler rethrew or threw; the route handler's recovery is never overridden.
- [ ] Registration order is consultation order; the first non-`undefined` return decides; `undefined` passes.
- [ ] A handler that throws is reported as `route:error-handler:failed` with `scope: "context"` and the chain continues to the next handler; the original error is what reaches the failure path if none decides.
- [ ] `context:error` and `route:exchange:failed` fire only when no handler decided; the events page and the changelog state the `context:error` change.
- [ ] A step-scope `.error()` wrapper either accepts `recovery.defer` under the resilience-wrapper contract or refuses it with a message naming why.

Part 3

- [ ] `.resume(mapper, { elevate })` is accepted; `elevate` is evaluated at step 3 of the ordered window immediately after `authorize` (or in its place when none is declared), above the lifecycle disclosure and the claim, with the raw payload; its result is applied in `rehydrate()` after the claim.
- [ ] A returned principal that is not live-branded, or that differs from the parked principal in anything other than `scopes` on the subject or on the outermost actor (compared set as in Part 3), is refused with RC5056, non-destructively, and logged the way `runAuthorizer` logs a refusal.
- [ ] The deferral record carries the framework-owned field with the RC5038 refused scopes; a lend wider than that field is RC5056, and a resumed exchange is not parked again for scopes already in it.
- [ ] A door without `elevate` behaves exactly as today: the continuation carries the restored principal.
- [ ] When `elevate` supplied a principal, the continuation re-runs the route's `.authorize()`; a principal that still lacks the scope fails with RC5038.
- [ ] `resumedBy` on the record still reflects the door's live principal, not the elevated one.
- [ ] The elevate reference page states that a lend on the actor's ring is read only by a gate with `effective: true`, and that a `predicate` reading `principal.claims` on the continuation sees the door's claims rather than the parked ones.

Part 4

- [ ] All `agent:*` and `route:agent:*` entries are declared in `@routecraft/ai` through declaration merging and removed from core's `types.ts`; the events reference page moves them under the AI package's section.
- [ ] No emit site changes; the type-level test that names every event still passes.

Boot and docs

- [ ] `assertDeferralConfigured()` fires RC5052 at start when a deferring context handler is registered and the runtime is absent; `routeCanDefer` reports true for every route while one is registered.
- [ ] A plugin that registers a deferring handler and fails to resolve the runtime in `apply` fails the build with the runtime's own error, per the plugin lifecycle standard.
- [ ] Reference pages: the recovery directive with `notify`, its bound, its ordering against `deferAside`'s `announce`, the forward-from-notify example and the no-token-on-the-event note; the context error handler (config and plugin forms); `elevate` on the resume page with the identity rule, the bound, the ordered-window placement and the claims sentence; the tool-bridge placeholder rule; and the pre-from filter chain standard updated for the resume column of authorize and the position-0 admission rule.
- [ ] Changelog entry under the canary notes, naming the `context:error` change.

ex0b1t:
## TLDR

An `.error()` handler can now answer `recovery.defer(request)`, so a failure **parks the exchange durably** instead of being recovered, dropped or rethrown. The same error ring now also exists **outside the route**, through `ctx.registerHandler("error", ...)`, so a plugin can decide failures for routes that never mention it. A resume door may **re-mint the parked principal** with `elevate`, which is what makes a generic step-up possible: refused for a missing scope, park, a human lends the scope, resume, succeed, with no route written for it.

Closes #810.

**Breaking for route authors:** `context:error` and `route:exchange:failed` now fire only when nothing decided the failure, so a failure a context handler resolves reaches neither.

---

## Status at `9a54dc9`

Read this before reviewing the diff; it is not ready to merge.

| | |
|---|---|
| CI | **Green.** Every completed run since `f3b23aa` has passed all 13 checks. The two `check-links.ts` failures were one cause: a heading carrying inline code does not slug the way its text reads, so the fix was an explicit anchor id rather than a retargeted link. |
| Bot review | cubic and CodeRabbit have both reviewed every head. CodeRabbit's last two runs reported **no actionable comments**. Every finding either landed as a fix below or was refused with reasoning on its thread. CodeRabbit's merge-risk banner is pinned to `d16c9a0` and names two defects that are both fixed; it cannot refresh while rate limited. |
| Local gate | `bun run all` at `9a54dc9`: exit 0, 3858 pass, 0 fail across 275 files, **including the link check** (18974 links, 273 pages). Ruling 4 is closed; the gate is now the one `CLAUDE.md` describes. |

**Two findings contradicted claims this description previously made. Both were real; both are now fixed.**

1. ~~`notify` throwing after the park commits gives one exchange **two terminal events**.~~ **Fixed at `26e0b0a`**, mutation-checked. `route:exchange:deferred` now fires after `notify` settles, so a successful park emits `deferred` and nothing else and a failed notify emits `failed` and nothing else. The ordering that mattered, the record being durable before anyone is told, is unchanged; only the event moved. Reverting it fails two tests, one of them the old contract's own ordering test.
2. `noteFailingStep` is **first-writer-wins on error object identity**, so one error object thrown from two different steps binds a park to the wrong position. Same bug class as the branch's correctness fix 2, surviving in a different form. Contrived trigger, serious consequence: a continuation replaying completed steps under a spent approval. **Still open**, being sized.

Stated plainly because it is the honest read: the branch's own seven-lens review found the right two bugs and closed neither completely.

**A third, where both bots agreed with each other and neither proposed fix was taken. Fixed at `fcaef5f`, mutation-checked.** The context error ring was consulted above the `rethrowUnhandled` escape, and the reviewers called it a retry suppression. Read against the code it is worse than that, and it is a precedence inversion rather than a new design question.

`nestedDeps` builds a resilience segment's definition with **no `errorHandler`**, so the route's own ring cannot fire inside `.retry()`, `.timeout()`, `.circuitBreaker()` or `.concurrency()`. Before this branch, no ring fired there at all: the segment rethrew, the wrapper reacted, and the route's handler got its turn at the outermost run once attempts were exhausted. This branch added a ring that fires in a place no ring has ever fired. Reproduced: a route with `.retry({ maxAttempts: 3 })`, its own `.error()`, and a context handler took **one attempt** and returned the context handler's decision; without the context handler it took three and returned the route's.

Such a run now consults no error ring at all and surfaces its failure to the wrapping segment, which restores the prior behaviour rather than choosing a new one. A park therefore lands after the attempts are exhausted rather than on attempt one, which is the wanted order: nobody should be asked to approve something an automatic retry would have fixed.

---

## What you are ruling on

Everything else in this PR is settled. These four are not.

### 1. WITHDRAWN. A route-scope `.cache()` and an error-path park

**I put this to you and my premise was wrong. Nothing to decide; no change made.**

I claimed a route with `.cache()` and no `.defer()` would "silently never cache" once error-path parks became reachable, and recommended moving the builder's refusal to context start time. Checked against the mechanism rather than the analogy, that is not what happens.

`builder.ts` refuses `.cache()` alongside a static `.defer()` because EVERY exchange then exits the pipeline, so the cache is completely dead. An error-path park is not that: only the exchanges that actually fail and park miss the store. Normal traffic caches exactly as declared.

Nothing leaks either. The route-scope check is a plain `provider.get(key)`, not the `getOrCompute` the step-scope wrapper uses, so no lock or in-flight entry is reserved and a park leaves nothing behind. And the store step reads the key the check stashed in exchange internals rather than re-deriving it, so after a park and rehydrate the key is gone: `survives: false` on that row matches what the mechanism could do even if the policy said otherwise.

What remains is that an exchange which needed a human does not populate the cache, so the next identical request recomputes. That is a minor cost on an exceptional path, and arguably the behaviour you want.

The one thing I got right is now fixed anyway: the row's stated reason cited a build refusal that does not apply to error-path parks.

### 2. The breaking event change

**Recommendation: accept it.**

`context:error` and `route:exchange:failed` now fire only when nothing decided the failure. A subscriber that counted failures will under-count once a handler starts resolving them.

It is the right way round. An event that fires for a failure something already resolved is a lie, and the alternative makes every subscriber re-derive whether the failure mattered. But it is a real behaviour change for anyone alerting on those events, and it is yours to accept rather than mine.

### 3. `executor.ts` is now roughly 1840 lines and owns two jobs

**Recommendation: agree with the decline, file the follow-up.**

Review asked for the error-path policy (`ErrorDecision`, the failing-step accessors, `runContextErrorHandlers`, `applyErrorDecision`, `parkFromErrorPath`) to move into `pipeline/error-path.ts`. Correct and real: step scheduling and error-ring policy change for different reasons.

Declined on this branch because it is a roughly 390-line move of code that has just absorbed two correctness fixes, and both of those have since turned out to be incomplete. Moving it now trades a reviewed diff for an unreviewed one at exactly the wrong moment. Overrule me if you would rather take it now.

### 4. CLOSED. `bun run all` is now the pre-PR gate `CLAUDE.md` says it is

**Done at `9a54dc9`. Recorded because it is a change to how everyone works, not just to this branch.**

`all` ran lint, format, typecheck, build and test, and never the link check. A clean local run could not catch a broken doc anchor, which is how this branch shipped two and learned it from CI.

The checker also walked prerendered output without checking it was current, so running it after a source edit reported the PREVIOUS build's answer: a link just fixed still read as broken, a link just broken still read as fine. It now refuses when any content file is newer than the build. That guard caught a stale build on its first run, against an edit made seconds earlier.

Overrule me if you would rather the link check stayed out of `all` for speed; it adds the docs-site prerender to a local run that already builds it.

---

### A failure can park instead of ending

```ts
craft()
  .id('payout')
  .error((error, exchange, forward) => {
    const refusal = insufficientAuthorityOf(error)
    if (!refusal || !refusal.effective) return undefined
    return recovery.defer({
      ttl: '4h',
      meta: { scopes: refusal.scopes },
      notify: (ack) => forward(sendStepUpMail, { token: ack.token }),
    })
  })
  .authorize({ scopes: ['payout:write'] })
  .from(http({ path: '/payout', method: 'POST' }))
  .to(payOut())
```

### And the resume can carry what a human lent

```ts
craft()
  .id('approvals')
  .from(mail('INBOX'))
  .authenticate(mailPrincipal)
  .resume(payloadFrom, {
    authorize: ({ principal }) => mayApprove(principal),
    elevate: ({ principal, record }) => lendRefusedScopes(principal, record),
  })
```

### The same ring, outside any route

```ts
defineConfig({
  handlers: { error: auditThenPass },
  deferral: {},
})

ctx.registerHandler(
  'error',
  (error, exchange, forward, ctx) =>
    isMine(error) && ctx.execution === 1
      ? recovery.defer({ ttl: '4h' })
      : undefined,
  { tags: ['gated'], mayDefer: true },
)
```

Both fragments are illustrative: `insufficientAuthorityOf`, `recovery`, `registerHandler` and `elevate` are real, the helper functions are not.

### Before and after

- **Before:** a failure could be recovered, dropped or rethrown. A park had to be written into the happy path ahead of time, and only a route could decide what its own failures meant.
- **After:** a failure can park at the step that failed or at admission, a plugin can decide recovery for routes that know nothing about it, and a resume can carry lent scopes the parked exchange never had.

---

## Security surface, restated for your final check

This is the part worth re-reading rather than skimming, because it is where the feature can go wrong quietly.

**`elevate` re-mints a parked principal.** It runs at step 3 of the resume window, after `authorize`, above lifecycle disclosure and above the claim, and its result is applied in `rehydrate()` only once the claim succeeds. The elevated principal is compared structurally against the parked one on every field but `scopes`, so the door may lend authority and may not swap identity.

**The lend bound is counted per ring.** Merged, a door could move a scope the park held on the **actor** onto the **subject**, lend nothing by the merged count, and satisfy a subject-ring gate that had already refused. `lentScopes` diffs `parked.scopes` against `elevated.scopes` and `parked.actor?.scopes` against `elevated.actor?.scopes` separately.

**Brands stay module-private.** `markAuthentic` and `markRestored` remain private WeakSets. Nothing in this diff exports a way to mint trust.

**The resume token does not reach the model.** The `@routecraft/ai` tool bridge records `DEFERRED_TOOL_PLACEHOLDER` when a downstream route parks, so the live token in the acknowledgment reaches neither the model's context, the session thread nor the telemetry snapshot.

**Still open on this surface**, both from bot review and both being fixed: `insufficientAuthorityOf` accepts any non-Error object carrying `rc: "RC5038"`, against its own contract, which lets a generic handler park an unrelated failure as an approval request; and `routecraft.deferral.refusedScopes` is accepted as ordinary caller input at ingress.

---

## What changed

**`packages/routecraft`, the directive.** `recovery.defer(request)` returns an `ErrorPathDeferRequest`, extending `DeferSignalRequest` with `notify`. `deferral/sites.ts` stamps an `ErrorPathSite` on every step during the walk, so the position a park resolves to is decided by the same walk that decides static `.defer()` sites, and a position a deferral could not be revived from (inside a `.split()`, under a sealed scope) is a recorded `RC5051` refusal rather than a park that cannot come back. `notify` is awaited **after the record is written and before `route:exchange:deferred` fires**: after the write, so nobody is handed a token for a park that never committed, which is the opposite ordering to `deferAside`'s `announce`; before the event, so a notification that fails does not announce a park it then denies. Both JSDocs state their own ordering and name the other.

**`packages/routecraft`, the ring.** `HandlerPointRegistry` is a declaration-merged interface mapping each handler point to its handler type, so `ctx.registerHandler(point, handler, selector?)` is typed per point and a duplicate name with a different signature is a compile error. Only `error` is registered here; the three remaining points are #816 and are deliberately not in this branch, with the registry and config key shaped so nothing is renamed when they land. The selector matches by `route.definition.id` and by discovery tags, as alternatives. The handler's fourth argument is an `ErrorContext` carrying `route` and `execution` (`1` on the first run, `2` on a resumed continuation), not the bare `Route` it began as: a handler that parks without reading `execution` parks the continuation again and asks the same human twice. A three-parameter `ErrorHandler` stays assignable. Neither ring is consulted inside a resilience segment, so the route's own handler and its declared retry policy both keep precedence over anything registered on the context.

**`packages/routecraft`, the door.** `elevate` as described under Security above. `deferral/authorize.ts` holds the comparison and the per-ring lend count.

**`packages/ai`.** The tool bridge placeholder as described above. Every `agent:*` and `route:agent:*` event moves out of core's `EventDetailsMap` into this package through declaration merging; no name, payload or emit site changes.

**Shared.** One `settleOrAbort` replaces three copies of the hook-race scaffold; one `refusalOf` replaces two copies of the refusal builder; `@routecraft/ai` keeps one `isThenable` rather than widening core's `@internal` export.

### Two correctness bugs found and fixed on this branch

Both mutation-checked: the fix was broken deliberately, the failure reproduced with the defect's own signature, then the fix restored. **See Status above: bot review showed both fixes incomplete, one of the two gaps is now closed and the other is open.**

1. **A refused or failed park escaped the failure path.** The decision was applied outside the guard, so an `RC5051` refusal, an `RC5052` for a missing deferral runtime, or a failed store write left `runPipeline` with no terminal event at all.
2. **A park could resolve to the wrong position.** The failing step was inferred from the error object, and a route handler that throws its own error hands the chain a fresh one with no entry, which read as "outside the step tree" and parked at position 0 with the whole body as continuation. The executor now passes the step it is holding.

Found while verifying those: a route-scope `.timeout()` firing after two steps had run was recorded as an admission park at position 0. Admission is now a tracked fact rather than inferred, and a post-admission failure with no site is `RC5051` rather than a replay. Swept across all four resilience carriers.

### An acceptance item that was not actually satisfied

`routeCanDefer(definition, context)` was never passed the context at its only shipped call site, so an MCP-exposed route stayed unadvertised under a deferring context handler. The parameter was dead and the item only looked met.

---

## Tests

51 new across four files: `error-path-defer.bun.test.ts` (16), `context-error-handlers.bun.test.ts` (23), `resume-elevate.bun.test.ts` (11, across the identity rule, the per-ring bound and the ordered window), `agent-downstream-deferred.bun.test.ts` (1). Every fix on this branch that closes a correctness defect has a test that fails with that defect's own signature when the fix is reverted: the two original ones, the per-ring lend bound, the `execution` flag, the `notify` terminal-event ordering, and the ring precedence inside a resilience segment.

Guarding behaviour that must not change: the existing continuation-filter-chain policy table gained the new `errorHandler` and `admission` kinds rather than being replaced, so a new detached kind still fails the build until it is classified, and the resume door's behaviour with no `elevate` is asserted identical.

Bot review has flagged several of these tests as asserting less than their titles claim. Those are on the fix list.

---

## Docs and changeset

Changeset: `.changeset/error-path-defer.md`, minor for `@routecraft/routecraft` and `@routecraft/ai` (breaking is minor during v0).

Pages: `reference/operations/error`, `reference/operations/resume`, `reference/operations/defer`, `reference/configuration`, `reference/events`, `reference/errors` with `_data/errors.json`, `advanced/error-handling`, `advanced/durable-agents`, and the changelog. Standards touched: `pre-from-filter-chain.md`, `resilience-wrappers.md`, `error-and-logging-policy.md`.

`DEFINITION_OF_DONE.md` carried a false merge gate: it required new event names to stay compatible with `*` and `**` glob patterns, while `EventBus.on()` throws `RC2001` for any name containing `*` once the bare catch-all is handled. The `**` half was false and the `*` half vacuous. Corrected here, along with the naming example in the same list, which still showed identity embedded in the event name.

---

## Leftovers

- Extract the error-path policy into `pipeline/error-path.ts`. See ruling 3.
- The three remaining handler points (`admission`, `entry`, `exit`) are #816. The open decision there is settled: admission sits after source-supplied verification and may not replace a transport-verified principal, and a lent elevation must survive it.

---


ex0b1t:
## Halted pending a redesign. Draft, not abandoned.

Green at `ce0c38e` and every review thread either fixed or answered, but it is not merging in this shape.

**Why.** Roughly two thirds of the findings on this branch were two design problems repeating rather than independent bugs, and both are structural rather than anything this branch did wrong.

`RouteDefinition` carries 21 fields. Three are identity, three are wiring, and fifteen are cross-cutting capability state bolted onto a route's identity: nine chain positions, five deferral resolver outputs, one principal flag. The executor reaches directly into twelve of them. `CHAIN_SURVIVAL` is keyed by `Exclude<keyof RouteDefinition, NonChainField>`, a policy table keyed by that bag with a hand-maintained ten-entry carve-out list.

The clearest evidence is in this branch's own follow-up work: #816 could not put its handler points into `CHAIN_SURVIVAL`, because a handler point is not a `RouteDefinition` key, so it built `POINT_SURVIVAL` as a parallel table. Two tables answering the same question for two kinds of participant, because the first is keyed by a field bag rather than a concept.

Everything we chased follows from that. Six separate carriers of "where is this exchange and what re-runs" exist because nothing owns that question, and each defect was two of them disagreeing. The header ingress strip exists because deferral state had nowhere to live but a caller-writable bag, which is also where the `refusedScopes` security finding came from.

**What is being considered instead**, with the DSL and `defineConfig` unchanged: one `Stage` concept that owns its own placement, its own continuation policy and its own namespaced state, so cache, throttle, breaker, retry, timeout, concurrency, authorize, error, defer, resume and the handler points all become contributed participants that answer for themselves; and one `Continuation` value computed once at park time, so nothing is re-derived or inferred at resume.

**What stays true of this branch.** The requirements it implements are not in question and are the input to the redesign. The work is preserved here, gated, including four defect fixes in the final commit that no bot found.

Do not merge. Follow-up ticket to come.

---

